PR #2, opened 2016-03-16
Writing

martindurant:
Uploads are multi-part, with in internal IO buffer, which is flushed to the server when it grows over 5MB or the given block_size.

Question: is there ever a circumstance with a need to call `abort_multipart_upload`?


mrocklin:
You may want to rebase out the first commit, which I rewrote on master soon after pushing.  I only changed authorship (to you) so this should be a clean rebase.

s3fs/core.py
```diff
@@ -13,6 +17,21 @@
...
+
+def tokenize(*args, **kwargs):
+    """ Deterministic token
+
+    >>> tokenize('Hello') == tokenize('Hello')
+    True
+    """
+    if kwargs:
+        args = args + (kwargs,)
+    return md5(str(tuple(args)).encode()).hexdigest()
```
mrocklin: I'm not sure that `str` on a dict is

s3fs/core.py
```diff
@@ -13,6 +17,21 @@
 
 logging.getLogger('boto3').setLevel(logging.WARNING)
 logging.getLogger('botocore').setLevel(logging.WARNING)
+S3_RETRYABLE_ERRORS = (
+    socket.timeout,
+    trans.ReadTimeoutError, trans.IncompleteReadError
+)
```
mrocklin: Why not steal this from botocore?

s3fs/core.py
```diff
@@ -142,7 +169,10 @@ def _ls(self, path, refresh=False):
                     f['Size'] = 0
                     del f['Name']
             else:
-                files = self.s3.list_objects(Bucket=bucket).get('Contents', [])
+                try:
+                    files = self.s3.list_objects(Bucket=bucket).get('Contents', [])
+                except ClientError:
+                    files = []
```
mrocklin: Should we raise a `FileNotFound` error if the user doesn't provide a nice location?

s3fs/core.py
```diff
@@ -150,6 +180,7 @@ def _ls(self, path, refresh=False):
         return files
 
     def ls(self, path, detail=False):
+        """ List single "dirtory" with or without details """
```
mrocklin: typo?

s3fs/core.py
```diff
@@ -383,13 +512,72 @@ def read(self, length=-1):
...
+                self.forced = True
+            self.buffer.seek(0)
+            try:
+                part = len(self.parts) + 1
+                out = self.s3.s3.upload_part(Bucket=self.bucket, Key=self.key,
+                            PartNumber=part, UploadId=self.mpu['UploadId'],
+                            Body=self.buffer.read())
+                self.parts.append({'PartNumber': part, 'ETag': out['ETag']})
+            except ClientError:
+                raise IOError('Write failed', self)
```
mrocklin: Do we need retry logic here?

s3fs/core.py
```diff
@@ -411,7 +598,7 @@ def _fetch_range(client, bucket, key, start, end, max_attempts=10):
             resp = client.get_object(Bucket=bucket, Key=key,
                                      Range='bytes=%i-%i' % (start, end - 1))
             return resp['Body'].read()
-        except boto3.s3.transfer.S3_RETRYABLE_ERRORS as e:
+        except S3_RETRYABLE_ERRORS as e:
```
mrocklin: Why the switch?

s3fs/core.py
martindurant: Different versions of botocore seem to have this in different places or not at all.

s3fs/tests/test_s3fs.py
```diff
@@ -25,21 +26,103 @@
...
-def test_non_anonymous_access():
+    # writable local S3 system
+    m = moto.mock_s3()
+    m.start()
+    import boto3
+    client = boto3.client('s3')
+    client.create_bucket(Bucket=test_bucket_name)
+    for k in [a, b, c, d]:
+        try:
+            client.delete_object(Bucket=test_bucket_name, Key=k)
```
mrocklin: Under what conditions would this fail in moto?

s3fs/tests/test_s3fs.py
```diff
@@ -141,3 +279,51 @@ def test_read_s3_block(s3):
...
+def test_new_bucket(s3):
+    assert not s3.exists('new')
+    s3.mkdir('new')
+    assert s3.exists('new')
+    s3.touch('new/temp')
+    with pytest.raises((IOError, OSError)):
+        s3.rmdir('new')
+    s3.rm('new/temp')
+    s3.rmdir('new')
+    assert not s3.exists('new')
```
mrocklin: Very cool to see this functionality arise

s3fs/core.py
martindurant: It isn't obvious to me. S3 doesn't actually have directories.

s3fs/core.py
martindurant: my version of boto3 doesn't have that defined

s3fs/core.py
martindurant: We'll see if the same issue arises - it always passed locally.

s3fs/core.py
mrocklin: My experience with `distributed.s3` is that I frequently make mistakes writing down the bucket/key of my data and only realize the mistake far later on, when I find that my bag/dataframe is empty.  I these cases I would have preferred receiving a `FileNotFound` exception early.  I think that erring when performing read operations on non-existent locations is also the standard expectation within other projects as well.

s3fs/tests/test_s3fs.py
martindurant: If the object didn't exist - i.e., a failing earlier test may or may not pollute the bucket with leftover stuff and cause other tests to fail

s3fs/core.py
martindurant: I would appreciate your help, in that case, on making a thorough set of conditions and test for `ls` - I found it a bit tricky to decide what it should do, is it always the same as `glob(path)`?

mrocklin: Under what conditions does the `ClientError` get raised?

Is there a cheap way to determine the existence of a bucket?

When listing out keys in the bucket if there are no matches perhaps this means that the implied directory does not exist?

mrocklin: Looking around on SO it appears that `ClientError` is raised on missing bucket?  This seems to be verified when using locally:

```python
In [1]: from distributed.s3fs import S3FileSystem
s3 
In [2]: s3 = S3FileSystem()

In [3]: _ = s3.s3.list_objects(Bucket='dask-data')

In [4]: _ = s3.s3.list_objects(Bucket='shffhhf')
---------------------------------------------------------------------------
ClientError                               Traceback (most recent call last)
```

Maybe raise a `FileNotFoundError` in the `except` branch?

This doesn't resolve the case when a bucket exists but no keys do.  Again, I'd be tempted to raise `FileNotFoundError` here if no matches, thoughts?

martindurant: - Is there a cheap way to determine the existence of a bucket?
A listing of buckets is not enough, as the path could be to a public bucket not associated with the user. Listing the objects ought to be slower than doing `head_bucket`, but probably not by much compared with the connection time; I can't tell the difference with `%time`.

Worth noting that if the following keys exist
bucket/file1
bucket/dir/dir/file2
then ls('bucket') will NOT give 'bucket/dir' or 'bucket/dir/dir', but perhaps it should. The trouble is that 'bucket/dir' can also be a file with data.

s3fs/core.py
martindurant: fixed

s3fs/core.py
martindurant: (the latter point is only relevant if we really want the filesystem to be quasi-posix, for instance for later FUSE integration)

OK to make ls with missing bucket an error; I don't think it's an error to list a bucket that contains no keys, though, only specifying a path where the root doesn't exist either,

mrocklin: You might take all files, and normalize them by split('/', 1), and then take unique entries?

```python
In [1]: fns = ['foo/a', 'foo/bar/a', 'foo/b', 'foo/bar/b']
In [2]: root = 'foo'
In [3]: {'%s/%s' % (root, fn[len(root) + 1:].split('/')[0]) for fn in fns}
Out[3]: {'foo/a', 'foo/b', 'foo/bar'}
```

Perhaps call sorted on the output as well?

mrocklin: > I don't think it's an error to list a bucket that contains no keys, though, only specifying a path where the root doesn't exist either,

Agreed that existing empty bucket should return `[]`

martindurant: OK, I can go with your example - ls returns empty entries (size=0) to fill in the directory space, but not overwriting any entries that happen to be keys sharing the same name. 

Should they, then, be available to `open()` as empty files? (probably not)

Should they show up in walk or glob? (probably not)

Entries are already sorted in `_ls`.

martindurant: Would it be fair to say that this is an enhancement unrelated to the PR, so should merge and make a new PR that does this well? PR #1 isn't even in yet.

mrocklin: Sure

mrocklin: I'm mostly leaving `s3fs` in your hands by the way.  Please feel free to merge anything as you like.

martindurant: Understood, but I still appreciate any review time you can spare for this. Once the writing is merged and other changes like ls above, I don't expect much to change. Will have to make sure, of course, that everything in dask/s3 runs after any merger.

mrocklin: Always happy to review.  Mostly I'm removing cleanliness blocks that I might have been imposing in the `distributed` codebase.

PR #6, opened 2016-03-17
Notfound

martindurant:


s3fs/core.py
```diff
@@ -185,7 +190,7 @@ def ls(self, path, detail=False):
         try:
             files = self._ls(path)
         except ClientError:
-            files = []
+            raise FileNotFoundError(path)
```
mrocklin: Things like this could use tests, ideally in the same commit where the functionality was added.

s3fs/core.py
```diff
@@ -193,7 +198,7 @@ def ls(self, path, detail=False):
                 try:
                     files = [self.info(path)]
                 except (OSError, IOError, ClientError):
-                    files = []
+                    []
```
mrocklin: ?

martindurant:
Actually, I'll just get rid of all of this for now - clearly not turning out quality.

s3fs/core.py
mrocklin: The fact that this change was made and tests still passed is very concerning.

PR #7, opened 2016-03-17
Add several tests

martindurant:
Introduced FileNotFoundError with ls on a bucket that cannot be accessed

### DIFF
--- a/s3fs/tests/test_s3fs.py
+++ b/s3fs/tests/test_s3fs.py
@@ -294,7 +294,8 @@ def test_errors(s3):
         s3.open('anyfile', 'ab')
 
     with pytest.raises(ValueError):
-        s3.open(test_bucket_name+'/temp', 'wb').read()
+        with s3.open(test_bucket_name+'/temp', 'wb') as f:
+            f.read()
 
     with pytest.raises(ValueError):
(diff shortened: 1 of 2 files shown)


s3fs/tests/test_s3fs.py
```diff
@@ -113,12 +132,19 @@ def test_rm(s3):
...
+    assert not s3.exists(test_bucket_name+'/nested/nested2/file1')
 
 
 def test_anonymous_access():
     with ignoring(NoCredentialsError):
-        fs = S3FileSystem(anon=True)
-        fs.ls(test_bucket_name)
+        s3 = S3FileSystem(anon=True)
+        assert s3.ls('') == []
+        ## TODO: public bucket doesn't work through moto
```
mrocklin: Is this TODO still the case with the `ACL` change above?

mrocklin:
This looks fantastic to me.

s3fs/tests/test_s3fs.py
martindurant: ACL did not seem to fix this. I'll look into it, but intend to leave as-is for now.

mrocklin:
Is it safe to delete these extra branches hanging out on the main fork?

PR #8, opened 2016-03-19
Add get and put methods, just like in hdfs3

martindurant:
Small convenience methods.

Note: the buffer size is the same as the default block-size (currently 5MB).

### DIFF
--- a/s3fs/tests/test_s3fs.py
+++ b/s3fs/tests/test_s3fs.py
@@ -267,6 +267,17 @@ def test_move(s3):
     assert not s3.exists(fn)
 
 
+def test_get_put(s3):
+    import tempfile
+    fn = tempfile.mktemp()
+    s3.get(test_bucket_name+'/test/accounts.1.json', fn)
+    data = files['test/accounts.1.json']
+    assert open(fn, 'rb').read() == data
+    s3.put(fn, test_bucket_name+'/temp')
+    assert s3.du(test_bucket_name+'/temp')[test_bucket_name+'/temp'] == len(data)
+    assert s3.cat(test_bucket_name+'/temp') == data
+
+
 def test_errors(s3):
     with pytest.raises((IOError, OSError)):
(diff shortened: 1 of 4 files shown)


s3fs/tests/test_s3fs.py
```diff
@@ -267,6 +267,17 @@ def test_move(s3):
...
 
+def test_get_put(s3):
+    import tempfile
+    fn = tempfile.mktemp()
+    s3.get(test_bucket_name+'/test/accounts.1.json', fn)
+    data = files['test/accounts.1.json']
+    assert open(fn, 'rb').read() == data
+    s3.put(fn, test_bucket_name+'/temp')
+    assert s3.du(test_bucket_name+'/temp')[test_bucket_name+'/temp'] == len(data)
+    assert s3.cat(test_bucket_name+'/temp') == data
```
mrocklin: Might want to delete the file afterwards, otherwise we end up filling up `/tmp`.  I recommend stealing the `with tmpfile() as fn:` context manager from `dask/utils.py`

mrocklin:
Looks great to me

s3fs/tests/test_s3fs.py
martindurant: I thought the OS took responsibility for clearing tempfiles from processes that end. If not, I may have many of these around...

tmpfile is a little more complex than I would have expected; anything wrong with the following:
```python
@ontextmanager
def tmpfile():
    fn = tempfile.mktmp()
    try:
        yield fn
    finally:
        if os.exists(fn):
            os.remove(fn)

mrocklin: I use `shutil` in case we use the generated name to create a file.  `mktemp` has, I think, some threadsafety issues, mkstemp is preferred I think.  `os.remove(fn)` fails in Windows if the file is still open.

`dask.utils.py:tmpfile` started out as something similar to what you propose.  It grew as we ran into more and more complex testing situations which inevitably arise in the real world (open files, windows, directories, threads.)

PR #9, opened 2016-03-21
Path corrections

martindurant:
This is in lieu of changing the behaviour of `ls` for the time being.

+ small cosmetic changes.


s3fs/core.py
```diff
@@ -215,6 +226,7 @@ def info(self, path):
 
     def walk(self, path):
         """ Return all entries below path """
+        path = path.lstrip('s3://')
```
mrocklin: Beware, `lstrip` doesn't work this way.  It strips all occurrences of `s, 3, :, /` from the front.

```python
In [1]: 'Hello'.lstrip('Hel')
Out[1]: 'o'
```

You probably want something like the following:

    path = path[len('s3://'):] if path.startswith('s3://') else path

s3fs/core.py
```diff
@@ -355,11 +368,12 @@ def rm(self, path, recursive=False):
                 raise IOError('Delete key failed', (bucket, key))
             self._ls(path, refresh=True)
         else:
-            if recursive or not self.s3.list_objects(Bucket=bucket).get('Contents'):
+            if not self.s3.list_objects(Bucket=bucket).get('Contents'):
                 try:
                     self.s3.delete_bucket(Bucket=bucket)
                 except ClientError:
                     raise IOError('Delete bucket failed', bucket)
+                self.dirs.pop(bucket, None)
```
mrocklin: What was the cause for this?

martindurant: If you delete a bucket, the entry will still be in the `dirs` cache (although it should be empty), so `ls(mybucket)` would not produce the not-found error that it should.

s3fs/core.py
martindurant: OK, thank you. 
I'm not happy how often such path mangling happens through the code, also in the path_split function.

mrocklin: Would some sort of `normalize_path` function be appropriate?

s3fs/core.py
mrocklin: Ah!  Hence this addition

```python
     assert not s3.exists('new')
 +    with pytest.raises((IOError, OSError)):
 +        s3.ls('new')
```

PR #16, opened 2016-03-23
Bump to 0.0.3 and use conda noarch in recipe

koverholt:


koverholt:
A `noarch` package for s3fs 0.0.3 is available from:

```
conda install s3fs -c dask
```

PR #18, opened 2016-03-27
WIP/ENH: Make S3File iterable, readline

TomAugspurger:
Adds a `readline` method to S3File, which is used to make them
iterable via py2 style `.next` and py3 style `__next__` methods.

### DIFF
--- a/s3fs/tests/test_s3fs.py
+++ b/s3fs/tests/test_s3fs.py
@@ -499,3 +499,26 @@ def test_write_blocks(s3):
         f.write(b'a' * 15*2**20)
         assert f.buffer.tell() == 0
     assert s3.info(test_bucket_name+'/temp')['Size'] == 15*2**20
+
+
+def test_readable(s3):
+    with s3.open(a, 'wb') as f:
+        assert not f.readable()
+
+    with s3.open(a, 'rb') as f:
+        assert f.readable()
+
+
+def test_seekable(s3):
+    with s3.open(a, 'wb') as f:
+        assert not f.seekable()
+
+    with s3.open(a, 'rb') as f:
+        assert f.seekable()
+
+def test_writable(s3):
+    with s3.open(a, 'wb') as f:
+        assert f.writable()
+
+    with s3.open(a, 'rb') as f:
+        assert not f.writable()
--- a/s3fs/tests/test_s3fs.py
+++ b/s3fs/tests/test_s3fs.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 import io
+from itertools import chain
 import pytest
 from s3fs.core import S3FileSystem
 from s3fs.utils import seek_delimiter, ignoring, tmpfile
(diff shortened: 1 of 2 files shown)


s3fs/tests/test_s3fs.py
```diff
@@ -79,6 +80,75 @@ def test_simple(s3):
...
+        f.write(data)
+    with s3.open(a, 'rb') as f:
+        result = r.readline()
+        assert result == data
+
+def test_next(s3):
+    expected = csv_files['2014-01-01.csv'].split(b'\n')[0] + b'\n'
+    with s3.open(test_bucket_name + '/2014-01-01.csv') as f:
+        result = next(f)
+        assert result == expected
```
mrocklin: Might want to try `__iter__` as well.  Something like the following:

    assert list(f) == csv_files['2014-01-01.csv'].split(b'\n')

s3fs/tests/test_s3fs.py
```diff
@@ -79,6 +80,75 @@ def test_simple(s3):
...
+
+def test_readline_empty(s3):
+    data = b''
+    with s3.open(a, 'wb') as f:
+        f.write(data)
+    with s3.open(a, 'rb') as f:
+        out = f.readline()
+        assert out == b''
+
```
mrocklin: Do these tests stress blocksize?

s3fs/tests/test_s3fs.py
TomAugspurger: I think I pushed a second commit after you made this note testing something like that https://github.com/dask/s3fs/pull/18/commits/a27d2916feba0743b8c75127e6a4049d2bf36b9b

s3fs/tests/test_s3fs.py
TomAugspurger: They do not, and it looks like I don't handle the case where the line is longer than blocksize correctly. Fixing now.

TomAugspurger:
https://github.com/dask/s3fs/pull/18/commits/d5ab2dfdedc40022e8014a7db8774771f65864d3 should take care of the blocksize issue.

TomAugspurger:
The last commit added some file-like properties `readable`, `writeable`, and `seekable`. These were enough to pass most of the pandas test-suite (which doesn't have too many s3 tests admittedly). I was briefly looking into making `S3File` a proper subclass of one of the AbstractBaseClasses in `io`, but this seems to be enough for now.

There are a couple more pandas issues I'm working around.

- We had some tests for `s3n` and `s3a` [URLs](https://wiki.apache.org/hadoop/AmazonS3). I'm not sure whether pandas was properly handling those in the past, we probably just treated them like regular s3 urls.
- Maybe something pandas dealt with [here](https://github.com/pydata/pandas/pull/10604). Basically a public bucket that has a public objects sitting next to private objects. When we go to list the objects in the directory, we get a permissions exception.

Will followup tonight hopefully. I can keep throwing changes in the PR if you want, or followup with another.

mrocklin:
Whatever makes you happier.  I'm really quite glad that you in particular
are handling this.

martindurant:
Please compare with my version in PR #19.
The readline function there is somewhat simpler, and I believe copes with cases where a seek precedes a readline.
Interesting to see the requirements for readable etc.

TomAugspurger:
The need for readable, seekable, etc. are probably peculiar to pandas. We've got the C parser, and a fallback parser written in Python. For the Python parser, I needed to wrap the byte stream from here in a `TextIOWrapper`, which required those methods to be present.

Close this one in favor of #19?

martindurant:
If you're happy with #19, then yes.

PR #19, opened 2016-03-28
Readline support with pandas additions from PR #18

martindurant:
Special attention to rerely-used length parameter

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -549,7 +549,7 @@ class S3File(object):
         '''
         Read and return a line from the stream.
         
-        If size is specified, at most size bytes will be read.
+        If length is specified, at most size bytes will be read.
         '''  
         self._fetch(self.loc, self.loc+1)
         while True:
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -545,6 +545,35 @@ class S3File(object):
         self.loc = nloc
         return self.loc
 
+    def readline(self, length=-1):
+        '''
+        Read and return a line from the stream.
+        
+        If size is specified, at most size bytes will be read.
+        '''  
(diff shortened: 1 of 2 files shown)


s3fs/core.py
```diff
@@ -545,6 +545,35 @@ def seek(self, loc, whence=0):
...
+        '''
+        Read and return a line from the stream.
+        
+        If size is specified, at most size bytes will be read.
+        '''  
+        self._fetch(self.loc, self.loc+1)
+        while True:
+            found = self.cache[self.loc-self.start:].find(b'\n') + 1
+            if length > 0 and found > length:
+                return self.read(length)
```
TomAugspurger: Much simpler to reuse the logic in `.read`, nice!

s3fs/core.py
```diff
@@ -545,6 +545,35 @@ def seek(self, loc, whence=0):
         self.loc = nloc
         return self.loc
 
+    def readline(self, length=-1):
+        '''
+        Read and return a line from the stream.
+        
+        If size is specified, at most size bytes will be read.
```
TomAugspurger: Maybe want to change `size` to `length`. I copied that from the `io` docs, which use `size` for the positional argument name, but s3fs uses `length`.

s3fs/core.py
```diff
@@ -652,6 +681,18 @@ def close(self):
                 self.s3.s3.put_object(Bucket=self.bucket, Key=self.key)
             self.s3._ls(self.bucket, refresh=True)
 
+    def readable(self):
```
jcrist: Would these be better as properties?

TomAugspurger: They have to be callable to be compatible with the `io` base classes (which surprised me that they weren't properties).

PR #20, opened 2016-03-28
Append wretry merge

martindurant:
Adds retries on write: raises after 10 tries by default. Recommend to catch that, and truncate at that point (self f.buffer to empty io.BytestIO, and call close()), could be done automatically.

Marge does multi-part copying on the server side, to allow splitting of big file and parallel upload.

Append mode either makes first chunk be the existing file (if size>=5MB) or downloads existing data if small. This means that the original file will *briefly* not exist.

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -355,6 +355,22 @@ class S3FileSystem(object):
         self.copy(path1, path2)
         self.rm(path1)
 
+    def merge(self, path, filelist):
+        """ Create single file from list of files
+
+        Uses multi-part, no data is downloaded.
+        """
+        bucket, key = split_path(path)
+        mpu = self.s3.create_multipart_upload(Bucket=bucket, Key=key)
+        out = [self.s3.upload_part_copy(Bucket=bucket, Key=key, UploadId=mpu['UploadId'],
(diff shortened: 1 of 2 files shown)


jcrist:
Looks like this is on top of #19 (not sure if intended). Overall this looks great to me. It might be nice to optionally apply the spit and merge operations in a fluid interface when writing (perhaps `s3.write(data, multipart=True)`), but that could come later if needed.

martindurant:
What would you expect `multipart=True` to do? 

At the moment, data is uploaded whenever the latest write took the internal buffer over the block+size (minimum 5MB, default 5MB).

jcrist:
Not sure. In a more general sense I was wondering if there may be a way to succinctly spell "take this large chunk of bytes and upload it in parts, then merge on the server".  This may be out of scope for s3fs though.

martindurant:
I think that belongs in `distributed.s3`. From a single thread, I don't see much advantage - uploads are already chunks merged on the server. 

I suppose if the S3File were pickleable, a write process could be restarted sometime later. It appears to be pickleable as is, I haven't checked yet whether that's enough.

mrocklin:
Generally this looks fine to me.

PR #21, opened 2016-03-28
ls using pagination

### LINKED ISSUE #17: Bug involving `ls` on a bucket with many Keys
I'm unable to read a CSV file that I know exists. It looks like it has to do with `ls` not finding the file, because there are many files above it, and `list_objects` returns a limit of 1000 objects.

```python
In [1]: import s3fs

In [2]: fs = s3fs.S3FileSystem(anon=False)

In [3]: df = pd.DataFrame(np.arange(20).reshape(-1, 2))

In [4]: with fs.open('vehcuity/a/b.csv', 'wb') as f:
   ...:     f.write(df.to_string().encode('utf8'))
   ...:

In [5]: import boto3

In [7]: client = boto3.client('s3')

In [8]: obj = client.get_object(Bucket='vehcuity', Key='test.csv')  # it exists
In [10]: with fs.open('vehcuity/a/b.csv', 'rb') as f:
    ...:     print(f.read())
 ... #OSError, file wasn't found
```

Looking at the traceback,

```python
ipdb> len(self._ls(path))
1000
```

It seems that `list_objects` only fetches the first 1000 keys. http://boto3.readthedocs.org/en/latest/reference/services/s3.html#S3.Client.list_objects, they suggest

> You can use the request parameters as selection criteria to return a subset of the objects in a bucket.

As a partial solution, we could filter on the subdirectory with `Prefix='a/'`. That could still fail if the `'a'` "directory" has more that 1,000 keys in it.

martindurant:
Test shows that this is very slow - or at least, calling ls on the bucket
every time a new file is created is very slow.

Fixes #17  (but not the final solution, I suspect)

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -491,8 +491,8 @@ def no_refresh(s3fs):
 
     Usage
     -----
-    >>> with no_refresh(s3fs) as fs:
-            [fs.touch('mybucket/file%i'%i) for i in range(1500)]
+    >>> with no_refresh(s3fs) as fs:    # doctest: +SKIP
+            [fs.touch('mybucket/file%i'%i) for i in range(1500)] # doctest: +SKIP
     """
     s3fs.refresh_on()
     try:
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -194,7 +194,11 @@ class S3FileSystem(object):
                 self.dirs[''] = files
             else:
                 try:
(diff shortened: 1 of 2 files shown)


martindurant:
Build sometimes times out for py2 - shows how slow calling `_ls` when creating 1200 keys. Perhaps should have an option to turn off `refresh=True` temporarily while creating/changing files.

PR #26, opened 2016-04-09
Fix mistery big

martindurant:
in S3File._fetch, caused missing bytes at the end of a file when
reading blocks big compared to the file blocksize.


mrocklin:
Great

PR #27, opened 2016-04-10
Direct info

martindurant:
Fixes for pandas interoperability

- `S3File.info()` to call client.head_object rather than going through S3FileSystem's info, which calls client.list_objects; 
the latter will fail if some keys cannot be accessed. Thus, can still open files where we know the key beforehand. This
means that the two `info()` methods do not do wuite the same thing.
- add checks for ParamValidationError, which can occur (instead of ClientError) for illegal bucket names


TomAugspurger:
I think this looks great, thanks.

PR #28, opened 2016-04-14
PR: Allow creation of temporary credentials

martindurant:
This is to enable passing around credentials inside distributed
without using the raw key/secret from the user

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -98,10 +98,11 @@ class S3FileSystem(object):
     connect_timeout = 5
     read_timeout = 15
 
-    def __init__(self, anon=None, key=None, secret=None, **kwargs):
+    def __init__(self, anon=None, key=None, secret=None, token=None, **kwargs):
         self.anon = anon
         self.key = key
         self.secret = secret
+        self.token = token
         self.kwargs = kwargs
         self.dirs = {}
         self.no_refresh = False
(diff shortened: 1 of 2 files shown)


s3fs/core.py
```diff
@@ -137,16 +139,39 @@ def connect(self, refresh=False):
...
+        if self.anon:
+            return {'anon': True}
+        if self.token:  # already has temporary cred
+            return {'key': self.key, 'secret': self.secret, 'token': self.token,
+                    'anon': False}
+        sts = boto3.Session(self.key, self.secret, self.token,
+                            **self.kwargs).client('sts')
+        cred = sts.get_session_token(DurationSeconds=3600)['Credentials']
+        return {'key': cred['AccessKeyId'], 'secret': cred['SecretAccessKey'],
+                'token': cred['SessionToken'], 'anon': False}
```
mrocklin: Should we cache the `token` or `secret` onto self after calling this?

martindurant: Plausibly, but perhaps we want new temporary credentials every time, and of course they are designed to expire (so we would have to at least check for expiry).

mrocklin: OK, presumably we call this infrequently anyway, right?  (i.e., less than once a second)

martindurant: Should be called just once in the main thread when launching a parallel distributed s3 job. An alternative to call even less frequently is to set `s3delegated = S3FileSystems(**s3. get_delegated_s3pars)` and then only use s3delegated.

mrocklin:
Should we merge?  Distributed tests are failing without this.

PR #33, opened 2016-04-27
Default S3FileSystem

mrocklin:
This adds a classmethod, `S3FileSystem.current()` which returns the most recently created `S3FileSystem` object.  It uses this method within `S3Map` to allow behavior like the following without explicitly creating an s3 object.

```python
d = S3Map('bucketname/directory')
```

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -95,6 +95,7 @@ class S3FileSystem(object):
     b'Hello, world!'
     """
     _conn = {}
+    _singleton = [None]
     connect_timeout = 5
     read_timeout = 15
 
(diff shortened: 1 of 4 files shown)


mrocklin:
Any thoughts on this @martindurant ?

martindurant:
Is there a use case of this for S3FileSystem alone? If not, could be stored in the Map class instead.

Instantiating without arguments currently creates a new instance with default connection (anon or not). I don't think we want to change that.
Does keeping the instance around in the class risk passing it and key/password to distributed workers?

mrocklin:
I'm happy dropping the last commit with `current`.  I often find myself wanting to refer to the standard `S3FileSystem` lying around and build one with defaults hoping that it will be the same.  No strong preference here.

Interestingly it looks like class attributes are not sent along with pickle.

```python
In [1]: class F(object):
    L = []
   ...:     

In [2]: f = F()

In [3]: import pickle

In [4]: pickle.dumps(f)
Out[4]: b'\x80\x03c__main__\nF\nq\x00)\x81q\x01.'

In [5]: F.L.append('hello')

In [6]: pickle.dumps(f)
Out[6]: b'\x80\x03c__main__\nF\nq\x00)\x81q\x01.'
```

mrocklin:
OK on the earlier two commits?

martindurant:
OK to keep it all, will merge unless I hear otherwise.

mrocklin:
You seem to take on most of the maintenance on s3fs so I generally defer to you.  I like the first commit.  No strong opinion on the others.

PR #37, opened 2016-05-04
PR: Expose use_ssl option for S3 client

martindurant:
For small writes can make a decent speed improvement to turn SSL off, reducing connection overhead.

Also, fix error on Mapping's clear(): the "root" shouldn't need to exist


mrocklin:
Looks good to me.  +1

PR #39, opened 2016-05-12
Ensure s3fs works with byte arrays

mrocklin:
See https://github.com/alimanfoo/zarr/pull/22

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -12,7 +12,7 @@ import boto3.s3.transfer as trans
 from botocore.exceptions import ClientError, ParamValidationError
 from botocore.client import Config
 
-from .utils import read_block, raises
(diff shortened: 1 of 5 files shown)


mrocklin:
@alimanfoo it looks like `array.array` doesn't support the buffer interface in 27.  Thoughts on the best way to handle this?

alimanfoo:
I forgot about that. I have to work around that in Zarr too. I'd suggest adding a check on entry into ``__setitem__``:

```
if PY2 and isinstance(value, array.array):
    value = value.tostring()
```

martindurant:
Looks good to me - all is good on the zarr side?

mrocklin:
I think we're good.  Merging.

PR #41, opened 2016-05-13
PR: Add bulk delete

### LINKED ISSUE #40: Fast clear
Clearing many keys from a single bucket has become a somewhat frequent operation.  Currently we do this by cycling through all of the keys one-by-one and removing them.  Is there a bulk remove-many or remove-bucket operation in boto3 that we can do instead ?

martindurant:
bulk delete takes a list of keys to remove, which must be in the same bucket.
rm, recursive, calls this instead of iterating over individual files, will
be much faster in the case of many files.

Fixes #40  - particularly useful for `clear()` on mappings.

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -454,6 +454,32 @@ class S3FileSystem(object):
             raise IOError('Copy failed', (path1, path2))
         self.invalidate_cache(buc2)
 
+    def bulk_delete(self, pathlist):
+        """
+        Remove multiple keys with one call
+
+        Parameters
+        ----------
+        pathlist : listof strings
+            The keys to remove, must all be in the same bucket.
+        """
+        buckets = {split_path(path)[0] for path in pathlist}
+        if len(buckets) > 1:
+            raise ValueError("Bulk delete files should refer to only one bucket")
+        bucket = buckets.pop()
+        if len(pathlist) > 1000:
+            for i in range((len(pathlist) // 1000) + 1):
+                print(i)
+                self.bulk_delete(pathlist[i*1000:(i+1)*1000])
+            return
+        delete_keys = {'Objects': [{'Key' : split_path(path)[1]} for path
+                                   in pathlist]}
+        try:
+            self.s3.delete_objects(Bucket=bucket, Delete=delete_keys)
+            self.invalidate_cache(bucket)
+        except ClientError:
+            raise IOError('Bulk delete failed')
+
     def rm(self, path, recursive=False):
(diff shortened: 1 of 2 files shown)


s3fs/core.py
```diff
@@ -454,6 +454,32 @@ def copy(self, path1, path2):
...
+        """
+        buckets = {split_path(path)[0] for path in pathlist}
+        if len(buckets) > 1:
+            raise ValueError("Bulk delete files should refer to only one bucket")
+        bucket = buckets.pop()
+        if len(pathlist) > 1000:
+            for i in range((len(pathlist) // 1000) + 1):
+                print(i)
+                self.bulk_delete(pathlist[i*1000:(i+1)*1000])
+            return
```
mrocklin: You may be interested in [toolz.partition_all](http://toolz.readthedocs.io/en/latest/api.html#toolz.itertoolz.partition_all)

s3fs/core.py
```diff
@@ -454,6 +454,32 @@ def copy(self, path1, path2):
...
+        delete_keys = {'Objects': [{'Key' : split_path(path)[1]} for path
+                                   in pathlist]}
+        try:
+            self.s3.delete_objects(Bucket=bucket, Delete=delete_keys)
+            self.invalidate_cache(bucket)
+        except ClientError:
+            raise IOError('Bulk delete failed')
```
mrocklin: I've found it pleasant to pass through the exceptions from boto3.  They tend to have more information about why things failed.  I see two options:

1.  Skip the try-except and let the exception bubble up.  This is simple but can result in users seeing confusing error messages
2.  Include the exception in the error that we produce

mrocklin:
Looks good to me

s3fs/core.py
martindurant: Was this implemented anywhere that I didn't notice? Should be done in a different PR.

s3fs/core.py
martindurant: I knew there was a more elegant way to do this, but somehow couldn't wrap it into my head; also, we iterate over the paths list more than once, would be tidier. Still, this works.

s3fs/core.py
mrocklin: Ah, I had it in local work.  I'll push up something for review.

s3fs/core.py
mrocklin: Yeah, not suggesting a change.  Just pointing out a function I find I use fairly often.

PR #42, opened 2016-05-13
Various small changes

mrocklin:
1.  Create a directory for mapping if it doesn't exist.
2.  Include original exception error messages

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -645,8 +645,8 @@ class S3File(object):
                 else:
                     try:
                         self.mpu = s3.s3.create_multipart_upload(Bucket=bucket, Key=key)
-                    except (ClientError, ParamValidationError):
-                        raise IOError('Open for write failed', path)
+                    except (ClientError, ParamValidationError) as e:
+                        raise IOError('Open for write failed', path, e)
                     self.loc = self.size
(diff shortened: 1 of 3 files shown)


martindurant:
S3 keys don't really need a "directory key" to live within. The empty key wouldn't hurt, I suppose.

martindurant:
The problem is with recursive rm: if there are not actually any files below to pass to bulk_delete.

mrocklin:
I ran into this when creating a mapping from a non-existent bucket, in which case the mkdir does become necessary (I think).

martindurant:
Ah, so it's not the directory you want to create per se, but the bucket for it to live within. I'm not sure we should do that automatically.

mrocklin:
What would you recommend?  Improved error message?

martindurant:
Error message I think, yes. Either happens at check time or on first write when `check=False`.

mrocklin:
Thoughts on adding a `create=False` keyword argument to create the bucket if necessary and having the error message point them towards that keyword argument?  This seems significantly more convenient and helps mapping users (e.g. zarr users) from having to be exposed to the S3FileSystem API.

martindurant:
OK, fine with me. The whole mapping thing of course is for zarr's convenience. Most users of s3fs itself probably won't touch it.

Issue #38, opened 2016-05-11
Can't list buckets


jseabold:
I have some AWS credentials for which I can't list buckets. I almost never have bucket level permissions, but I do have permissions on prefixes. So something like

```
fs = s3fs.S3FileSystem(profile_name='user')
fs.ls('s3://home/sseabold/prefix')
```

Won't work because I can't list any buckets (S3Client.list_buckets fails), and I also don't generally have permissions at the bucket level.

Similar issues to those here [1, 2] for the default (old) boto behavior.

[1] https://github.com/conda/conda/pull/2126
[2] https://github.com/blaze/odo/pull/448

martindurant:
I don't yet have a solution for this, we might indeed have to change the way that `ls` works. 
In the meantime, you can still access details of individual files using `info()` and open them.

jseabold:
It looks like `info` has the same issue with bucket permissions.

martindurant:
Apologies, seems we only implemented the direct `info()` for the `S3File` object, i.e,. one you have already opened. Would `info()` on a given file, sidestepping the `ls` solve this problem for you?

jseabold:
Ah, I suppose. I was just looking at `info` for the first time. My main use is for list. No worries. s3 ACLs are a pain.

jseabold:
It seems that all of my errors that I've dug in to are coming down to things that assume I can list the bucket. The most recent `dask` error is because of the `list_objects` paginator [1]. It tries to list the objects on the bucket rather than the prefix that I asked for in `dask.DataFrame.read_csv`.

I don't have my head completely around things here yet and the paginator stuff is new to me, but my files are in a bucket that has a *lot* of objects and I'm not interested in any of them. I only want the keys in the prefix I'm asking for, which I have permissions for and there are only a few. The dask error comes because I called `read_csv` on an object from a single key and it's listing the whole bucket AFAICT.

https://github.com/dask/s3fs/blob/master/s3fs/core.py#L256

jseabold:
I wonder if you shouldn't also pop out the prefix from the path in the `S3FileSystem._ls` and pass that along for pagination. I assume this is why all of the AWS SDK stuff is very particular about trailing slashes to identify prefixes vs keys.

Happy to test any changes.

martindurant:
It is a valid model.
[...]

PR #44, opened 2016-05-13
PR: When deleting bucket, invalidate list cache for ''

### LINKED ISSUE #43: Delete bucket does not invalidate ls cache of buckets
```python
>>> s3.ls('')
['temp']
>>> s3.rm('temp')
>>> s3.ls('')
['temp']
(should be [])
```

martindurant:
Otherwise bucket name would still show up for `ls('')`.

Fixes #43

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -361,7 +361,7 @@ class S3FileSystem(object):
         if split_path(path)[1]:
             return bool(self.ls(path))
         else:
-            return (path in self.ls('') and
+            return (path in self.ls('') or
                     not raises(FileNotFoundError, lambda: self.ls(path)))
 
     def cat(self, path):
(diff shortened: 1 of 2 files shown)


martindurant:
Related question: currently `exists()` returns `False` for any bucket public bucket we can see but don't own (i.e,. all buckets, when `anon=True` in the credentials), even if we've already accessed it. 
One solution would be to to change the `and` to an `or` in the last line of `exists` (so we actually check if the bucket is readable), or instead to check `dirs` only for buckets already accessed. @mrocklin , any thoughts about this and the `ls` issue #38 ? Changes probably belong in this PR.

mrocklin:
I think that we should return true for files that we can see but down own, especially if we have read privileges.  I would generally follow posix rules for this sort of thing.

PR #52, opened 2016-05-28
Add bucket name to non-existent bucket error

mrocklin:


martindurant:
OK, looks fair enough.

PR #56, opened 2016-06-16
added process ID to connection key

mheilman:
fixes https://github.com/dask/dask/issues/1292

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -1,9 +1,9 @@
 # -*- coding: utf-8 -*-
 import io
 import logging
+import os
 import re
 import socket
-import threading
 from hashlib import md5
 
 import boto3
@@ -138,10 +138,9 @@ class S3FileSystem(object):
         anon, key, secret, kwargs, token, ssl = (self.anon, self.key,
                             self.secret, self.kwargs, self.token, self.use_ssl)
 
-        # Include the current thread ID in the connection key so that different
-        # SSL connections are made for each process or thread.
-        tok = tokenize(anon, key, secret, kwargs, token, ssl,
-                       threading.get_ident())
+        # Include the current PID in the connection key so that different
+        # SSL connections are made for each process.
+        tok = tokenize(anon, key, secret, kwargs, token, ssl, os.getpid())
         if refresh:
             self._conn.pop(tok, None)
         logger.debug("Open S3 connection.  Anonymous: %s", self.anon)
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -3,8 +3,8 @@ import io
 import logging
 import re
 import socket
+import threading
 from hashlib import md5
-from contextlib import contextmanager
 
 import boto3
 import boto3.compat
@@ -137,7 +137,11 @@ class S3FileSystem(object):
         """
         anon, key, secret, kwargs, token, ssl = (self.anon, self.key,
                             self.secret, self.kwargs, self.token, self.use_ssl)
-        tok = tokenize(anon, key, secret, kwargs, token, ssl)
+
+        # Include the current thread ID in the connection key so that different
+        # SSL connections are made for each process or thread.
+        tok = tokenize(anon, key, secret, kwargs, token, ssl,
+                       threading.get_ident())
         if refresh:
             self._conn.pop(tok, None)
         logger.debug("Open S3 connection.  Anonymous: %s", self.anon)
--- a/s3fs/tests/test_s3fs.py
+++ b/s3fs/tests/test_s3fs.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+from concurrent.futures import ProcessPoolExecutor
 import io
 import pytest
 from itertools import chain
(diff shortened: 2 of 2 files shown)


mheilman:
Though I've only observed the SSL problems with multiprocessing, it seems safer to just put the thread ID in the connection ID/key/token.

s3fs/core.py
```diff
@@ -3,8 +3,8 @@
 import logging
 import re
 import socket
+import threading
 from hashlib import md5
-from contextlib import contextmanager
```
mheilman: this was an unused import

mrocklin:
`os.get_ident` may still be safer.

```python
In [1]: from multiprocessing import Pool

In [2]: p = Pool()

In [3]: from threading import get_ident

In [4]: import os

In [5]: p.apply_async(os.getpid).get()
Out[5]: 29219

In [6]: p.apply_async(os.getpid).get()
Out[6]: 29220

In [7]: p.apply_async(os.getpid).get()
Out[7]: 29221

In [8]: p.apply_async(os.getpid).get()
Out[8]: 29222

In [9]: p.apply_async(get_ident).get()
Out[9]: 140167402886976

In [10]: p.apply_async(get_ident).get()
Out[10]: 140167402886976

In [11]: p.apply_async(get_ident).get()
Out[11]: 140167402886976

In [12]: p.apply_async(get_ident).get()
Out[12]: 140167402886976
```

mrocklin:
Forking is strange

mheilman:
whoa, yeah, that's weird.  i'll change it back to `os.getpid`.  That should work in python 2 also.

(you meant "`os.getpid` may still be safer", right?)

martindurant:
OK, looks reasonable like this. So threads within a process probably do share the socket, right? That doesn't appear to cause a problem, but worth keeping in mind.

Is anyone testing multiprocessing on windows? I ask because windows uses spawn rather than fork for new processes.

mheilman:
> So threads within a process probably do share the socket, right? 

Yes, probably.  I could *also* include `threading.get_ident()` (or whatever the 2.7 equivalent is) just in case, if you think that'd be safer.

> Is anyone testing multiprocessing on windows?

I haven't, and I don't really have an easy way to do so.

mrocklin:
Because Windows spawns rather than forks it's usually safer (though slower) than unix-based systems.  I wouldn't worry too much here.

mrocklin:
Does anything remain to be done here?

Can people think of a decent test to ensure that people keep around this feature in the future?

mheilman:
> Does anything remain to be done here?

I don't think so.

btw, a colleague of mine suggested that the reason problems only showed up with processes might be a lock being copied.  That is, a lock might be ensuring only one thread within the same process can use the s3 connection at a time, but new processes might make copies of the lock so they could all try and use the connection simultaneously.  I'm not sure that's it, but I haven't bothered going down the rabbit hole further.

>
> Can people think of a decent test to ensure that people keep around this feature in the future?

How about this?

```
def _get_s3_id(s3):
        return id(s3.connect())


def test_no_connection_sharing_among_processes(s3):
    executor = ProcessPoolExecutor()
    conn_id = executor.submit(_get_s3_id, s3).result()
    assert id(s3.connect()) != conn_id, \
        "Processes should not share S3 connections."
```

mrocklin:
Looks great.  Can I ask you to add that into `s3fs/tests/test_s3fs.py` and perhaps add a line `pytest.importorskip('concurrent.futures')` within the test just so that it skips on Python 2 if concurrent.futures isn't around.  I think we can merge afterwards.

mheilman:
I added the test.  look OK?

s3fs/tests/test_s3fs.py
```diff
@@ -661,3 +662,15 @@ def test_array(s3):
     with s3.open(a, 'rb') as f:
         out = f.read()
         assert out == b'A' * 1000
+
+
+def _get_s3_id(s3):
+        return id(s3.connect())
+
+
+pytest.importorskip('concurrent.futures')
```
mrocklin: This line needs to be within the test.  If you move it down two lines and indent it all should be well

s3fs/tests/test_s3fs.py
```diff
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+from concurrent.futures import ProcessPoolExecutor
```
mrocklin: Oh, we'll probably want to move this import within the test as well

s3fs/tests/test_s3fs.py
mheilman: ah, oops!

mheilman:
Actually, I don't think we need to skip that test in python 2. This depends on `boto3`, which depends on the `futures` [backport](https://pypi.python.org/pypi/futures).  I just tried installing `s3fs` in a local 2.7 environment, and the tests ran fine without the `importorskip` call.  Am I missing something?

mrocklin:
Ah, I didn't realize that we were implicitly depending on concurrent.futures.  My mistake.  Please ignore my previous comments.

mheilman:
> Ah, I didn't realize that we were implicitly depending on concurrent.futures. My mistake. Please ignore my previous comments.

Well, I completely didn't think about breaking things for 2.7 users, so it's good that somebody did. :-)

mrocklin:
Merged.  Thanks @mheilman !

PR #62, opened 2016-08-04
Pin boto3 version to <1.4.0.

rolando:
Latest release of boto3 (a few days ago) breaks ``s3fs`` due to the
imports ``boto3.s3.transfer.{ReadTimeoutError, IncompleteReadError}``
being gone.

See:
https://github.com/boto/boto3/commit/d02dc9b57ae28ccb2265dcf1e560afd7dbad8098


mrocklin:
Thanks for the fix.  Do you have any insight on how s3fs should be changed in the future to support `boto3>=1.4.0`?

rolando:
@mrocklin I haven't dig further as I stumble upon this bug while fixing another in ``dask`` :)

It seems they moved the retry handling to the ``TransferManager``  class. **Perhaps** the retry logic in ``s3fs`` is no longer needed.

I think ``s3fs`` could opt to either:

* Support ``1.3.x`` and ``1.4x`` by adding backward compatible code.
* Require ``boto3>=1.4.0`` in further releases.

I don't know what would be the best for the ``s3fs`` user base (or ``dask`` user base for that matter).

mrocklin:
At first glance my recommendation would be to do the following

```python
try:
    from boto.s3.transfer import S3_RETRYABLE_ERRORS
except ImportError:
    S3_RETRYABLE_ERRORS = {socket.timeout, # whatever reasonable errors we can find}
```

Either the `except` clause is entirely unnecessary (if as @rolando suggests may be the case) or we'll offer less protection for boto 1.4 and learn about that loss of protection as users complain about errors.

cc @martindurant in case he has thoughts.

mrocklin:
Merging this for now.

PR #64, opened 2016-08-05
Try-except around importing S3_RETRYABLE_ERRORS

mrocklin:
cc @martindurant thoughts?


s3fs/core.py
```diff
@@ -18,10 +17,14 @@
...
-S3_RETRYABLE_ERRORS = (
-    socket.timeout,
-    trans.ReadTimeoutError, trans.IncompleteReadError
-)
+
+try:
+    from boto3.s3.transfer import S3_RETRYABLE_ERRORS
+except ImportError:
+    S3_RETRYABLE_ERRORS = (
+        socket.timeout,
```
martindurant: This is the problem, I don't know which exceptions, and whether boto chooses to wrap some of them

mrocklin: Agreed.  For the time being I suggest that we merge this and see what exceptions arise that we need to add here.  

It could also be that boto3 has gotten better about retreis.

mrocklin: This issue is currently halting all dask/distributed PRs, so I'm fairly motivated to get *something* in soon.

martindurant: OK; well I don't trust that boto necessarily did anything good, but I do trust that actual transfer errors when talking with s3 are rare (unless your local connection goes down, in which case retries won't help).

Didn't pinning the version already solve the test failures?

mrocklin: Sadly not.  We could do more pinning in travis.yml, but the issue is still live.

I still think that transfer errors are serious (even though rare) but I think that this fix is more likely to make us aware of them more quickly.

martindurant: So, merge... and watch this space.
I should have time next week to look into it, if there are no reports between now and then.

mrocklin:
Merged.  Thanks @martindurant

PR #71, opened 2016-10-14
Allow parts of multipart upload to be equal to the blocksize

scizo:
When S3FileSystem.put is used, because chunks are written to
S3File.write in chunks `blocksize` large this results in multipart
uploads that are twice as large, because it takes two writes in order to
surpass `blocksize` number of bytes in the buffer. This change will
cause it to send parts that are equal to `blocksize` when using put.

I have had consistent problems with connection timeouts to s3 with
larger files. With this simple change, I have uploaded many large files
with no problems. I also plan to submit a pull request that allows you
to override the blocksize from S3FileSystem APIs. I hope this is welcome.

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -863,7 +863,7 @@ class S3File(object):
         """
         Write data to buffer.
 
-        Buffer only sent to S3 on flush() or if buffer is bigger than blocksize.
+        Buffer only sent to S3 on flush() or if buffer is greater than or equal to blocksize.
 
         Parameters
         ----------
@@ -876,7 +876,7 @@ class S3File(object):
             raise ValueError('I/O operation on closed file.')
         out = self.buffer.write(ensure_writable(data))
         self.loc += out
-        if self.buffer.tell() > self.blocksize:
+        if self.buffer.tell() >= self.blocksize:
             self.flush()
         return out
 
--- a/s3fs/tests/test_s3fs.py
+++ b/s3fs/tests/test_s3fs.py
@@ -743,3 +743,25 @@ def test_upload_with_s3fs_prefix(s3):
 
     with s3.open(path, 'ab') as f:
         f.write(b'b' * (10 * 2 ** 20))
+
+def test_multipart_upload_blocksize(s3):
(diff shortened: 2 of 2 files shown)


mrocklin:
cc @martindurant

mrocklin:
Cool work @scizo .  Is there a convenient way to test the issues you were seeing earlier?  It would be good to set up a test so that if a future developer reverts this back they'll get a failed test.

scizo:
I added a test. I was going to add one originally, but didn't see where they were located. I guess I should have looked a little harder. :smile:

This definitely tests against double blocksize parts for the multipart upload, but given that it doesn't use `put` directly, it wouldn't necessarily catch changes to `put` that somehow effected this. It does demonstrate the change made in `S3File` though. Let me know if this isn't the type of test you were expecting or if you would prefer some changes to it.

mrocklin:
The style looks fine to me.  I'll let Martin be the judge of the substance here :)

martindurant:
Hi @scizo , thanks for bringing this up.
What you have done looks totally fine to me.
I'm still a little confused about why you were getting more parts to the upload with the previous version, but since this solves it for you, I'm happy to accept it.

mrocklin:
Thanks @scizo !

scizo:
@martindurant The problem wasn't getting more parts to the upload, but not enough, making each of the parts bigger and therefor much more prone to connection timeouts.

The default blocksize is 5 Mb. So when S3FileSystem.put is called it writes 5 Mb chunks to the S3File instance. After the first chunk is written the following check was performed on line 879

```python
if self.buffer.tell() > self.blocksize:
```

Because `self.buffer.tell()` returns 5 Mb and `self.blocksize` equals 5 Mb this condition is false meaning flush isn't called until after another write of 5 Mb, resulting in the part of the upload to be 10 Mb. With the change, the condition is true after a single 5 Mb write causing each part to be 5 Mb instead of 10 Mb.

I hope that was a better description of why this helps.

Thanks for the library and accepting the pull-request!

martindurant:
OK, now it makes sense ;)

PR #72, opened 2016-10-28
Add fill_cache option to S3File

martindurant:
If False (not the default), skipping to later in the file beyond the read-ahead,
will not read to fill the in-between. This is appropriate for reading chunks in
the order they occur in the file, but skipping large chunks, e.g., for specific
columns spread through a parquet file.

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -214,7 +214,8 @@ class S3FileSystem(object):
         self.__dict__.update(state)
         self.s3 = self.connect()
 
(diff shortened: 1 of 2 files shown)


mrocklin:
Are there cases where we want `fill_cache=True`?

martindurant:
For more random-access-like read patterns (back and forth), probably. I didn't want to make `False` the default (yet), because it changes the behaviour from previous.

mrocklin:
Should this be merged?

PR #77, opened 2016-12-14
ENH: add support for requester pays buckets

tacaswell:
Only read-operations are supported

Ran into this trying to access arXiv data (https://arxiv.org/help/bulk_data_s3) for an example of this in the wild.

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -109,6 +109,22 @@ Furthermore, `S3FileSystem.current()` will return the most-recently created
 instance, so this method could be used in preference to the constructor in
 cases where the code must be agnostic of the credentials/config used.
 
+Requester Pays Buckets
+----------------------
+
+Some buckets, such as the `arXiv raw data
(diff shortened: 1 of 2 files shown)


mrocklin:
I don't suppose there is a way to mock this out with moto?

mrocklin:
(also awesome, thanks for doing this!)

tacaswell:
`git grep -i payer` on the moto code base does not turn anything up.  

Due to the way RequesterPays works, you _must_ authenticate (so they know who to charge).   To test this you would have to set up moto buckets with interesting ACL, but it seems to not mock those out (tested by setting 'ACL' in the fixture to jibberish or the value of RequestPayer to jibberish and the tests still pass).

mrocklin:
In that case I'm +1 to merge this. 

Merging soon if no further comments.

martindurant:
Thanks for doing this!
Any chance we could have a sentence or two in the docs on when to use this (perhaps with your arxiv example).

PR #82, opened 2017-02-07
Add support for metadata via {set,get}xattr calls

tigermonkey:
Hi, 

Thanks for this project. It makes working with S3 so much simpler.

Could you add support for S3 metadata? Here is an initial implementation, where I tried to mimic the [xattr Linux interface][1]. Let me know if I should change something.

Thanks!
Pablo

[1]: https://manpages.debian.org/jessie-backports/manpages/xattr.7.en.html

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -364,6 +364,74 @@ class S3FileSystem(object):
             except (ClientError, ParamValidationError):
                 raise FileNotFoundError(path)
 
+    _metadata_cache = {}
+
+    def metadata(self, path, refresh=False):
+        """ Return metadata of path.
+
+        Metadata is cached unless `refresh=True`.
+
+        Parameters
+        ----------
+        path : string/bytes
+            filename to get metadata for
+        refresh : bool (=False)
+            if False, look in local cache for file metadata first
+        """
+        bucket, key = split_path(path)
+
+        if refresh or path not in self._metadata_cache:
+            response = self.s3.head_object(Bucket=bucket,
+                                           Key=key,
+                                           **self.req_kw)
+            self._metadata_cache[path] = response['Metadata']
+
+        return self._metadata_cache[path]
+
+    def getxattr(self, path, attr_name):
+        """ Get an attribute from the metadata.
+
+        Examples
+        --------
+        >>> mys3fs.getxattr('mykey', 'attribute_1')  # doctest: +SKIP
+        'value_1'
+        """
+        xattr = self.metadata(path)
+        if attr_name in xattr:
+            return xattr[attr_name]
+        return None
+
+    def setxattr(self, path, **kw_args):
+        """ Set metadata.
+
+        Attributes have to be of the form documented in the `Metadata Reference`_.
+
+        Examples
+        --------
+        >>> mys3file.setxattr(attribute_1='value1', attribute_2='value2')  # doctest: +SKIP
+
+        .. Metadata Reference:
+        http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html#object-metadata
+        """
+
+        bucket, key = split_path(path)
+        metadata = self.metadata(path)
+        metadata.update(**kw_args)
+
+        # remove all keys that are None
+        for kw_key in kw_args:
+            if kw_args[kw_key] is None:
+                metadata.pop(kw_key, None)
+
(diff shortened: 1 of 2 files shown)


s3fs/core.py
```diff
@@ -364,6 +364,74 @@ def info(self, path, refresh=False):
...
+                metadata.pop(kw_key, None)
+
+        self.s3.copy_object(CopySource="{}/{}".format(bucket, key),
+                            Bucket=bucket,
+                            Key=key,
+                            Metadata=metadata,
+                            MetadataDirective='REPLACE')
+
+        # refresh metadata
+        self.metadata(path, refresh=True)
```
martindurant: If the operation succeeded, can we not assume that the metadata was correctly updated, rather than refreshing?

s3fs/core.py
```diff
@@ -364,6 +364,74 @@ def info(self, path, refresh=False):
...
+        # remove all keys that are None
+        for kw_key in kw_args:
+            if kw_args[kw_key] is None:
+                metadata.pop(kw_key, None)
+
+        self.s3.copy_object(CopySource="{}/{}".format(bucket, key),
+                            Bucket=bucket,
+                            Key=key,
+                            Metadata=metadata,
+                            MetadataDirective='REPLACE')
```
martindurant: Since we're being careful here about exact replication:
- does the ACL remain the same 
- I assume the etag changes

s3fs/core.py
```diff
@@ -788,6 +856,35 @@ def info(self):
...
+
+    def setxattr(self, **kwargs):
+        """ Set metadata.
+        See :func:`~s3fs.S3Filesystem.setxattr`.
+
+        Examples
+        --------
+        >>> mys3file.setxattr(attribute_1='value1', attribute_2='value2')  # doctest: +SKIP
+        """
+        return self.s3.setxattr(self.path, **kwargs)
```
martindurant: I don't suppose this can work for a file open for writing? Perhaps in that case, we need to store the metadata and apply upon close. For a file in append mode, we need to store the original metadata and apply to the final file.

s3fs/core.py
```diff
@@ -364,6 +364,74 @@ def info(self, path, refresh=False):
...
+        xattr = self.metadata(path)
+        if attr_name in xattr:
+            return xattr[attr_name]
+        return None
+
+    def setxattr(self, path, **kw_args):
+        """ Set metadata.
+
+        Attributes have to be of the form documented in the `Metadata Reference`_.
+
```
martindurant: Parameters: key-value pairs like ``field="value"``, where the values must be strings. Does not alter existing fields, unless the field appears here - if the value is None, delete the field.

s3fs/core.py
martindurant: (or simply not allow File.setxattr if not in read mode.)

s3fs/core.py
tigermonkey: Yes, but the cache is dirty. Maybe having a dirty flag would be better instead? That way we delay the request until the next time we query metadata.

martindurant: Wouldn't `self._metadata_cache[path] = metadata` correctly update the cache? 
In the case of directories, we don't mark dirty, but rather remove outdated entries, so they are updated next time.

tigermonkey: True. Will update to do that.

s3fs/core.py
tigermonkey: The ACL is 'private' by default[1]. 

The ETag will remain the same[2]:

> Returns the ETag of the new object. The ETag reflects changes only to the contents of an object, not its metadata. The source and destination ETag will be identical for a successfully copied object.

[1]: http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectCOPY.html#rest-object-put-access-permissions
[2]: http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectCOPY.html#RESTObjectCOPY-responses

martindurant: Can we ensure that the ACL is the same as the original item? A user may be surprised if their public object becomes private when they set metadata.

tigermonkey: Looks like ACL is preserved in this case.

s3fs/core.py
tigermonkey: updated docstring

martindurant:
Thank you, @tigermonkey

PR #86, opened 2017-02-25
add `config_kwargs` parameter to S3FileSystem

jgysland:
It may have other applications for other use cases, but in my case, I need to be able to pass `signature_version='s3v4'` to the `botocore.client.Config` object in order to use s3fs with [Minio](https://minio.io).

### DIFF
--- a/s3fs/tests/test_s3fs.py
+++ b/s3fs/tests/test_s3fs.py
@@ -93,6 +93,11 @@ def test_client_kwargs():
     assert s3.s3.meta.endpoint_url.startswith('http://foo')
 
 
+def test_config_kwargs():
+    s3 = S3FileSystem(config_kwargs={'signature_version': 's3v4'})
+    assert s3.connect(refresh=True).meta.config.signature_version == 's3v4'
+
+
 def test_tokenize():
     from s3fs.core import tokenize
     a = (1, 2, 3)
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -103,10 +103,10 @@ class S3FileSystem(object):
     default_block_size: None, int
         If given, the default block size value used for ``open()``, if no
         specific value is given at all time. The built-in default is 5MB.
-    default_fill_cache: Bool (True)
+    default_fill_cache : Bool (True)
         Whether to use cache filling with open by default. Refer to
         ``S3File.open``.
-    config_kwargs: dict of parameters passed to ``botocore.client.Config``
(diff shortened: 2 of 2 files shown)


mrocklin:
This seems reasonable to me.  @martindurant ?

@jgysland Is there a way to test this functionality to ensure that it sticks around through future code changes?  Perhaps there is some config parameter that you could check makes its way through the `S3FileSystem` constructor all the way to the generated config file ?

martindurant:
Seems reasonable to me.

jgysland:
I've added a test that should ensure that the parameter isn't inadvertently removed in future changes and that at least my use case functions. ;)

How do we rerun the tests?

s3fs/tests/test_s3fs.py
```diff
@@ -93,6 +93,11 @@ def test_client_kwargs():
     assert s3.s3.meta.endpoint_url.startswith('http://foo')
 
 
+def test_config_kwargs():
+    s3 = S3FileSystem(config_kwargs={'signature_version': 's3v4'})
+    assert s3.connect(refresh=True).meta.config.signature_version == 's3v4'
```
martindurant: Should this not be `s3.connect(refresh=True).s3.meta.config.signature_version` - i.e., the boto3 client's meta, not the S3FileSystem's?

martindurant:
The tests should run automatically, as they did for the previous commit. TravisCI seems to be slow now, so it may take a while.

s3fs/tests/test_s3fs.py
jgysland: Aren't they the same? `self.s3 = self.connect()` in `S3FileSystem.__init__` -- I only call the function explicitly to pass `refresh=True`, without which it uses a cache from a previous `S3FileSystem` instance returns the wrong `signature_version`.

martindurant: Right, got you - sorry.

jgysland:
It would seem like the test runner for Python 3.5 failed during (and presumably as a consequence of) yesterday's S3 outage. Travis CI is [reporting](https://www.traviscistatus.com/incidents/hmwq9yy5dh9d) that things that issues caused by the outage are resolved (after having been processing a backlog for several hours), so do we think this will run again on its own, or should I push another commit to trigger it, or is there another way to trigger it?

mrocklin:
Looks like someone restarted it.

martindurant:
All good, thank you.

Issue #89, opened 2017-04-14
Implement server-side encryption


martindurant:
See http://stackoverflow.com/questions/43359943/dask-dataframes-to-parquet-support-server-side-encryption

Any number of "file parameters" could be included in the constructor, for use in open() and other methods by default within the instance, but over-ridable in the call; or there could be just the specific SSE-related ones.

dan-coates:
+1, I just ran into this in a case where I wanted to use s3fs for writing a simple file to S3 but SSE is required in my bucket...

Seems like the `open` method would be a natural place to place that kwarg.

martindurant:
@dan-coates , would you like to experiment with #90, to see if it meets your needs? That code touches quite a few parts, and I haven't had the chance yet to trial it.

dan-coates:
I looked through the code a bit - I'm not sure how much I like the idea of cramming everything into the `S3FileSystem` constructor and filtering things down into all the methods in the background. This isn't a use case I have right now, so this may be unnecessarily hypothetical, but what if I want to write some files that are encrypted and some that aren't, or read different files encrypted with different keys? Then I have to create multiple `S3FileSystem` objects and keep track of that.

Now, maybe that's actually easier than appending additional kwargs onto the methods where I want to open the files and whatnot, I don't know. It just seems like more of a file-level option than a global filesystem option.

I also feel like it makes the internals more convoluted with all the `self._call_s3` indirection, but I probably shouldn't throw stones there unless I'm willing to try a better implementation. :)

martindurant:
@dan-coates , I tend to agree with you, and started to code out simply adding **kwargs to all methods that call s3 and as an instance variable on S3Files. However, we get into trouble for users accessing s3 via dask: they only have a route to provide arguments to the FileSystem constructor (with `storage_options=`), not to the individual open calls.

dan-coates:
Yeah, I had a feeling approaching this would get down the rabbit hole of how Dask uses the package. Not having looked at how Dask uses s3fs, couldn't it bear the responsibility of filtering options down to the methods that need them? It seems wrong to tailor what is effectively a general purpose tool to a specific use case in a way that's not intuitive for the general case.

Again, just my opinion - I would like to be able to spend some time on this but don't know if I'll get the chance...

martindurant:
[...]

PR #90, opened 2017-04-27
Support for kwargs in order to support ServerSideEncryption

mariusvniekerk:
This is a rough POC for allowing us to set various additional kwargs that get used during writing to s3.

Related to #89

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -447,7 +447,7 @@ class S3FileSystem(object):
                             Key=key,
                             Metadata=metadata,
                             MetadataDirective='REPLACE',
-                            **self.s3.writer_kwargs,
+                            **self.writer_kwargs
                             )
 
         # refresh metadata
@@ -627,7 +627,7 @@ class S3FileSystem(object):
             The paths, in order, to assemble into the final file.
         """
         bucket, key = split_path(path)
-        mpu = self.s3.create_multipart_upload(Bucket=bucket, Key=key, **self.s3.writer_kwargs)
+        mpu = self.s3.create_multipart_upload(Bucket=bucket, Key=key, **self.writer_kwargs)
         out = [self.s3.upload_part_copy(Bucket=bucket, Key=key, UploadId=mpu['UploadId'],
                             CopySource=f, PartNumber=i+1) for (i, f) in enumerate(filelist)]
         parts = [{'PartNumber': i+1, 'ETag': o['CopyPartResult']['ETag']} for (i, o) in enumerate(out)]
@@ -643,7 +643,7 @@ class S3FileSystem(object):
         try:
             self.s3.copy_object(Bucket=buc2, Key=key2,
                                 CopySource='/'.join([buc1, key1]),
-                                **self.s3.writer_kwargs)
+                                **self.writer_kwargs)
         except (ClientError, ParamValidationError):
             raise IOError('Copy failed', (path1, path2))
         self.invalidate_cache(path2)
@@ -731,7 +731,7 @@ class S3FileSystem(object):
         if key:
             if acl and acl not in key_acls:
                 raise ValueError('ACL not in %s', key_acls)
-            self.s3.put_object(Bucket=bucket, Key=key, ACL=acl, **self.s3.writer_kwargs)
+            self.s3.put_object(Bucket=bucket, Key=key, ACL=acl, **self.writer_kwargs)
             self.invalidate_cache(path)
         else:
             if acl and acl not in buck_acls:
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -72,19 +72,6 @@ buck_acls = {'private', 'public-read', 'public-read-write',
              'authenticated-read'}
 
 
-class SSEParams(object):
-
-    def __init__(self, server_side_encryption=None, sse_customer_algorithm=None,
-                 sse_customer_key=None, sse_kms_key_id=None):
-        self.ServerSideEncryption = server_side_encryption
-        self.SSECustomerAlgorithm = sse_customer_algorithm
-        self.SSECustomerKey = sse_customer_key
-        self.SSEKMSKeyId = sse_kms_key_id
-
-    def to_kwargs(self):
-        return {k: v for k, v in self.__dict__.items() if v is not None}
-
-
 class S3FileSystem(object):
     """
     Access S3 as if it were a file system.
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -131,7 +131,7 @@ class S3FileSystem(object):
     def __init__(self, anon=False, key=None, secret=None, token=None,
                  use_ssl=True, client_kwargs=None, requester_pays=False,
                  default_block_size=None, default_fill_cache=True,
-                 config_kwargs=None, writer_kwargs=None, **kwargs):
+                 config_kwargs=None, s3_additional_kwargs=None, **kwargs):
         self.anon = anon
         self.session = None
         self.key = key
@@ -150,7 +150,7 @@ class S3FileSystem(object):
         self.config_kwargs = config_kwargs
         self.dirs = {}
         self.req_kw = {'RequestPayer': 'requester'} if requester_pays else {}
-        self.writer_kwargs = writer_kwargs or {}
+        self.s3_additional_kwargs = s3_additional_kwargs or {}
         self.use_ssl = use_ssl
         self.s3 = self.connect()
         self._kwargs_helper = ParamKwargsHelper(self.s3)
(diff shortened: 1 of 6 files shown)


martindurant:
Thank you for this!
Do you know if SSE can be tested with moto? [This](https://github.com/spulec/moto/blob/master/tests/test_s3/test_s3.py) suggests it can be done.

Naturally, we will need a section in the docs describing what you can pass in these kwargs to get SSE working.

mariusvniekerk:
looks like we can make a bucket that enforces some encryption from that.

So do you think the kwargs option is sufficient or should we rather make a Security object that houses these various settings and pull from them as neccesary?

(guess the latter also makes it simpler to add a `security` optional parameter for the various write operations that a user can supply.

martindurant:
I think it's fine to leave the more general kwargs, and document what they can be used for. A user can always make a dictionary of them, which is what the security structure would look like anyway.

Do you think it's intuitive to have one s3fs instance for encrypted use, and potentially another on the same account/bucket for other, non-encrypted files; or would it be better to allow passing override parameters to open() and other methods so that a single instance can handle both?

mariusvniekerk:
The only reason i ask is that not ALL of the security params apply to every method.  

We can look up which things are valid from something like

```
model = s3.meta.service_model.operation_model("PutObject")
model.input_shape.members
```

martindurant:
I take it, that passing extra parameters causes an exception. Presumably boto is checking exactly the same thing as you have above on every call. So every method would need a set of kwargs defined to pluck from the general set, looks a bit messy, but may work.

mariusvniekerk:
Okay cleaned up the implementation a bit here.  Should be easier to add **kwargs whereever we want now.

s3fs/core.py
```diff
@@ -71,6 +71,59 @@ def split_path(path):
...
+    s3 : S3FileSystem
+    """
+
+    def __init__(self, s3):
+        self.s3 = s3
+        self._kwarg_cache = {}
+
+    def _get_valid_keys(self, model_name):
+        if model_name not in self._kwarg_cache:
+            model = self.s3.meta.service_model.operation_model(model_name)
```
mariusvniekerk: Basically this just lazily caches

s3fs/core.py
```diff
@@ -146,10 +201,23 @@ def __init__(self, anon=False, key=None, secret=None, token=None,
...
+        self.writer_kwargs = writer_kwargs or {}
         self.use_ssl = use_ssl
         self.s3 = self.connect()
+        self._kwargs_helper = ParamKwargsHelper(self.s3)
         S3FileSystem._singleton[0] = self
 
+    def _filter_kwargs(self, s3_method, kwargs):
+        return self._kwargs_helper.filter_dict(s3_method.__name__, kwargs)
+
+    def _call_s3(self, method, *akwarglist, **kwargs):
```
mariusvniekerk: So right now this is only used for the writing methods but will work for everything

mariusvniekerk:
Mostly just had the file have its own call_s3 to simplify the code. It also allows us to easily have various levels of override. (s3fs,  file, method)  and apply them in sequence

s3fs/core.py
```diff
@@ -71,6 +71,59 @@ def split_path(path):
...
+    Parameters
+    ----------
+    string : underscore seperated string
+    """
+    return ''.join([x.capitalize() for x in string.split('_')])
+
+
+class SSEParams(object):
+
+    def __init__(self, server_side_encryption=None, sse_cusomer_algorithm=None,
```
dan-coates: Typo in "customer" here.

mariusvniekerk:
Sadly there is no good way to test using moto.

s3fs/core.py
```diff
@@ -71,6 +71,59 @@ def split_path(path):
...
+    actually using
+
+    Parameters
+    ----------
+    s3 : S3FileSystem
+    """
+
+    def __init__(self, s3):
+        self.s3 = s3
+        self._kwarg_cache = {}
```
martindurant: Can different s3 instances have different values? I'm thinking not. Perhaps this should be a class-level dict, and the S3FileSystem instances won't need their own instances of ParamKwargsHelper, just call like a function.

s3fs/core.py
```diff
@@ -71,6 +71,59 @@ def split_path(path):
              'authenticated-read'}
 
 
+def title_case(string):
```
martindurant: It would be nice to put this and the helpers below into util instead, a little cleaner.

s3fs/core.py
```diff
@@ -71,6 +71,59 @@ def split_path(path):
...
+    TitleCases a given string.
+
+    Parameters
+    ----------
+    string : underscore seperated string
+    """
+    return ''.join([x.capitalize() for x in string.split('_')])
+
+
+class SSEParams(object):
```
martindurant: So this is not really used, since we can provide arbitrary writer_kwargs - do you think it's necessary now? The only thing it really does, is list the fields we expect to be required specifically for SSE, which to me seems more like a documentation task.

s3fs/core.py
```diff
@@ -71,6 +71,59 @@ def split_path(path):
...
+        if model_name not in self._kwarg_cache:
+            model = self.s3.meta.service_model.operation_model(model_name)
+            valid_keys = set(model.input_shape.members.keys())
+            self._kwarg_cache[model_name] = valid_keys
+        return self._kwarg_cache[model_name]
+
+    def filter_dict(self, method_name, d):
+        model_name = title_case(method_name)
+        valid_keys = self._get_valid_keys(model_name)
+        if isinstance(d, SSEParams):
```
martindurant: It basically is a dictionary already.

s3fs/tests/test_s3fs.py
```diff
@@ -519,13 +549,27 @@ def test_dynamic_add_rm(s3):
...
     assert s3.cat(test_bucket_name+'/test') == b'hello'
     s3.open(test_bucket_name+'/test', 'wb').close()
     assert s3.info(test_bucket_name+'/test')['Size'] == 0
 
+
+def test_write_small_secure(s3):
+    # Unfortunately moto does not yet support enforcing SSE policies.  It also
+    # does not return the correct objects that can be used to test the results
+    # effectively.
+    # This test is left as a placeholder in case moto eventually supports this.
```
martindurant: but this same code works on real s3, right?

s3fs/core.py
mariusvniekerk: I think it is still helpful to provide some settings that users can discover via tab completion.

s3fs/tests/test_s3fs.py
mariusvniekerk: Yes

s3fs/core.py
mariusvniekerk: yeah agree

s3fs/core.py
martindurant: How about, instead of a class, a simple function with keyword arguments like the __init__ here, that gives you back a dictionary to be used as kwargs?

martindurant:
@mariusvniekerk , let me know when you think this is ready for me to look through again.

mariusvniekerk:
Do we want to just add in kwargs support for everything?

martindurant:
I suppose we must

mariusvniekerk:
I think we basically need some documentation now for the pieces.  Not sure if we want to litter every method with `**kwargs : forwarded to underlying s3 method see [ref] for details`.

martindurant:
No, I don't think there's any reason to explicitly document kwargs like that, it should be in the class docstring and in documentation where SSE is mentioned, and perhaps only on the open method (because a file needs to keep hold of the values, and so acts a little differently).

s3fs/core.py
```diff
@@ -97,6 +98,8 @@ class S3FileSystem(object):
     use_ssl : bool (True)
         Whether to use SSL in connections to S3; may be faster without, but
         insecure
+    writer_kwargs : dict of parameters that are used when writing data to s3.
+        Typically used for things like "ServerSideEncryption".
```
martindurant: So this should be "file_kwargs" and I would put here that they can be over-ridden by kwargs to any of the methods which accept them.

s3fs/core.py
```diff
@@ -259,6 +277,9 @@ def open(self, path, mode='rb', block_size=None, acl='',
             out of a file, performance may be better if False.
         acl: str
             Canned ACL to set when writing
+        kwargs: dict-like
+            Additional parameters used for file writing.  Typically used for
```
martindurant: Not only writing, now

mariusvniekerk:
Probably.  Added mock to the conda env creation.

mariusvniekerk:
and we are ✅

martindurant:
Excellent, thank you very much!

PR #92, opened 2017-05-18
Use additional kwargs

scizo:
Recently there have been additional kwargs that are aggregated so
that can be passed through when calling boto methods. These aren't
currently being used. This uses them.


martindurant:
cc @mariusvniekerk

mariusvniekerk:
Whoops

martindurant:
Thanks for picking this up @scizo

PR #93, opened 2017-05-25
Simply ignore flushes if buffer smaller than block size

lsglick:
The current behavior of raising an assertion causes some small files to be unwritable.

For example, with [`io.TextIOWrapper`](https://docs.python.org/2/library/io.html#io.TextIOWrapper) used by `dask.bytes`, a `close()` call internally first calls a `flush()`. And if the buffer was less than the block size, the exception would be raised and the file couldn't be written.

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -1092,7 +1092,7 @@ class S3File(object):
         """
         Write data to buffer.
 
-        Buffer only sent to S3 on flush() or if buffer is greater than or equal to blocksize.
+        Buffer only sent to S3 on close() or if buffer is greater than or equal to blocksize.
 
         Parameters
         ----------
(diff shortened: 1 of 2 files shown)


martindurant:
Thank you for getting to this so quickly!
Looks good.
Would you please update the docstring for flush(), saying that calling it when the buffer is smaller than the blocksize is a no-op.

lsglick:
Updated!

martindurant:
"It is unsafe to call this function repeatedly." is still in there :|  Sorry.

martindurant:
I mean, could you please remove that phrase, since it is no longer true.

martindurant:
Excellent, thank you!

PR #94, opened 2017-05-26
Ensure current() uses subclass

lsglick:
If subclassing `S3FileSystem`, `current()` should give you an instance of the subclass.

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -154,7 +154,7 @@ class S3FileSystem(object):
         self.use_ssl = use_ssl
         self.s3 = self.connect()
         self._kwargs_helper = ParamKwargsHelper(self.s3)
-        self.__class__._singleton[0] = self
+        self._singleton[0] = self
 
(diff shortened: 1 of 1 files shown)


s3fs/core.py
```diff
@@ -154,7 +154,7 @@ def __init__(self, anon=False, key=None, secret=None, token=None,
         self.use_ssl = use_ssl
         self.s3 = self.connect()
         self._kwargs_helper = ParamKwargsHelper(self.s3)
-        S3FileSystem._singleton[0] = self
+        self.__class__._singleton[0] = self
```
martindurant: I think this could be `self._singleton[0]`, which is a bit cleaner.

lsglick:
Thanks! For my own needs I was going to use s3fs as a `FileSystem` client in Luigi rather than their S3Client. It seemed much simpler to subclass and override the methods necessary to satisfy the interface, rather than compose it since the methods were so close already.

martindurant:
OK, thank you!

PR #98, opened 2017-06-14
filter all kwargs in _call_s3 to prevent botocore ParamValidationError (issue #97)

### LINKED ISSUE #97: Write fails with large files and `s3_additional_kwargs`
Writing with server-side encryption with files large enough to trigger multipart upload fails:

```python
s3 = s3fs.S3FileSystem(s3_additional_kwargs = {'ServerSideEncryption': 'AES256'})
with s3.open('example-bucket/test.file', 'wb') as f:
        f.write(b'a' * 10**7) # force multipart upload
```

```python

  File "/Users/del82/Documents/git/s3fs/s3fs/core.py", line 1164, in flush
    Body=self.buffer.read(), Key=self.key)
  File "/Users/del82/Documents/git/s3fs/s3fs/core.py", line 941, in _call_s3
    return self.s3._call_s3(method, self.s3_additional_kwargs, *kwarglist, **kwargs)
  File "/Users/del82/Documents/git/s3fs/s3fs/core.py", line 175, in _call_s3
    return method(**additional_kwargs)
  File "/Users/del82/anaconda3/envs/dask/lib/python3.6/site-packages/botocore/client.py", line 253, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/Users/del82/anaconda3/envs/dask/lib/python3.6/site-packages/botocore/client.py", line 531, in _make_api_call
    api_params, operation_model, context=request_context)
  File "/Users/del82/anaconda3/envs/dask/lib/python3.6/site-packages/botocore/client.py", line 586, in _convert_to_request_dict
    api_params, operation_model)
  File "/Users/del82/anaconda3/envs/dask/lib/python3.6/site-packages/botocore/validate.py", line 291, in serialize_to_request
    raise ParamValidationError(report=report.generate_report())
botocore.exceptions.ParamValidationError: Parameter validation failed:
Unknown parameter in input: "ServerSideEncryption", must be one of: Body, Bucket, ContentLength, ContentMD5, Key, PartNumber, UploadId, SSECustomerAlgorithm, SSECustomerKey, SSECustomerKeyMD5, RequestPayer
```

(I recognize that this may not be [the preferred way to do this](https://github.com/dask/s3fs/blob/master/s3fs/tests/test_s3fs.py#L566-L568), but it mirrors the way that dask does it, so failing to handle this case means not being able to write large objects to S3 using SSE from dask.)

Multipart uploads use not only `botocore.S3.client.put_object` but also `client.create_multipart_upload`, `client.upload_part`, etc. which don't accept `ServerSideEncryption` as a kwarg, but it's passed to them and results in the above exception.

del82:
This fixes #97; I'd be grateful for pointers on how to write tests exposing the problem.

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -162,9 +162,11 @@ class S3FileSystem(object):
     def _call_s3(self, method, *akwarglist, **kwargs):
         additional_kwargs = self.s3_additional_kwargs.copy()
         for akwargs in akwarglist:
-            additional_kwargs.update(self._filter_kwargs(method, akwargs))
+            additional_kwargs.update(akwargs)
         # Add the normal kwargs in
         additional_kwargs.update(kwargs)
+        # filter all kwargs
+        additional_kwargs = self._filter_kwargs(method, additional_kwargs)
         return method(**additional_kwargs)
 
     @classmethod
@@ -445,7 +447,7 @@ class S3FileSystem(object):
 
         Parameters
         ---------
-        kw_args : key-value pairs like field="value", where the values must be strings. Does not alter existing fields, 
+        kw_args : key-value pairs like field="value", where the values must be strings. Does not alter existing fields,
             unless the field appears here - if the value is None, delete the field.
         copy_args : dict, optional
             dictionary of additional params to use for the underlying s3.copy_object.
@@ -455,7 +457,7 @@ class S3FileSystem(object):
         >>> mys3file.setxattr(attribute_1='value1', attribute_2='value2')  # doctest: +SKIP
         # Example for use with copy_args
         >>> mys3file.setxattr(copy_kwargs={'ContentType': 'application/pdf'}, attribute_1='value1')  # doctest: +SKIP
-        
+
 
         .. Metadata Reference:
         http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html#object-metadata
--- a/s3fs/tests/test_s3fs.py
+++ b/s3fs/tests/test_s3fs.py
@@ -569,6 +569,19 @@ def test_write_small_secure(s3):
     assert s3.cat(secure_bucket_name+'/test') == b'hello'
     head = s3.s3.head_object(Bucket=secure_bucket_name, Key='test')
 
+def test_write_large_secure(s3):
+    mock = moto.mock_s3()
+    mock.start()
+
+    # build our own s3fs with the relevant additional kwarg
+    s3 = S3FileSystem(s3_additional_kwargs = {'ServerSideEncryption': 'AES256'})
+    s3.mkdir('mybucket')
+
+    with s3.open('mybucket/myfile', 'wb') as f:
+        f.write(b'hello hello' * 10**6)
+
+    assert s3.cat('mybucket/myfile') == b'hello hello' * 10**6
+
 
 def test_write_fails(s3):
     with pytest.raises(NotImplementedError):


martindurant:
Unfortunately, moto (the s3 mocking package) does not implement SSE, so I don't think we can test this in CI. You could write a test just like the code in the issue, which uses the default S3FileSystem (with default system credentials), and mark it to skip when those credentials are not defined. This is problematic, though, because the real AWS S3 bucket names are global, we cannot depend on the availability of some test bucket.

del82:
The test I just added seems to work;  I created a new mocked s3fs for the purpose of the test with the relevant kwarg.  The fact that moto doesn't support SSE doesn't end up mattering, because the issue is that we're passing unrecognized kwargs to botocore; it's just coincidence that the kwarg that caused the problem was SSE related.  And moto seems to check for unrecognized kwargs, or at least without the above change in `s3fs/core.py` the test fails with a `ParamValidationError` and with the change it passes.

martindurant:
Well OK, then - thank you!

kinzleb:
I'm running into this ParamValidationError when trying to use mv with the s3_additional_kwargs set.  The mv method just wraps the copy and rm method, and the copy method runs fine but it appears the rm doesn't work with the s3_additional_kwargs values set even though it looks like the intention was to not pass kwargs when rm is called within mv.

```python
# setup file system to use encryption key
fs = s3fs.S3FileSystem(
    use_ssl = True,
    config_kwargs={
        'signature_version':'s3v4'
    },
    s3_additional_kwargs={
        'ServerSideEncryption':'aws:kms',
        'SSEKMSKeyId':kms_id
    }
)

# extend the timeout since these are large files
fs.read_timeout = 3600
fs.connect_timeout = 3600

pprint(vars(fs))
```

    {'_kwargs_helper': <s3fs.utils.ParamKwargsHelper object at 0x1075c4278>,
     'anon': False,
     'client_kwargs': {},
     'config_kwargs': {'signature_version': 's3v4'},
     'connect_timeout': 3600,
     'default_fill_cache': True,
     'dirs': {},
     'key': None,
     'kwargs': {},
     'read_timeout': 3600,
     'req_kw': {},
     's3': <botocore.client.S3 object at 0x10a393f98>,
     's3_additional_kwargs': {'SSEKMSKeyId': 'XXXXXX-XXXXX-XXXX-XXXX-XXXXXXXXXX',
                              'ServerSideEncryption': 'aws:kms'},
     'secret': None,
     'session': Session(region_name='us-east-1'),
     'token': None,
     'use_ssl': True}



```python
# overwrite the original file with an encrypted copy
fs.mv('bucket/nukeme.txt', 'bucket/nukeme.txt')
```


    ---------------------------------------------------------------------------

    ParamValidationError                      Traceback (most recent call last)

    <ipython-input-3-190634ab4682> in <module>()
    ----> 1 fs.mv('bucket/nukeme.txt', 'bucket/nukeme.txt')
    

    ~/notebook-env/lib/python3.6/site-packages/s3fs/core.py in mv(self, path1, path2, **kwargs)
        645         """ Move file between locations on S3 """
        646         self.copy(path1, path2, **kwargs)
    --> 647         self.rm(path1)
        648 
        649     def merge(self, path, filelist, **kwargs):


    ~/notebook-env/lib/python3.6/site-packages/s3fs/core.py in rm(self, path, recursive, **kwargs)
        746             try:
        747                 self._call_s3(
    --> 748                     self.s3.delete_object, kwargs, Bucket=bucket, Key=key)
        749             except ClientError:
        750                 raise IOError('Delete key failed', (bucket, key))


    ~/notebook-env/lib/python3.6/site-packages/s3fs/core.py in _call_s3(self, method, *akwarglist, **kwargs)
        166         # Add the normal kwargs in
        167         additional_kwargs.update(kwargs)
    --> 168         return method(**additional_kwargs)
        169 
        170     @classmethod


    ~/notebook-env/lib/python3.6/site-packages/botocore/client.py in _api_call(self, *args, **kwargs)
        308                     "%s() only accepts keyword arguments." % py_operation_name)
        309             # The "self" in this scope is referring to the BaseClient.
    --> 310             return self._make_api_call(operation_name, kwargs)
        311 
        312         _api_call.__name__ = str(py_operation_name)


    ~/notebook-env/lib/python3.6/site-packages/botocore/client.py in _make_api_call(self, operation_name, api_params)
        571         }
        572         request_dict = self._convert_to_request_dict(
    --> 573             api_params, operation_model, context=request_context)
        574 
        575         handler, event_response = self.meta.events.emit_until_response(


    ~/notebook-env/lib/python3.6/site-packages/botocore/client.py in _convert_to_request_dict(self, api_params, operation_model, context)
        626 
        627         request_dict = self._serializer.serialize_to_request(
    --> 628             api_params, operation_model)
        629         prepare_request_dict(request_dict, endpoint_url=self._endpoint.host,
        630                              user_agent=self._client_config.user_agent,


    ~/notebook-env/lib/python3.6/site-packages/botocore/validate.py in serialize_to_request(self, parameters, operation_model)
        289                                                     operation_model.input_shape)
        290             if report.has_errors():
    --> 291                 raise ParamValidationError(report=report.generate_report())
        292         return self._serializer.serialize_to_request(parameters,
        293                                                      operation_model)


    ParamValidationError: Parameter validation failed:
    Unknown parameter in input: "ServerSideEncryption", must be one of: Bucket, Key, MFA, VersionId, RequestPayer
    Unknown parameter in input: "SSEKMSKeyId", must be one of: Bucket, Key, MFA, VersionId, RequestPayer



Somehow the s3_additional_kwargs are still getting passed when this is called:
        747                 self._call_s3(
    --> 748                     self.s3.delete_object, kwargs, Bucket=bucket, Key=key)

del82:
Can you verify you're running the tip of the master branch?  This patch hasn't made it into a release yet, and it looks from the trace you posted from `s3fs/core.py` it looks like you've got the release version rather than master.

kinzleb:
Oh yes @del82 , I see now that was the first commit that didn't make it into the latest release.  I did verify that this change fixed my problem when I switched to using the tip of master, I no longer get the ParamValidationError errors; so thanks for the fix and for taking the time to look at this!

PR #108, opened 2017-12-14
Change underlying implementation of info

tomlaube:
In the current implementation in order to open a key for read, info method is called, which subsequently performs _ls on file's "directory". This feature makes it impossible to use s3fs when there is a lot of keys under a prefix.

Since one of the most common ways of using s3 is like a key/val store, where metadata about keys are stored in rds and s3 is used as a cheap storage, it would to support this usa case.

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -646,8 +646,11 @@ class S3FileSystem(object):
 
     def rmdir(self, path, **kwargs):
         """ Remove empty key or bucket """
+        if path.startswith('s3://'):
+            path = path[len('s3://'):]
+        path = path.rstrip('/')
         bucket, key = split_path(path)
-        if (key and self.du(path, total=True) == 0) or not key:
+        if ((key and self.du(path, total=True) == 0) or not key) and not self._ls(path):
             self.rm(path, **kwargs)
         else:
             raise IOError('Path is not directory-like', path)
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -405,7 +405,8 @@ class S3FileSystem(object):
                     'LastModified': out['LastModified'],
                     'Size': out['ContentLength'], 'StorageClass': "STANDARD"}
             return out
-        except (ClientError, ParamValidationError):
+        except (ClientError, ParamValidationError) as e:
+            logger.debug("Failed to head path %s", path, exc_info=True)
             raise FileNotFoundError(path)
 
     _metadata_cache = {}
@@ -646,7 +647,7 @@ class S3FileSystem(object):
     def rmdir(self, path, **kwargs):
         """ Remove empty key or bucket """
         bucket, key = split_path(path)
-        if (key and self.info(path)['Size'] == 0) or not key:
+        if (key and self.du(path, total=True) == 0) or not key:
             self.rm(path, **kwargs)
         else:
             raise IOError('Path is not directory-like', path)
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -380,29 +380,33 @@ class S3FileSystem(object):
         else:
             return [f['Key'] for f in files]
 
-    def info(self, path, refresh=False, **kwargs):
+    def info(self, path, **kwargs):
         """ Detail on the specific file pointed to by path.
 
         Gets details only for a specific key, directories/buckets cannot be
         used with info.
         """
         parent = path.rsplit('/', 1)[0]
-        files = self._lsdir(parent, refresh=refresh)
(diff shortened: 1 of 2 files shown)


martindurant:
I generally agree with the idea.

You seem to be working from an out-of-dat branch.

The same idea in GCSFS: xref https://github.com/dask/gcsfs/pull/22

martindurant:
I think I would make the logic:
- if info() is called, whether directly or as part or an open(mode='r'), I would check if the key is already in the dir cache, and return it, else do the HEAD call
- all other uses (ls, du, ...) list the directory as before.

That way, we don't need an explicit argument from the user on whether to call HEAD or not.

tomlaube:
Changed as suggested. Together with that i removed part of test_info, since it was assuming that info is filling the dir cache by calling _lsdir

s3fs/core.py
```diff
@@ -386,23 +386,21 @@ def info(self, path, refresh=False, **kwargs):
...
-                       'LastModified': out['LastModified'],
-                       'Size': out['ContentLength'], 'StorageClass': "STANDARD"}
-                return out
-            except (ClientError, ParamValidationError):
-                raise FileNotFoundError(path)
+        if path in self.dirs:
+            files = self.dirs[path]
+            if len(files) == 1:
+                return files[0]
+
```
martindurant: Also could be in `self.dirs[path.rsplit('/', 1)[0]]` (i.e., 'parent' from above).

setup.py
```diff
@@ -4,7 +4,7 @@
 from setuptools import setup
 
 setup(name='s3fs',
-      version='0.1.2',
+      version='0.1.3',
```
martindurant: I'll do this at release time.

s3fs/tests/test_s3fs.py
```diff
@@ -152,10 +152,6 @@ def test_info(s3):
     s3.touch(a)
     s3.touch(b)
     assert s3.info(a) == s3.ls(a, detail=True)[0]
-    parent = a.rsplit('/', 1)[0]
-    s3.dirs[parent].pop(0)  # disappear our file!
```
martindurant: Should still be there, because of the `ls`, above - no?

tomlaube: What the ls adds to cache is the full path of a, it's actually the old implementation of info that adds it's parent to the cache, so no

martindurant:
Please can you see whether #109  can be fixed here?

tomlaube:
Well the issue isn't the implementation of info, but it's usage of it in the context of rmdir. Since rmdir is suppose work only on dirs/bucket you can simply replace the info with du as:
```
du(path, Total=True)
```

Adding the to PR..

s3fs/core.py
```diff
@@ -642,7 +647,7 @@ def mkdir(self, path, acl="", **kwargs):
     def rmdir(self, path, **kwargs):
         """ Remove empty key or bucket """
         bucket, key = split_path(path)
-        if (key and self.info(path)['Size'] == 0) or not key:
+        if (key and self.du(path, total=True) == 0) or not key:
```
martindurant: I guess this isn't quite right: if the path happens to contain keys of zero size, we'll delete anyway. Conversely, I don't think this can delete a bucket. I am tempted to accept the PR anyway as is, but if you can think of a quick fix and test, that would be welcome.

tomlaube:
Actually it will delete bucket since it's really considered as another directory, I'll add a test for that. I'll modify the behavior so that it behaves more as unix rmdir...

s3fs/core.py
tomlaube: Added here checking if there are any sub directories. Since the du underneath calls ls everything should come from dirs cache

martindurant: I understand, but there could be legitimate keys under this directory that happen to have zero size, no?

tomlaube: yes i agree please see the follow up commit

s3fs/core.py
```diff
@@ -646,8 +646,11 @@ def mkdir(self, path, acl="", **kwargs):
 
     def rmdir(self, path, **kwargs):
         """ Remove empty key or bucket """
+        if path.startswith('s3://'):
+            path = path[len('s3://'):]
+        path = path.rstrip('/')
         bucket, key = split_path(path)
-        if (key and self.du(path, total=True) == 0) or not key:
+        if ((key and self.du(path, total=True) == 0) or not key) and not self._ls(path):
```
martindurant: Got you :)
In this case, you might not need the `du` at all.

PR #115, opened 2018-01-29
add mode paramter to put_tags to allow for merging tags instead of ov…

kinzleb:
@martindurant Just a thought, but it may be convenient to add a parameter to the put_tags method which would allow merging in the new tag values with existing tags instead of overwriting.

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -447,7 +447,7 @@ class S3FileSystem(object):
                                  Bucket=bucket, Key=key)
         return {v['Key']: v['Value'] for v in response['TagSet']}
 
-    def put_tags(self, path, tags):
+    def put_tags(self, path, tags, mode='o'):
         """Set tags for given existing key
 
         Tags are a str:str mapping that can be attached to any key, see
@@ -461,12 +461,25 @@ class S3FileSystem(object):
         path: str
             Existing key to attach tags to
         tags: dict str, str
-            Tags to apply. Will over-write any existing tags.
+            Tags to apply.
+        mode:
+            One of 'o' or 'm'
+            'o': Will over-write any existing tags.
+            'm': Will merge in new tags with existing tags.  Incurs two remote calls.
         """
         bucket, key = split_path(path)
-        tag = {'TagSet': [
-                   {'Key': k, 'Value': v} for k, v in tags.items()]
-               }
+
+        if mode == 'm':
+            existing_tags = self.get_tags(path=path)
+            merged_tags = existing_tags.copy()
+            merged_tags.update(tags)
+            new_tags = [{'Key': k, 'Value': v} for k, v in merged_tags.items()]
+        elif mode == 'o':
+            new_tags = [{'Key': k, 'Value': v} for k, v in tags.items()]
+        else:
+            raise NotImplementedError("Mode must be {'o', 'm'}, not %s" % mode)
+
+        tag = {'TagSet': new_tags}
         self._call_s3(self.s3.put_object_tagging,
                       Bucket=bucket, Key=key, Tagging=tag)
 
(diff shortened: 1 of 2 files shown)


s3fs/core.py
```diff
@@ -461,12 +461,25 @@ def put_tags(self, path, tags):
...
+
+        if mode == 'm':
+            existing_tags = self.get_tags(path=path)
+            merged_tags = existing_tags.copy()
+            merged_tags.update(tags)
+            new_tags = [{'Key': k, 'Value': v} for k, v in merged_tags.items()]
+        elif mode == 'o':
+            new_tags = [{'Key': k, 'Value': v} for k, v in tags.items()]
+        else:
+            raise NotImplementedError("Mode must be {'o', 'm'}, not %s" % mode)
```
martindurant: Probably a ValueError - there are no other values of mode that we intend to implement.

s3fs/core.py
```diff
@@ -461,12 +461,25 @@ def put_tags(self, path, tags):
...
         bucket, key = split_path(path)
-        tag = {'TagSet': [
-                   {'Key': k, 'Value': v} for k, v in tags.items()]
-               }
+
+        if mode == 'm':
+            existing_tags = self.get_tags(path=path)
+            merged_tags = existing_tags.copy()
+            merged_tags.update(tags)
+            new_tags = [{'Key': k, 'Value': v} for k, v in merged_tags.items()]
```
martindurant: could be a little simpler? 
```
if mode == 'm':
    new_tags = [{'Key': k, 'Value': v} for k, v in list(tags.items()) + list(self.get_tags(path=path).items())]
```
Maybe you don't think that's simpler... in any case, I don't think the `copy()` line is necessary.

kinzleb: I think you're right, we could get rid of copy.  I wasn't sure how something like you suggest would handle updating the values of existing keys.  I used "update" since it is compatible with python 2.7 -- as of python 3.5 it looks like there is simpler way to merge two dicts e.g. {**x, **y}.   I will add a test and change the raise to ValueError.

s3fs/tests/test_s3fs.py
```diff
@@ -939,3 +939,9 @@ def test_tags(s3):
     s3.touch(fname)
     s3.put_tags(fname, tagset)
     assert s3.get_tags(fname) == tagset
+
+    # Ensure merge mode updates value of existing key and adds new one
+    new_tagset = {'tag2': 'updatedvalue2', 'tag3': 'value3'}
+    tagset.update(new_tagset)
+    s3.put_tags(fname, tagset, mode='m')
```
martindurant: Should be `s3.put_tags(fname, new_tagset, mode='m')`, else you are not really testing merge :)

kinzleb: Sorry my code didn't make it very clear what I was doing there -- I'm updating tagset with the values in new_tagset and then just reusing tagset:
```
>>> tagset = {'tag1': 'value1', 'tag2': 'value2'}
>>> new_tagset = {'tag2': 'updatedvalue2', 'tag3': 'value3'}
>>> tagset.update(new_tagset)
>>> tagset
{'tag1': 'value1', 'tag2': 'updatedvalue2', 'tag3': 'value3'}
```

martindurant: Yes, I see that the values are correctly updated, but `s3.put_tags(fname, tagset, mode='m')` uses the complete merged set of tags, not only the modifications - so it's not relying on grabbing the existing tags for the object first, it would work with mode 'o' too.

kinzleb: Ah, I see what you mean.

kinzleb:
Alright, I think that should make the test a little more useful now -- thanks for catching that :)

martindurant:
Thank you!

PR #117, opened 2018-03-05
Add copy_managed to use the managed copy functions in boto3

mariusvniekerk:
These are required if an object exceeds 5GB in size.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,8 @@ language: python
 matrix:
   include:
     - python: 2.7
-    - python: 3.4
     - python: 3.5
+    - python: 3.6
 
 install:
   # Install conda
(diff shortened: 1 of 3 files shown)


martindurant:
What's the real difference here? It looks like `CopySource='/'.join([buc1, key1])` versus `CopySource={'Bucket': buc1,'Key': key1}`, and calling `s3.copy` instead of `s3.copy_object`.

Why the move for the boto3 install? Suggests the library won't work with the conda boto3?

mariusvniekerk:
Mostly just fiddling with getting CI passing.  The conda boto3 is pretty heavily out of date.

martindurant:
Perhaps specify conda-forge? Looks pretty up to date on there https://anaconda.org/conda-forge/boto3/files

mariusvniekerk:
The key different is copy vs copy_object.   Copy is a higher order boto3 method that uses s3transfer under the hood allowing things like callbacks etc.  Ultimately it uses uploadPartCopy instead of copyObject.

mariusvniekerk:
I'm not sure if those are supported by other s3 compatible engines though.

martindurant:
@mariusvniekerk , if this ready for merger?

mariusvniekerk:
Yep. The rest of the apis we use in s3fs are safe with large files

kinzleb:
This is working great for me, really improved copy times for large files -- thanks!

martindurant:
Glad it helped. This was released today in v0.1.4 on conda-forge and pypi.

PR #123, opened 2018-03-29
Change glob()'s treatment of dot and slash

### LINKED ISSUE #122: Dots (.) in glob
`glob` doesn't seem to be dealing with dots (`.`) correctly:
```python3
>>> from s3fs import S3FileSystem
>>> s3 = S3FileSystem()
>>> s3.ls("s3fs-glob-test")
['s3fs-glob-test/ab.txt']
>>> s3.glob("s3fs-glob-test/a.*.txt")
['s3fs-glob-test/ab.txt']
```
The path is returned even though `ab.txt` shouldn't match the pattern `a.*.txt`.

zhangwen0411:
Fixes #122.

I also slightly changed how slashes are treated in some places, squashing multiple consecutive slashes into one. Not sure if that's the desired behavior, though.

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -584,6 +584,33 @@ class S3FileSystem(object):
             self._call_s3(self.s3.put_bucket_acl,
                           kwargs, Bucket=bucket, ACL=acl)
 
+    @staticmethod
+    def _normalize_slashes(path):
+        """Normalizes slashes (/) in path.
+
+        Specifically, replaces multiple consecutive slashes with a single slash
+        and removes any trailing slashes.
+        """
+        return re.sub(r"/+", "/", path).rstrip("/")
+
+    @staticmethod
+    def _compile_glob_to_re(glob_pattern):
+        """Translates a glob pattern into a compiled regex pattern.
+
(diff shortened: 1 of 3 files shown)


martindurant:
Thank you for this, what you've done looks very reasonable - especially as it reuses `fnmatch`'s logic. I would request that, rather than adding more methods onto the S3FileSystem class, the code is either inlined within `glob` or moved as functions within the utils module.
If you are interested, exactly the same code could be ported to gcsfs.

zhangwen0411:
Ah I see! So what's the purpose of replacing `//` with `/` in the original implementation (e.g., https://github.com/dask/s3fs/blob/master/s3fs/core.py#L613)? Should I just leave the slashes alone?

Thanks!

martindurant:
I see what you mean, and honestly I don't know. This may have been copied from hdfs3, which requires posix-compliant paths where such separator mergers happen. Keys can also end with '/' legally, without being directories - so probably best to try dropping the slash handling and see if it breaks any assumptions in the tests.

martindurant:
Thank you!

PR #134, opened 2018-05-08
Minor code cleanup and unit test fixes

### LINKED ISSUE #133: LICENSE.txt is not deployed on pypi
As per summary the License fileis not shipped with the archive uploaded to pypi.

dargueta:
* Fix a couple asserts in tests that were either missing or misplaced. For example, one assert was outside the loop it was supposed to be in.
* Remove custom `tmpfile` context manager, since `pytest` has a built-in one. I pegged `pytest` to ≥ 2.7.3 because it has a bugfix that's nice to have for temporary files.
* Use set literal notation in a few places for cleaner code. Also avoids global lookups.
* Add `futures` as a test requirement for Python 2. It was working by chance since it's an indirect dependency of another package.
* Move test requirements to their own file so developers can easily install them locally with pip
* Add a few things to `.gitignore` -  `pytest`'s new cache directory and the `virtualenv` environment file for those of us who don't use conda. 🙂 
* Removed an extraneous `read()` in `read_block()`.
* Fix #133

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,8 +17,9 @@ install:
   - conda update conda
 
   # Install dependencies
-  - conda create -n test -c conda-forge python=$TRAVIS_PYTHON_VERSION --file requirements.txt --file test_requirements.txt
+  - conda create -n test -c conda-forge python=$TRAVIS_PYTHON_VERSION pip
   - source activate test
+  - pip install -r requirements.txt -r test_requirements.txt
 
 script:
   - py.test -x --doctest-modules --pyargs s3fs
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ install:
   - conda update conda
 
   # Install dependencies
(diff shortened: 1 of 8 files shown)


.travis.yml
```diff
@@ -17,8 +17,9 @@ install:
   - conda update conda
 
   # Install dependencies
-  - conda create -n test -c conda-forge python=$TRAVIS_PYTHON_VERSION pytest mock moto boto3
+  - conda create -n test -c conda-forge python=$TRAVIS_PYTHON_VERSION pip
   - source activate test
+  - pip install -r requirements.txt -r test_requirements.txt
```
martindurant: What was not being installed previously?

dargueta: Nothing, I just moved the installation from `conda` *outside* the environment to `pip` *inside* the environment. The reason for this is that we now can change the requirements in only one place and won't have to worry about updating the Travis file as well.

dargueta:
Just fixed #133. Build seems to be stalled on Travis for some reason.

martindurant:
Thank you @dargueta

Issue #135, opened 2018-05-08
Changing default block size also changes it for S3FileSystems already created


dargueta:
You can accidentally change the default block size for `S3FileSystem` objects already created by changing `s3fs.S3FileSystem.default_block_size`, but only if they didn't override the default block size when they were created.

To reproduce:

```py
import s3fs

fs_default = s3fs.S3FileSystem()
assert fs_default.default_block_size == 5 * (1024 ** 2)        # OK

fs_overridden = s3fs.S3FileSystem(default_block_size=64 * (1024 ** 2))
assert fs_overridden.default_block_size == 64 * (1024 ** 2)     # OK

# Suppose I want all subsequent file systems to have a block size of 1 GiB
# instead of 5 MiB:

s3fs.S3FileSystem.default_block_size = 1024 ** 3

fs_big = s3fs.S3FileSystem()
assert fs_big.default_block_size == 1024 ** 3             # OK

# Test the other file systems created to see if their cache sizes have changed
assert fs_overridden.default_block_size == 64 * (1024 ** 2)     # OK
assert fs_default.default_block_size == 5 * (1024 ** 2)        # Fails, block size changed
```

This can result in different files created by `open` in the same `S3FileSystem` to have different block sizes and therefore different flush behaviors. Personally I can't readily think of a situation where this could have an adverse effect on performance or crash user code, but in any case it *is* an unexpected behavior. I guess if someone subclasses `S3FileSystem` that could cause problems?

A quick solution would be to remove the class-level attribute `default_block_size` and set the default value in the constructor, like so:

```py
class S3FileSystem(object):
    # as before...
    # Remove this line
    # default_block_size = 5 * 2**20

    def __init__(self, ..., default_block_size=5 * 2**20, ...):
        # as before...
        # Unconditionally set the block size here
        self.default_block_size = default_block_size
```

Unfortunately this is a breaking change, since we're removing the class attribute. Another potential fix is to leave it as is and fall back to the class-level one:

```py
class S3FileSystem(object):
    # ...
    def __init__(...):
        self.default_block_size = default_block_size or type(self).default_block_size
```

We'd use `type(self)` here so the code works as expected in subclasses.

martindurant:
Interesting discovery!
I think that the following is a reasonable solution
```
self.default_block_size = default_block_size or self.default_block_size
```
since this would pick up the attribute of the parent class to when subclassing, and now it always makes an instance attribute.

[...]

PR #136, opened 2018-05-12
Fix issue #135

dargueta:
See #135 for full details of bug and solution.


martindurant:
Looks great, thank you.

Issue #139, opened 2018-07-25
mkdir (create_bucket) throws IllegalLocationConstraintException


delijati:
Error:
```
fs = s3fs.S3FileSystem(client_kwargs={"region_name": region_name})
# tried also config_kwargs
...

botocore.exceptions.ClientError: An error occurred (IllegalLocationConstraintException) when calling the CreateBucket operation: The unspecified location constraint is incompatible for the region specific endpoint this request was sent to.
```

Solution:
```
s3.create_bucket(                                                                                                                                                                                
    Bucket=target_uri.bucket,
    CreateBucketConfiguration={'LocationConstraint': REGION})
```

More info [1]

[1] https://github.com/boto/boto3/issues/781

martindurant:
The error seems to have a very clear and obvious solution. Would you care to provide it?

martindurant:
Are you interested in making the appropriate PR, @delijati ? The boto authors worried that it isn't reasonable to guess the default region, so it would be reasonable to include the region to use as a keyword in the `mkdir` call.

delijati:
Done #140

PR #140, opened 2018-07-27
#139 add region_name to create_bucket

delijati:
#139 add region_name to create_bucket

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -961,7 +961,12 @@ class S3FileSystem(object):
             if acl and acl not in buck_acls:
                 raise ValueError('ACL not in %s', buck_acls)
             try:
-                self.s3.create_bucket(Bucket=bucket, ACL=acl)
+                params = {"Bucket": bucket, 'ACL': acl}
+                if "region_name" in self.client_kwargs:
+                    params['CreateBucketConfiguration'] = {
+                        'LocationConstraint': self.client_kwargs["region_name"]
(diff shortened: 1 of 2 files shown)


martindurant:
I think `region_name=None` should be an optional argument to `mkdir`, and that the fallback default in this code should only be used as the fallback for the `None`. What do you think?

delijati:
Changed it to this comment https://github.com/boto/boto3/issues/781#issuecomment-241916064

delijati:
btw. is it also possible to copy from user a bucket to a user b bucket, with different permissions?

martindurant:
Yep, I like that.
Please also add the parameter to the docstring in `mkdir`. I have no idea why py35, in particular, is failing, but it looks unrelated, perhaps a different version of the boto/moto stack. It would probably be reasonable to not test on py35 any more.

martindurant:
> is it also possible to copy from user a bucket to a user b bucket, with different permissions?

I don't know - I don't see why not.

delijati:
> is it also possible to copy from user a bucket to a user b bucket, with different permissions?
> I don't know - I don't see why not.

it seams only over first copying to local disk :/

https://stackoverflow.com/questions/41939170/copy-ami-from-one-aws-account-to-another-aws-account-using-boto

martindurant:
OK, going to merge this, and should open an issue to look into the py35 failure.

Issue #142, opened 2018-07-29
Allow users to provide a botocore session on creation of S3FileSystem


mfulleratlassian:
Currently the code creates botocore.session.Session objects during execution. We have reasons that we would like to provide S3FileSystem the session that is to be used (see: https://pypi.org/project/boto3_extensions/  RefreshableAssumeRoleProvider) . I will raise a PR that allows you to pass in a session object and S3FileSystem will use this session for the life of its execution instead of making new sessions.

jrbourbeau:
Closing as passing a session to `S3FileSystem` was added in #143. Thanks @mfulleratlassian!

PR #143, opened 2018-07-29
Issue 142 - Allow users to provide a botocore session on creation of S3FileSystem

mfulleratlassian:
See: https://github.com/dask/s3fs/issues/142

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -119,6 +119,9 @@ class S3FileSystem(object):
         objects.
     config_kwargs : dict of parameters passed to ``botocore.client.Config``
     kwargs : other parameters for boto3 session
+    session : botocore Session object to be used for all connections.
+         This session will be used inplace of creating a new session inside S3FileSystem.
+
 
     Examples
     --------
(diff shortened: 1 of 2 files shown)


martindurant:
A very reasonable request. Perhaps I would remove the code duplication by having 
```
if self.passed_in_session:
    ..
elif self.anon:
```

Can this be tested? The simplest would be to create a new instance, passing a session, and ensure that the `session` attribute is indeed the session we passed.

Have you thoughts on what should happen in the case of pickling?

mfulleratlassian:
Yeah I was just looking at the pickling code now. I will simplify the duplication and see what can be done with pickling. At the least I can make it raise a NotImplemented if you pass in a session.

martindurant:
> I can make it raise a NotImplemented if you pass in a session

that is probably suitable. Wouldn't mind  a test for that too :)

mfulleratlassian:
OK I have added some tests, prevented Pickling with a passed in session and tried to clean up the code. Let me know what else you would like updated. Thanks

mfulleratlassian:
Sweet, any ideas when you are planning the next release to PyPi? Once this goes out we can remove the custom version from our repos. 😃

martindurant:
There was no release since the end of April, so I suppose it could - I'll try to fit it in in the near future.

mfulleratlassian:
Thanks

PR #151, opened 2018-09-19
Set version_id on a file after writiting

mariusvniekerk:
Ensure that we set the version_id after closing a file.


martindurant:
You actually make use of this file attribute?
Is there any way to test this, or is it outside of moto's capabilities?

mariusvniekerk:
So i have an external to s3 metadata store that having this attribute exist would make updating a lot simpler.
We can probably test it with moto.

PR #153, opened 2018-09-27
Use raise_from from six in order to provide more context for failures.

### LINKED ISSUE #118: ClientErrors should bubble up to callers
Hi dask friends!

I spent a frustrating bit of time trying to figure out why s3fs was not able to find files in a bucket where I was sure they actually lived. After some digging I realized that when listing the contents of a bucket, `ClientError`s are swallowed and the contents of the bucket returned as empty: https://github.com/dask/s3fs/blob/master/s3fs/core.py#L320

What do you all think about letting that error bubble up to callers? I think this would be a friendlier user experience for people trying to sort out bucket permissions, etc.

mariusvniekerk:
This is an attempt to get some better exception handling for s3fs.
closes #118
The one key change here is that if a user lacks permissions to a directory `ls` will raise instead of silently returning nothing.  
Alternatively we could throw a warning here.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,3 @@
 botocore
 boto3
+six
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -34,6 +34,9 @@ except NameError:
         pass
 
 
+from six import raise_from
+
+
 # py2 has no ConnectionError only OSError with different error number for each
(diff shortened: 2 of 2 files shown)


martindurant:
Sounds very reasonable to me.

Issue #156, opened 2018-11-30
In consistent behavior of existence check for empty folders


drorata:
Consider the following minimal example:

```python
import moto
import boto3
import s3fs

with moto.mock_s3():
    print("Case 1")
    conn = boto3.client('s3', 'us-west-2')
    conn.create_bucket(Bucket='test-bucket')
    
    s3 = s3fs.S3FileSystem()
    
    s3.mkdir('test-bucket/foo/')
    print(s3.ls('test-bucket'))
    print(s3.exists('test-bucket/foo/'))
    s3.touch('test-bucket/foo/bar')
    print(s3.exists('test-bucket/foo/'))
    
with moto.mock_s3():
    print("Case 2")
    conn = boto3.client('s3', 'us-west-2')
    conn.create_bucket(Bucket='test-bucket')
    
    s3 = s3fs.S3FileSystem()
    
    s3.mkdir('test-bucket/foo')  # A missing trailing / is the only difference
    print(s3.ls('test-bucket'))
    print(s3.exists('test-bucket/foo/'))
    s3.touch('test-bucket/foo/bar')
    print(s3.exists('test-bucket/foo/'))
```

For which the output is:

```
Case 1
['test-bucket/foo']
False
True
Case 2
['test-bucket/foo']
True
True
```

Note that in the first case, as long as the folder `foo` is empty it can be listed but not detected by `s3.exists`. In contrast, in the second case, when the trailing `/` is removed when creating the directory, `.exists` detects the folder even when it is empty.

I also created empty folders directly from the S3 console, and still `.exists` couldn't find them.

If find it an un-expected behavior. Is it a bug or a feature?

drorata:
Related, and somehow confusing as well, compare `s3.mkdir('my-bucket/foo')` and `s3.mkdir('my-bucket/bar/'). The former creates an empty file and the latter a directory.

martindurant:
s3 doesn't actually have directories, only files. mkdir probably shouldn't do anything, ever. I'm not sure what making a directory in the console really does.

drorata:
It is not 100% correct. 

Consider:

```python
s3.mkdir('my-bucket/testing/with_trailing_slash/')
s3.mkdir('my-bucket/testing/without_trailing_slash')
```

Yields the following:

[...]

PR #158, opened 2018-12-04
Initial fix

drorata:
As discussed on #156, this natural fix breaks the test `test_s3fs.py::test_dynamic_add_rm`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,9 +6,6 @@ matrix:
   include:
     - python: 2.7
     - python: 3.6
-    - python: 3.7
-      dist: xenial
-      sudo: true
 
 install:
   # Install conda
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,8 +19,7 @@ install:
   # Install dependencies
   - conda create -n test -c conda-forge python=$TRAVIS_PYTHON_VERSION pip
   - source activate test
-  - pip install -r requirements.txt
-  - pip install -r test_requirements.txt
+  - conda install boto3 botocore moto six pytest mock -c conda-forge -y
 
 script:
   - py.test -x --doctest-modules --pyargs s3fs
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ install:
   - conda install boto3 botocore moto six pytest mock -c conda-forge -y
 
 script:
-  - py.test -x --doctest-modules --pyargs s3fs
+  - py.test -vv s3fs
 
 notifications:
   email: false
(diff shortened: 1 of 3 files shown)


martindurant:
The failure appears to be a misalignment of versions of moto and botocore. Perhaps try setting the version of botocore installed in the `.travis.yaml` file to a few versions back? I take ti this passes for you locally, so you might want to use whatever versions of boto3 and botocore you happen to have.

drorata:
Indeed, I was unable to install the dependencies using `test_requirements.txt`. I changed it to:

```
# test_requirements.txt
boto3==1.9.38
botocore>=1.12.13
futures     ; python_version<"3.2"
mock
moto>=1.0
pytest>=2.7.3
```

Regardless (i.e. it is not a versions issue), I get a specific fail for the test I mentioned above.

martindurant:
Can you please push those versions, so that I can see your failure in the CI?
Indeed, it may be better in the future to install from conda rather than pip, in which case there would be a single alternate line in .travis.yaml

drorata:
I have just pushed the dependencies that I ended up with.

martindurant:
Seems to fix:
```
diff --git a/s3fs/core.py b/s3fs/core.py
index dd576e8..deb5d6c 100644
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -375,7 +375,7 @@ class S3FileSystem(object):
                 if dirs:
                     files.extend([{'Key': l['Prefix'][:-1], 'Size': 0,
                                   'StorageClass': "DIRECTORY"} for l in dirs])
-                files = [f for f in files if len(f['Key']) > len(prefix)]
+                # files = [f for f in files if len(f['Key']) > len(prefix)]
                 for f in files:
                     f['Key'] = '/'.join([bucket, f['Key']])
             except ClientError as e:
@@ -957,9 +957,7 @@ class S3FileSystem(object):
             raise FileNotFoundError(path)
         if recursive:
             self.invalidate_cache(path)
-            self.bulk_delete(self.walk(path), **kwargs)
-            if not self.exists(path):
-                return
+            self.bulk_delete(self.walk(path, directories=True), **kwargs)
         if key:
             try:
```

drorata:
Issue is solved locally! TravisCI still fails.

drorata:
Something is still not working as expected on CI.

martindurant:
Getting there :)
It seems moto isn't good about specifying its requirements

martindurant:
That will have to do.
Thanks, @drorata !

PR #160, opened 2018-12-05
Add text mode support for S3FileSystem.cat()

### LINKED ISSUE #159: cat() only supports binary data
Now that s3fs supports text I/O I think `cat()` should support it as well. We can't pass `mode` to the function because it'll crash:

```
>>> s3.cat('bucket/file.txt', mode='r'))                                                                        
  File "<stdin>", line 1, in <module>
  File "/Users/dargueta/.pyenv/versions/3.7.0/lib/python3.7/site-packages/s3fs/core.py", line 757, in cat
    with self.open(path, 'rb', **kwargs) as f:
TypeError: open() got multiple values for argument 'mode'

open() got multiple values for argument 'mode'
```

dargueta:
Fixes #159

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,11 @@ sudo: False
 
 language: python
 
+env:
+  global:
+    # Fixes https://github.com/travis-ci/travis-ci/issues/7940
+    - BOTO_CONFIG=/dev/null
+
 matrix:
   include:
     - python: 2.7
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -752,7 +752,11 @@ class S3FileSystem(object):
 
     def cat(self, path, **kwargs):
         """ Returns contents of file """
-        with self.open(path, 'rb', **kwargs) as f:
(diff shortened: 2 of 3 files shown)


martindurant:
I am in principle OK with your change, but I have a few comments.

- would you say that opening the file in text mode like this is exactly equivalent to doing `cat` and then `decode`? That might be a more straight-forward way of expressing it.

- "`as_text`" feels like a strange parameter name here. I would have used `decode` (True/False) or `encoding` (bytes if False, default, decode to text if otherwise). I realise there is some overlap and possible confusion with the parameters to `open`. 

- I encourage you to think about how this could be made general in the context of [fsspec](https://github.com/martindurant/filesystem_spec/blob/master/fsspec/spec.py), to which s3fs will be conformed in the near future. Note, in particular, that `open_files` (in the `core` module) already allows opening of one or more files and applies text encoding *and* possible compression on read or write.

dargueta:
You've made some good points. This probably needs to be rethought some, so I'll close it for now.

PR #161, opened 2019-01-07
Compatibility for fsspec

martindurant:
(much removed code!)

Many path/directory-specific things fail.
Text mode no longer supported directly in the class.
Mapper no longer required at all (tests can be adapted).


dvukolov:
Martin, could you please add fsspec to requirements.txt?

martindurant:
Yes, will do. First, I need to get moto working, then fix the tests... This is just an experiment for now, and s3fs based on fsspec will not be released for quite some time.

martindurant:
cc @TomAugspurger ; this is pretty close to done.

TomAugspurger:
👍 looks nice on a quick glance. Do you plan to update `setup.py` to include fsspec in a future PR? Also is this PR dropping py27 support? May want to include a `python_requires` in the setup.py if that's the case.

martindurant:
Yes it will, both those things - but it's still WIP for now. Nice to see that there is a lot of red in the changes.

martindurant:
@mariusvniekerk , would appreciate if you could subject this to heavy testing

PR #163, opened 2019-02-06
Implement isfile and isdir

gsakkis:
As per the title, implement `S3FileSystem.isfile` and `isdir`. These will be still useful even after `s3fs` is derived from [fsspec](https://github.com/martindurant/filesystem_spec). `isdir` in particular:
- cannot be delegated to an `s3.info()` call since this would raise `FileNotFoundError` 
- is relatively inexpensive regardless of the number of files in the directory, unlike `ls`.

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -469,7 +469,14 @@ class S3FileSystem(object):
         if path.startswith('s3://'):
             path = path[len('s3://'):]
         path = path.rstrip('/')
-        return not path or bool(self._lsdir(path, refresh=refresh, max_items=1))
+        if not path:
+            return True
+        if not refresh:
+            parent = path.rsplit('/', 1)[0]
+            if parent in self.dirs:
+                return any(f['Key'] == path and f['StorageClass'] == 'DIRECTORY'
+                           for f in self.dirs[parent])
+        return bool(self._lsdir(path, refresh=refresh, max_items=1))
 
     def isfile(self, path, refresh=False):
         """ Check if path points to a file.
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -355,7 +355,7 @@ class S3FileSystem(object):
             return fdesc
         return io.TextIOWrapper(fdesc, encoding=encoding)
 
-    def _lsdir(self, path, refresh=False):
+    def _lsdir(self, path, refresh=False, max_items=None):
         if path.startswith('s3://'):
             path = path[len('s3://'):]
         path = path.rstrip('/')
@@ -364,8 +364,11 @@ class S3FileSystem(object):
         if path not in self.dirs or refresh:
             try:
                 pag = self.s3.get_paginator('list_objects_v2')
+                config = {}
+                if max_items:
+                    config.update(MaxItems=max_items, PageSize=2 * max_items)
                 it = pag.paginate(Bucket=bucket, Prefix=prefix, Delimiter='/',
-                                  **self.req_kw)
+                                  PaginationConfig=config, **self.req_kw)
                 files = []
                 dirs = []
                 for i in it:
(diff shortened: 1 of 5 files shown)


martindurant:
For the failure, see https://github.com/dask/dask/pull/4276
You will need `BOTO_CONFIG=/dev/null` and probably all moto use should be via the context manager `with moto.mock_s3():` rather than using start and stop methods.

s3fs/core.py
```diff
@@ -364,8 +364,11 @@ def _lsdir(self, path, refresh=False):
         if path not in self.dirs or refresh:
             try:
                 pag = self.s3.get_paginator('list_objects_v2')
+                config = {}
+                if max_items:
+                    config.update(MaxItems=max_items, PageSize=2 * max_items)
```
martindurant: If using max_items, the cache `self.dirs` should not be updated

s3fs/core.py
```diff
@@ -450,6 +453,37 @@ def ls(self, path, detail=False, refresh=False, **kwargs):
...
+        ----------
+        path : string/bytes
+            location to check
+        refresh : bool
+            If true, don't look in the info cache
+        """
+        if path.startswith('s3://'):
+            path = path[len('s3://'):]
+        path = path.rstrip('/')
+        return not path or bool(self._lsdir(path, refresh=refresh, max_items=1))
```
martindurant: So the logic is, if it contains anything, it must be a directory. I think it would be equally had from the listing of the parent directory: it should turn up in the CommonPrefixes and be marked as a directory.

s3fs/tests/test_s3fs.py
```diff
@@ -256,6 +256,38 @@ def test_ls_touch(s3):
...
+    assert not s3.isfile(test_bucket_name + '/test')
+    assert not s3.isfile(test_bucket_name)
+
+    assert not s3.isfile(a)
+    s3.touch(a)
+    assert s3.isfile(a)
+
+    assert not s3.isfile(b)
+    s3.mkdir(b)
+    assert not s3.isfile(b)
```
martindurant: I am surprised that this assert passes - mkdir works by calling touch. Perhaps isfile only fails because the path that is passed to touch has '/' appended, i.e., `s3.isfile(b + '/') and s3.isdir(b + '/') is True`.

gsakkis:
> For the failure, see [dask/dask#4276](https://github.com/dask/dask/pull/4276)
> You will need `BOTO_CONFIG=/dev/null` and probably all moto use should be via the context manager `with moto.mock_s3():` rather than using start and stop methods.

Are you suggesting to fix this as part of this pull request? I think it should be a separate commit as it's unrelated. 

By the way all tests are passing for me locally without any change after `pip install -r test_requirements.txt`:
```
$ pip freeze | grep oto
boto==2.49.0
boto3==1.9.38
botocore==1.12.91
moto==1.3.7
```

s3fs/core.py
gsakkis: Fixed at https://github.com/dask/s3fs/pull/163/commits/61b73d35d580e3cd64beb72cece85508c35642f1

s3fs/tests/test_s3fs.py
gsakkis: Good catch, fixed and added more tests at c442c5d17edde54e8a3ea1232f5577ff8dad28e7

s3fs/core.py
gsakkis: If I understand what you mean, this would require listing the whole parent directory (no `max_items`) which is potentially more expensive.

martindurant:
Thanks @gsakkis for all the work here, good to see everything passing, I'll have another look later today.

s3fs/core.py
martindurant: Yes, it would mean listing the parent directory. There is a good chance that the parent directory is cached, though, so it could involve no download at all. Perhaps we could check for the cached listing and use it if it exists, or use this method if it does not?

gsakkis: Good idea, implemented at ceb7def3d9e897c0fe3d4774b87ed61a70dadb2b.

martindurant:
Excellent, thank you.
At some point, you may be interested in thinking about how this would work int he context of #161, where we try to farm out as much of the file-system logic to fsspec as possible; but maybe specific functionality like this would have to be overridden anyway.

PR #165, opened 2019-02-12
Fix for S3FileSystem.info()

gsakkis:


s3fs/tests/test_s3fs.py
```diff
@@ -168,6 +168,14 @@ def test_info(s3):
     assert s3.info(a) == s3.dirs[parent][0]  # correct value
     assert id(s3.info(a)) == id(s3.dirs[parent][0])  # is object from cache
 
+    new_parent = test_bucket_name + '/foo'
+    s3.mkdir(new_parent)
```
martindurant: I still feel like this should make the file `test_bucket_name + '/foo/'` (with the trailing slash), which is a real empty file, even if it is conventionally treated as a directory placeholder. It *could* contain data. 
So is the assertion that `info()` should raise because the path created is not the same as the path given, or that this should be the behaviour for all directories?

gsakkis: The issue here is that calling `info` for the same path (without trailing slash) has different behavior depending on whether the result is cached or not; the first time it raises and the second it succeeds. What the semantics of a directory should be is debatable and indeed there can be different options but at the very least the behaviour should be consistent regardless of the cache, i.e. raise on both calls or on neither.

martindurant:
OK, fair enough, let's do it this way around, but when consolidating into fsspec, the meaning of directories should become more formalised, and hopefully unambiguous :)

PR #177, opened 2019-04-12
a mmap'ed cache

blackmagic02881:
pls do not merge. just show the idea for discussion. this is a file backed cache which
* create a tmp file with lifecycle same as the object
* file get mmap as bytestream. data get pulled from remote on 1st cache miss. 

pros
* larger cache, better performance on random access.
cons
* cache lost as soon as object released
* accessing every part of a large object could lead to OOS in tmp directory.

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -6,6 +6,8 @@ import logging
 import os
 import re
 import socket
+import tempfile
+import mmap
 from hashlib import md5
 
 import boto3
@@ -302,7 +304,7 @@ class S3FileSystem(object):
         self._kwargs_helper = ParamKwargsHelper(self.s3)
 
     def open(self, path, mode='rb', block_size=None, acl='', version_id=None,
-             fill_cache=None, encoding=None, **kwargs):
+             fill_cache=None, encoding=None, file_backed_cache=False, **kwargs):
         """ Open a file for reading or writing
 
         Parameters
@@ -347,6 +349,7 @@ class S3FileSystem(object):
         mode2 = mode if 'b' in mode else (mode.replace('t', '') + 'b')
         fdesc = S3File(self, path, mode2, block_size=block_size, acl=acl,
                        version_id=version_id, fill_cache=fill_cache,
+                       file_backed_cache=file_backed_cache,
                        s3_additional_kwargs=kw)
         if 'b' in mode:
             return fdesc
@@ -1145,7 +1148,8 @@ class S3File(object):
     """
 
     def __init__(self, s3, path, mode='rb', block_size=5 * 2 ** 20, acl="",
-                 version_id=None, fill_cache=True, s3_additional_kwargs=None):
+                 version_id=None, fill_cache=True, file_backed_cache=False,
+                 s3_additional_kwargs=None):
         self.mode = mode
         if mode not in {'rb', 'wb', 'ab'}:
             raise NotImplementedError("File mode must be {'rb', 'wb', 'ab'}, "
@@ -1163,7 +1167,8 @@ class S3File(object):
         self.start = None
         self.end = None
         self.closed = False
-        self.trim = True
+        self.file_backed_cache = file_backed_cache
+        self.trim = not file_backed_cache
         self.mpu = None
         self.version_id = version_id
         self.acl = acl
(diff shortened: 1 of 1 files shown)


martindurant:
An interesting idea, as I said before. Would be nice to see what kind of performance you get here - presumably depends on the type of disc and OS caching policy (so complicated to measure); also, I suppose memory use does not necessarily stay low if large parts of a large file are accessed. Do you know what are the system requirements for sparse file support and mem-mapping?

Am I right in thinking that you split the file into block-sized pieces and keep track, in memory, of which blocks have been seen?

More comments to come on the code itself.

s3fs/core.py
```diff
@@ -1339,7 +1356,7 @@ def _fetch(self, start, end):
         if start < self.start:
             if not self.fill_cache and end + self.blocksize < self.start:
                 self.start, self.end = None, None
-                return self._fetch(start, end)
```
martindurant: No need to change these - you made sure that `_fetch` does the same thing as before when not mmapping.

s3fs/core.py
```diff
@@ -1475,7 +1516,12 @@ def close(self):
         """
         if self.closed:
             return
-        self.cache = None
+
+        if self.file_backed_cache:
+            self.cache.close()
```
martindurant: If you just do `self.cache = None`, so the reference drops, doesn't the file get cleaned up anyway upon garbage collection?

s3fs/core.py
```diff
@@ -1209,6 +1214,16 @@ def __init__(self, s3, path, mode='rb', block_size=5 * 2 ** 20, acl="",
...
                 raise_from(IOError("File not accessible", path), e)
+                if self.file_backed_cache:
+                    fd = tempfile.TemporaryFile()
+                    fd.seek(self.size - 1)
+                    fd.write(b'1')
+                    fd.flush()
+                    f_no = fd.fileno()
+                    self.start = 0
+                    self.end = self.size
+                    self.cache = mmap.mmap(f_no, self.size)
```
martindurant: Note that the windows version of mmap will automatically extend the file size here, rather than doing the sparse trick (but this might take up the fill space of the file - to be tried). I suspect the sparse trick might not work on windows, you'd get "write past EOF" error or similar.

s3fs/core.py
```diff
@@ -1349,13 +1366,37 @@ def _fetch(self, start, end):
...
+    def _file_backed_cache_fetch(self, start, end):
+        if self.s3.version_aware:
+            version_id = self.version_id
+        else:
+            version_id = None
+        if end > self.end:
+            end = self.end
+        start_block = start // self.blocksize
+        end_block = end // self.blocksize
+        for i in range(start_block, end_block + 1):
```
martindurant: This loop may be wasteful if requesting contiguous blocks, often the overhead of establishing the connection it comparable to the download time; instead of doing a fetch for each block, they should be combined where possible.

s3fs/core.py
```diff
@@ -1296,6 +1311,8 @@ def readline(self, length=-1):
 
         If length is specified, at most size bytes will be read.
         """
+        if self.file_backed_cache:
+            raise ValueError('readline not available in file backed cache mode')
```
martindurant: I notice that the mmap API supports readline directly https://docs.python.org/3/library/mmap.html#mmap.mmap.readline

blackmagic02881:
thanks for the comments. yes, it is just to show the idea that how a mmap one could looks like. performance wise, it highly depends on the workload. we have a pattern that it will read a portion of block X, then seek to read block Y. then later it comes back to block X again. the current cache with trim will not be able to deal with this. and that is why i proposed this mmap idea. 

i like the idea of extracting this into a separate caching manager so it could be extensible.

s3fs/core.py
```diff
@@ -1209,6 +1214,16 @@ def __init__(self, s3, path, mode='rb', block_size=5 * 2 ** 20, acl="",
                     self.version_id = info.get('VersionId')
             except (ClientError, ParamValidationError) as e:
                 raise_from(IOError("File not accessible", path), e)
+                if self.file_backed_cache:
```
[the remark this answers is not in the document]
blackmagic02881: yes it was wrong. a c&p error.

s3fs/core.py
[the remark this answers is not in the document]
blackmagic02881: yes

s3fs/core.py
blackmagic02881: true.

s3fs/core.py
blackmagic02881: What is windows?  lol. sorry i only have mac and linux at hand.

s3fs/core.py
blackmagic02881: totally agreed. i just want to show how it works it really should check and merge small requests like this to be as large as possible.

blackmagic02881:
that is why up front i said this code is not for merged but to show the idea. i think original s3fs need some refactoring before adding any code like this.

martindurant:
Implemented in fsspec, no need to duplicate here

PR #182, opened 2019-05-02
Add errors parameter to open function

akaguy:
Add errors parameter to open function, which is passed to TextIOWrapper. this will help handle cases of decode errors in utf-8 etc..

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -302,7 +302,7 @@ class S3FileSystem(object):
         self._kwargs_helper = ParamKwargsHelper(self.s3)
 
     def open(self, path, mode='rb', block_size=None, acl='', version_id=None,
-             fill_cache=None, encoding=None, **kwargs):
+             fill_cache=None, encoding=None, errors=None, **kwargs):
         """ Open a file for reading or writing
 
         Parameters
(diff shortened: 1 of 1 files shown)


martindurant:
This is fine, but could do with a test.
Once more, though, I should point out that this functionality is slated to move to fsspec, so I would appreciate your opinion on whether its `open` does the right thing for text, from your point of view.

martindurant:
It seems like moto has updated it version ID strategy. The following should fix the failing test
```diff
--- a/s3fs/tests/test_s3fs.py
+++ b/s3fs/tests/test_s3fs.py
@@ -1095,11 +1095,11 @@ def test_versions(s3):
     assert len(versions) == 2

     with s3.open(versioned_file) as fo:
-        assert fo.version_id == '1'
+        assert fo.version_id == versions[1]
         assert fo.read() == b'2'

-    with s3.open(versioned_file, version_id='0') as fo:
-        assert fo.version_id == '0'
+    with s3.open(versioned_file, version_id=versions[0]) as fo:
+        assert fo.version_id == versions[0]
         assert fo.read() == b'1'
```

martindurant:
Ping - are you intending to finish off this PR, @akaguy ?

martindurant:
Implemented in fsspec

PR #183, opened 2019-05-07
support for large writes & test updates

jesteria:
#### ensure multi-parts written by ``flush()`` respect ``block_size`` and S3 max part size

1) ``flush()`` was properly ensuring that the buffer had reached at
least ``block_size`` in length before writing a part to S3; however, it
was not ensuring that resulting part(s) were themselves at all on the
order of ``block_size``.

2) Nor did the method ensure every part is less than S3's maximum part
size, (regardless of desired ``block_size``).

---

(1) was relatively "easy" to cross -- if a single ``write()`` payload
was of *any* size equal to or greater than ``buffer_size`` -- even many
times larger than ``buffer_size`` -- it was uploaded as is, as a single
part. This itself caused ``flush()`` to violate the expectation of
``buffer_size``, and to operate differently depending on whether the
user wrote their payload in small chunks or big chunks.

Moreover, (1) caused a breaking issue, when a single write payload was
greater than S3's maximum upload (part) limit of 5 GiB. This is a large
string to write, certainly; however, not an unreasonable one, when your
objects are that large -- (and, S3 objects can be much larger than
that).

(For example, see dssg/triage#687 – a pre-existing interface was
extended to enforce proper chunking of 5+ GiB write payloads.)

(1) is fixed by making ``flush()`` read and upload ``block_size`` chunks
from the buffer, (only writing a larger part than that at the end of the
buffer, in order to avoid a smaller-than-``buffer_size`` part).

---

(2) is even more of an edge case, but likely worth handling, at the same
time -- while ``buffer_size`` is (hypothetically) configurable, and up
to the configurer to ensure that it is neither too small nor too large,
so long as ``flush()`` is ever writing parts *larger* than
``buffer_size``, then it is still conceivable for it to request to
upload a part that is larger than S3's maximum. However, this
implies that ``buffer_size`` has been configured to be far larger than
S3's *minimum* -- really, it must be more than half as large as the
maximum -- and therefore ``buffer_size`` may be safely, and trivially,
violated, by uploading the remainder of the buffer in two equal-size
parts, (and satisfying the part size minimum for both).

#### update test ``test_url`` to silence warning

Raw string literal required for regex special characters

#### update test ``test_versions`` for moto v1.3.8

Having created a fresh testing environment, I got moto v1.3.8, which
returns different file version information, requiring a test update.

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -741,7 +741,6 @@ class S3FileSystem(object):
         refresh : bool (=False)
             if False, look in local cache for files
         """
-        path0 = path
         if path.startswith('s3://'):
             path = path[len('s3://'):]
         path = path.rstrip('/')
@@ -1142,9 +1141,12 @@ class S3File(object):
     See Also
     --------
     S3FileSystem.open: used to create ``S3File`` objects
-    """
 
-    def __init__(self, s3, path, mode='rb', block_size=5 * 2 ** 20, acl="",
+    """
+    part_min = 5 * 2 ** 20
+    part_max = 5 * 2 ** 30
+
+    def __init__(self, s3, path, mode='rb', block_size=part_min, acl="",
                  version_id=None, fill_cache=True, s3_additional_kwargs=None):
         self.mode = mode
         if mode not in {'rb', 'wb', 'ab'}:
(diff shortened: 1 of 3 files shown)


s3fs/tests/test_s3fs.py
```diff
@@ -1092,14 +1140,15 @@ def test_versions(s3):
     with s3.open(versioned_file, 'wb') as fo:
         fo.write(b'2')
     versions = s3.object_version_info(versioned_file)
-    assert len(versions) == 2
+    version_ids = [version['VersionId'] for version in versions]
+    assert len(version_ids) == 2
 
     with s3.open(versioned_file) as fo:
-        assert fo.version_id == '1'
+        assert fo.version_id == version_ids[1]
```
martindurant: I'm sure I already wrote this fix too!

s3fs/core.py
```diff
@@ -1433,38 +1435,63 @@ def flush(self, force=False, retries=10):
...
+                if 0 < data1_size < self.blocksize:
+                    # last part would be smaller than specified
+                    remainder = data0 + data1
+                    remainder_size = self.blocksize + data1_size
+
+                    if remainder_size <= self.part_max:
+                        # we can combine the last two parts without issue
+                        (data0, data1) = (remainder, None)
+                    else:
+                        # combination of last two blocks is too large for a single part
```
martindurant: This is the extreme edge case?

s3fs/core.py
```diff
@@ -1433,38 +1435,63 @@ def flush(self, force=False, retries=10):
...
-
             try:
                 self.mpu = self.mpu or self._call_s3(
                     self.s3.s3.create_multipart_upload,
                     Bucket=self.bucket, Key=self.key, ACL=self.acl)
-            except (ClientError, ParamValidationError) as e:
-                raise_from(IOError('Initiating write failed: %s' %
-                                   self.path), e)
+            except (ClientError, ParamValidationError) as exc:
+                raise_from(IOError('Initiating write failed: %s' % self.path), exc)
```
martindurant: Why the reformatting? I notice this in a number of places, perhaps it was automatically done by your IDE?

s3fs/core.py
```diff
@@ -741,7 +741,6 @@ def glob(self, path, refresh=False):
         refresh : bool (=False)
             if False, look in local cache for files
         """
-        path0 = path
```
martindurant: I take it this was unused?

s3fs/tests/test_s3fs.py
jesteria: Yep, I think I saw that in another PR, (but not in master, I don't think).

s3fs/core.py
jesteria: Indeed. I believe you'd have to set the `blocksize` to more than 2.5 GiB in order to hit it. But, so long as that's configurable, and so long as we're deciding what size parts to send here, it seems wise to handle.

s3fs/core.py
jesteria: Yep, I was uncertain what to do about that stuff – the `vim-flake8` plugin, (as configured of course), complained about a couple things: one-letter variables and unused variables. (And then I'm sure other settings resulted in different formatting.)

But, certainly, I'm happy to narrow the scope of changes, to exclude such things, in so far as they're unrelated to support for large writes.

s3fs/core.py
jesteria: (Yep.)

martindurant:
I am happy with this, thanks for putting in the work here.

PR #195, opened 2019-07-03
Add conformant walk() implementation, rename old one to find()

### LINKED ISSUE #107: Walk method doesn't match python standard lib interface
The implementation of walk in `s3fs` returns a flat list of all paths encountered when traversing the tree. In contrast, `os.walk` from the standard library returns an iterator of `dirname, dirs, files`, where `dirname` is the current directory prefix in the traversal, and `fils` and `dirs` are the names (relative, not absolute) of all files and directories respectively.

This difference in behavior is confusing for those familiar with the standard `os.walk` method, and also makes providing the interface used by `pyarrow` (which does match the standard library) tricky.

Pyarrow has created a wrapper class that provides the proper method interface, see [here](https://github.com/apache/arrow/blob/master/python/pyarrow/filesystem.py#L319-L349). Is there any reason not to just upstream this? We'd potentially need to figure out a good deprecation path, as the current behavior *may* be used by users.

dargueta:
Closes #107.

**Breaking change:**

* `walk()` now behaves like os.walk() instead of returning a list of all files
* The old `walk()` was renamed to `find()`.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
 boto3>=1.9.91
 botocore>=1.12.91
 fsspec>=0.2.2
+more-itertools>=6.0
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -2,12 +2,14 @@
 from six import raise_from
 import errno
 import logging
+import posixpath
 import socket
 from hashlib import md5
 
 from fsspec import AbstractFileSystem
 from fsspec.spec import AbstractBufferedFile
 import boto3
+import more_itertools as m_iter
 from botocore.client import Config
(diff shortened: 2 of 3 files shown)


martindurant:
You are completely right, but this was recently addressed by the switch to depending on fsspec and its [glob](https://github.com/intake/filesystem_spec/blob/master/fsspec/spec.py#L398)  (with improvements in https://github.com/intake/filesystem_spec/pull/64). Funny that we both chose "find" as the alternate name for the old behaviour!

dargueta:
Ah, well. Never mind. Should 107 be closed then?

> Funny that we both chose "find" as the alternate name for the old behaviour!

I just looked at the ticket. 🙂

PR #196, opened 2019-07-03
Make boto3 exceptions more accurate

dargueta:
Right now, if you try reading a file with expired credentials, you get a `FileNotFoundError`. This is pretty confusing. Now that we support Python 3 only, we can be *much* more specific about the kinds of errors we're throwing.

I pulled a list of all the S3 error codes from [here](https://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html) and made approximations of what the most logical error code could be.

### Bugfixes

* When attempting to delete a nonexistent bucket, an `IOError` was thrown with the message "not empty." The proper exception is a `FileNotFoundError` because the bucket wasn't found.
* Fixed a number of `IOError` calls that were passing invalid arguments. According to [the documentation](https://docs.python.org/3/library/exceptions.html#OSError), if multiple arguments are passed, the first argument is [an errno code](https://docs.python.org/3/library/errno.html) and the second argument is an error message. Passing invalid arguments can break handler code that's expecting an integer as the error code, not a string.
* Some logging calls were missing arguments to the string formatter.
* Some logging calls used `%e` to display the error message. `%e` is used for floating-point numbers, so I changed this to `%r`.

### Breaking(ish) changes

* `ParamValidationError` is now rethrown as a `ValueError`. Before it was always rethrown as an `IOError` even though this was a programmer error, not an I/O problem.
* `FileNotFoundError` is no longer defined in `core.py`. Since this is Python 3 only now, the only code that should be importing it is Python 2 code, which would immediately break anyway because this can't be installed.

### Other changes

Removed legacy `raise_from` since it's no longer necessary. `s3fs` no longer relies on `six`.

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -25,17 +25,6 @@ except ImportError:
         socket.timeout,
     )
 
-
-# py2 has no ConnectionError only OSError with different error number for each
-# error so we need to catch OSError and compare errno to the following
-# error numbers - same as catching ConnectionError on py3
-# ref: https://docs.python.org/3/library/exceptions.html#ConnectionError
-_CONNECTION_ERRORS = frozenset({
-    errno.ECONNRESET,  # ConnectionResetError
-    errno.EPIPE, errno.ESHUTDOWN,  # BrokenPipeError
-    errno.ECONNABORTED,  # ConnectionAbortedError
-    errno.ECONNREFUSED,  # ConnectionRefusedError
-})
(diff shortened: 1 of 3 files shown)


dargueta:
Don't merge yet, still got a few things I need to fix

martindurant:
> Don't merge yet,

OK, let me know. Looking good so far.

dargueta:
Okay I think this is good to go now.

martindurant:
OK, I think this looks great, thanks for doing such a thorough job!

Issue #200, opened 2019-07-18
Unnecessary _parent() call


nmodhipallise:
https://github.com/dask/s3fs/blob/b7638b52f98d819960426c05fc9b72862165105b/s3fs/core.py#L487

I think the `_parent(path)` call is not necessary here. `super().info()` already does this:
https://github.com/intake/filesystem_spec/blob/8175fbdf963fa459f3f69253356b0d3e82dd9d37/fsspec/spec.py#L470

Because of the dual `_parent()` call,
`S3FileSystem.info("bucket/somefile.ext")`
will cause:
`S3FileSystem._ls("")`
which causes:
`S3FileSystem._lsbuckets()`
while it should actually be doing:
`S3FileSystem._lsdir("bucket")`

martindurant:
Hmmm. You are right, that listing two directories up is not useful.
However, if you simply remove it, then 
```
S3FileSystem.ls("bucket/somefile.ext") != [S3FileSystem.info("bucket/somefile.ext")]
```
as it should

PR #203, opened 2019-07-23
listbuckets should return empty when permission denied

martindurant:
There are two problems. One is #200 , which I have not decided how to solve yet. However, this should get things working OK for you, @AntonFriberg . Would you mind checking?

Fixes https://github.com/dask/dask/issues/5141


AntonFriberg:
@martindurant Tested and works as expected! Thanks a lot for the quick fix!

PR #204, opened 2019-07-23
Only list parent in ls if files were not found

### LINKED ISSUE #200: Unnecessary _parent() call

martindurant:
The output doesn't change, we just prevent the extra lookup if it is not
needed

Fixes #200

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -488,10 +488,11 @@ class S3FileSystem(AbstractFileSystem):
             additional arguments passed on
         """
         path = self._strip_protocol(path).rstrip('/')
-        out = self._ls(self._parent(path), refresh=refresh)
-        out = [o for o in out
-               if o['name'].rstrip('/') == path and o['type'] != 'directory']
-        files = self._ls(path, refresh=refresh) or out
+        files = self._ls(path, refresh=refresh)
+        if not files:
+            files = self._ls(self._parent(path), refresh=refresh)
+            files = [o for o in files if o['name'].rstrip('/') == path
+                     and o['type'] != 'directory']
         if detail:
             return files
         else:


drorata:
I'm not sure it is related to this PR, but it seems like it introduced a confusing behavior. 

Assume:
```
my_bucket
  |
  |- dir1
  |     |- some_key
  |     |- some_key2
  |- dir2
``` 

the `s3.ls("my_bucket/dir1")` would return `["my_bucket/dir1/some_key", "my_bucket/dir1/some_key2"]`, but `s3.ls("my_bucket/dir2")` returns `["my_bucket/dir2"]`. I would expect the latter case to return an empty list. Or am I missing something?

martindurant:
Indeed: what is "dir2"? We cannot emphasise the point enough, S3 does not have dictionaries. In your diagram, apparently "dir2" is a key, and ought to be listed. "dir1" does not need to exist as a key, but might do. If it doesn't exist, it is implied by the existence of the deeper keys.

drorata:
So `dir2` is a key but `dir1` is *not*, is that right?

martindurant:
The way you have shown the tree, `dir2` must be a key, and `dir1` does not need to be a key, but might be.

PR #208, opened 2019-07-30
Specify an s3uri as {bucket}/{key}?versionId={versionId}

mariusvniekerk:
This should in principal allow us to do things like

```
dask.dataframe.read_parquet([
    's3://bucket/somepath/file_part0001.parquet?versionId=ksfgsdkfkjadsf',
    's3://bucket/somepath/file_part0002.parquet?versionId=lijserslifhsdiu',
])
```

This should work for most things that make sense.

Not everything has proper test cases yet.

### DIFF
--- a/s3fs/tests/test_s3fs.py
+++ b/s3fs/tests/test_s3fs.py
@@ -1108,6 +1108,26 @@ def test_list_versions_many(s3):
     assert len(versions) == 1200
 
 
+def test_fsspec_versions_multiple(s3):
+    """Test that the standard fsspec.core.get_fs_token_paths behaves as expected for versionId urls"""
+    s3 = S3FileSystem(anon=False, version_aware=True)
+    versioned_file = versioned_bucket_name + '/versioned_file3'
+    version_lookup = {}
+    for i in range(20):
+        contents = str(i).encode()
+        with s3.open(versioned_file, 'wb') as fo:
+            fo.write(contents)
+        version_lookup[fo.version_id] = contents
+    urls = [f"s3://{versioned_file}?versionId={version}" for version in version_lookup.keys()]
+    fs, token, paths = fsspec.core.get_fs_token_paths(urls)
+    assert isinstance(fs, S3FileSystem)
+    assert fs.version_aware
+    for path in paths:
+        with fs.open(path, 'rb') as fo:
+            contents = fo.read()
+            assert contents == version_lookup[fo.version_id]
+
+
 def test_versioned_file_fullpath(s3):
     versioned_file = versioned_bucket_name + '/versioned_file_fullpath'
     s3 = S3FileSystem(anon=False, version_aware=True)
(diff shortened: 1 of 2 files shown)


s3fs/tests/test_s3fs.py
```diff
@@ -1096,6 +1098,27 @@ def test_list_versions_many(s3):
     assert len(versions) == 1200
 
 
+def test_versioned_file_fullpath(s3):
+    versioned_file = versioned_bucket_name + '/versioned_file_fullpath'
+    s3 = S3FileSystem(anon=False, version_aware=True)
```
martindurant: Isn't this per-bucket?

mariusvniekerk: This is a per bucket setting, but we are not detecting anywhere right now in s3fs whether a particular bucket has versioning enabled,  consequently it is just mostly used blanketly (similar in the other versioned tests)

mariusvniekerk:
> * there is a general need across all filesystems to be able to specify more than just a URL to identify a bunch of bytes. In principle, I am OK with including a query parameter like this - as far as I know, only actual HTTP uses this for anything at the moment. It would be better if it was a known syntax elsewhere that we could refer to, though.

Let me know where in fsspec we can define with this extended url format could look like.

Reasonably okay with changing the split_path method back, but....

```
```

is no longer guaranteed to actually return a `key` or a valid prefix.


Should these splitting methods be defined more explicitly in `fsspec`?

> * we don't actually test the path that this was set to solve, `fsspec.open_files` with versioning. Do the query fragments survive passing through `fsspec.utils.infer_storage_options`?

Yes, That is still a bit of a TODO.

> * how will this interact with other invokations of `open_files`, e.g., with `mode='wb'`, or paths with globs? Do we need to explicitly raise?

`w` of any kind should raise, since you cannot control how s3 assigns version ids.   
`globs` are largely meaningless with versions since they apply to only a single key.  There exists no fast way to get a file within a prefix with a given versionId.

mariusvniekerk:
@martindurant green now, ready for rereview

mariusvniekerk:
Whoops the last one we me accidentally posting as another user.

mariusvniekerk:
@martindurant This is ready for review again

martindurant:
Is `info(..., version_id=)` now deprecated, then?

martindurant:
ping @mariusvniekerk

mariusvniekerk:
@martindurant This right now still uses that method to get things like length and etag.  If a uri AND a version_id are given and they mismatch we should probably raise.

martindurant:
> we should probably raise.

Please!

mariusvniekerk:
@martindurant Added this functionality

martindurant:
ooops...
```
verssion_ids.remove(None)
```
(extra s)

martindurant:
`if len(version_ids > 1):`
->
`if len(version_ids) > 1:`

mariusvniekerk:
> `if len(version_ids > 1):`
> ->
> `if len(version_ids) > 1:`

The amount of times i've written that....

martindurant:
Do remind me to release this sometime.

PR #209, opened 2019-07-31
Fix issue for exists('bucket/prefix/') when versioning enabled

mariusvniekerk:


mariusvniekerk:
Building a newer version of moto on conda-forge.  That should hopefully fix the issues seen here

martindurant:
Please let me know when moto is behaving

mariusvniekerk:
Moto 1.3.12 just built.  Should clear the CDN in about 30 minutes

s3fs/core.py
```diff
@@ -923,6 +928,10 @@ def __init__(self, s3, path, mode='rb', block_size=5 * 2 ** 20, acl="",
                 self.size = self.details['size']
             elif self.fs.version_aware:
                 self.version_id = self.details.get('VersionId')
+                if self.version_id is None:
+                    self.fs.invalidate_cache(self.path)
+                    self.details = self.fs.info(self.path)
```
martindurant: I am a bit puzzled that this is necessary - so the file listing was out of date? Or is it that we only get versions on `info()` (i.e., for the one specific file) and not `ls()` ? If that guess is correct, then please add a comment here to explain. Does this update the listings cache, so that opening again avoids this step? Should this go into `info()` instead? 
By the way, does `info(refresh=True)` not work, instead of explicitly calling `invalidate_cache`?

s3fs/tests/test_s3fs.py
```diff
@@ -519,14 +534,19 @@ def test_read_keys_from_bucket(s3):
                 s3.cat('s3://' + '/'.join([test_bucket_name, k])))
 
 
+@pytest.mark.skip(reason="misbehaves in modern versions of moto?")
```
martindurant: Should be xfail - I would expect it to run, and hopefully if will xpass in the future

s3fs/core.py
mariusvniekerk: The main issue stems from the fact that we only get versions on info not on ls

mariusvniekerk: Also `def info` does not take a refresh kwarg

martindurant: It takes arbitrary kwargs, so it could pass it on.
To be sure: is there no way to get s3 to include version information when doing a full listing? I think that's what I wanted to ask.

It would still be good to cache the version info, if possible, so `info()` could have, rather than refresh (i.e., destroy cache) a mode specifically for this, in the case we have versioning and the entry in the cache has no version info yet. Is that making things too complicated?

mariusvniekerk: https://github.com/dask/s3fs/blob/master/s3fs/core.py#L446

pretty sure it doesn't take kwargs at least atm.  It should in the future.

Unfortunately `list_objects_v2` does not return a `VersionId` so the only method that will do so in bulk is `list_object_versions` which will return ALL versions (which will be far too big in the general case)

Issue #213, opened 2019-08-06
Error when using with pd.read_parquet with threading on


mtrbean:
Since updating to latest version I randomly get errors like this:
```python
Traceback (most recent call last):
  File "print_parquet_metadata.py", line 28, in <module>
    df = pd.read_parquet(args.file, use_threads=True)
  File "/Users/mtrbean/.pyenv/versions/stats3/lib/python3.6/site-packages/pandas/io/parquet.py", line 282, in read_parquet
    return impl.read(path, columns=columns, **kwargs)
  File "/Users/mtrbean/.pyenv/versions/stats3/lib/python3.6/site-packages/pandas/io/parquet.py", line 129, in read
    **kwargs).to_pandas()
  File "/Users/mtrbean/.pyenv/versions/stats3/lib/python3.6/site-packages/pyarrow/parquet.py", line 1216, in read_table
    use_pandas_metadata=use_pandas_metadata)
  File "/Users/mtrbean/.pyenv/versions/stats3/lib/python3.6/site-packages/pyarrow/parquet.py", line 216, in read
    use_threads=use_threads)
  File "pyarrow/_parquet.pyx", line 1086, in pyarrow._parquet.ParquetReader.read_all
  File "pyarrow/error.pxi", line 87, in pyarrow.lib.check_status
pyarrow.lib.ArrowIOError: Unexpected end of stream: Page was smaller (26292) than expected (59582)
```
I'm pretty sure that the file is not corrupted because sometimes it can be read successfully, and it can also be read when the file is local.

It is also very hard to reproduce consistently but if I try to read the file in a loop, I realize a pattern:

This will succeed 30 times in a row without a problem 
```python
f = "s3://<bucket>/<parquet_file>"
for i in range(30):
    pd.read_parquet(f, use_threads=False)
```

This will fail sometimes on the first iteration, sometimes on the second, and never got to iteration number 5:
```python
f = "s3://<bucket>/<parquet_file>"
for i in range(30):
    pd.read_parquet(f, use_threads=True)
```

I suspect that it has something to do with random access with multiple threads, and the cache is not handling it correctly.

Finally this is the debug log which I hope will help
```
DEBUG:urllib3.util.retry:Converted retries value: False -> Retry(total=False, connect=None, read=None, redirect=0, status=None)
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): mybucket.s3.us-west-2.amazonaws.com
[...]

PR #214, opened 2019-08-07
allow setting default S3File's cache_type in S3FileSystem

mtrbean:
Related to #213. The motivation is that sometimes we want to turn off caching by default (e.g. when reading parquet files with threads)

Since different fsspec implementations overrides `_open()` differently, I've chosen to add `default_cache_type` to `s3fs` instead of `fsspec`

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -1,5 +1,4 @@
 # -*- coding: utf-8 -*-
-import errno
 import logging
 import socket
 from hashlib import md5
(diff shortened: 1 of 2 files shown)


s3fs/core.py
```diff
@@ -100,12 +99,15 @@ class S3FileSystem(AbstractFileSystem):
...
-    default_block_size: None, int
+    default_block_size: int (None)
         If given, the default block size value used for ``open()``, if no
         specific value is given at all time. The built-in default is 5MB.
     default_fill_cache : Bool (True)
         Whether to use cache filling with open by default. Refer to
         ``S3File.open``.
+    default_cache_type : string ('bytes')
+        "bytes", "mmap" or "none". If given, the default cache_type value used
+        for ``open()``. The default is 'bytes'.
```
martindurant: (perhaps refer to the fsspec doc or module, since there could be more cache types later)

martindurant:
Looks good, just the one nitpick remark

s3fs/core.py
mtrbean: updated wordings as you suggested

martindurant:
Thank you!

PR #224, opened 2019-08-28
API/ENH: Validate uploaded content

### LINKED ISSUE #103: Write consistency check
Counterpart to https://github.com/dask/gcsfs/pull/20

Note that for S3, the ETag is an MD5 hash for simple uploads, and an MD5 hash of concatenated MD5 hashes in the case of multi-part uploads https://stackoverflow.com/questions/6591047/etag-definition-changed-in-amazon-s3

TomAugspurger:
Validates that the data received by S3 matches what we send.

This matches the API from gcsfs. 3 options: md5, size, none. The default is md5.

One thing I don't understand: doesn't S3 validate `size` by default, and there's no way to disable it? From https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.upload_part

> ContentLength (integer) -- Size of the body in bytes. This parameter is useful when the size of the body cannot be determined automatically.

I guess it Boto should be able to determine the length of a bytestring?

Closes #103

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+import base64
 import logging
 import socket
 from hashlib import md5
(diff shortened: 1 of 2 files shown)


mariusvniekerk:
So for this on s3 for validation we can compute the ETag and compare it with the aws generated version.
https://stackoverflow.com/questions/12186993/what-is-the-algorithm-to-compute-the-amazon-s3-etag-for-a-file-larger-than-5gb
This is the basic algorithm used to generate them

martindurant:
Closing as stale. We could still come back to this.

Issue #225, opened 2019-08-28
EOFError with Gzip read


mccarthyryanc:
I created an [issue](https://github.com/pandas-dev/pandas/issues/28206) on the Pandas repo, but it looks like it might be an s3fs, versions 0.3.0 and greater, error.

```python
import pandas as pd
data = pd.read_csv("s3://bucketname/file.csv.gz")
```

Gives the following error:
```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/ubuntu/miniconda3/envs/test/lib/python3.7/site-packages/pandas/io/parsers.py", line 685, in parser_f
    return _read(filepath_or_buffer, kwds)
  File "/home/ubuntu/miniconda3/envs/test/lib/python3.7/site-packages/pandas/io/parsers.py", line 463, in _read
    data = parser.read(nrows)
  File "/home/ubuntu/miniconda3/envs/test/lib/python3.7/site-packages/pandas/io/parsers.py", line 1154, in read
    ret = self._engine.read(nrows)
  File "/home/ubuntu/miniconda3/envs/test/lib/python3.7/site-packages/pandas/io/parsers.py", line 2059, in read
    data = self._reader.read(nrows)
  File "pandas/_libs/parsers.pyx", line 881, in pandas._libs.parsers.TextReader.read
  File "pandas/_libs/parsers.pyx", line 896, in pandas._libs.parsers.TextReader._read_low_memory
  File "pandas/_libs/parsers.pyx", line 950, in pandas._libs.parsers.TextReader._read_rows
  File "pandas/_libs/parsers.pyx", line 937, in pandas._libs.parsers.TextReader._tokenize_rows
  File "pandas/_libs/parsers.pyx", line 2124, in pandas._libs.parsers.raise_parser_error
  File "/home/ubuntu/miniconda3/envs/test/lib/python3.7/_compression.py", line 68, in readinto
    data = self.read(len(byte_view))
  File "/home/ubuntu/miniconda3/envs/test/lib/python3.7/gzip.py", line 482, in read
    raise EOFError("Compressed file ended before the "
EOFError: Compressed file ended before the end-of-stream marker was reached
```

### Setup
Installed Pandas and s3fs via pip:
```bash
pip install pandas s3fs
```
**Pandas version:** 0.25.1
**s3fs version:** 0.3.3

mccarthyryanc:
Looks like the problem is unrelated to pandas:
```python
import gzip
import s3fs
[...]

PR #226, opened 2019-08-29
Locks down version of fsspec for #225

shawnzhu:
Applies workaround from https://github.com/dask/s3fs/issues/225#issuecomment-525990684 (Thanks to @meneal)

this fix can be applied while pending on a fix from `fsspec`.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
 boto3>=1.9.91
 botocore>=1.12.91
-fsspec>=0.2.2
+fsspec>=0.2.2,<=0.4.1


TomAugspurger:
Generally, libraries shouldn't pin upper bounds. I'll work on a fix for fsspec instead.

shawnzhu:
> Generally, libraries shouldn't pin upper bounds.

I agree. this is just a tmp workaround to unblock my (or anyone else who uses `s3fs`) work if we can get a fix soon. expecting https://github.com/intake/filesystem_spec/pull/121

TomAugspurger:
If you need a temporary fix you can pin it in your application's
dependencies.

TomAugspurger:
Not needed after #229

PR #240, opened 2019-09-20
Reinstate PUT for files less than one blocksize

martindurant:
Companion to https://github.com/intake/filesystem_spec/pull/146


martindurant:
@mariusvniekerk might have some thoughts here

PR #243, opened 2019-10-10
Always use multipart uploads when not autocommitting

mariusvniekerk:
When not autocommitting the mpu that is created is the best way to
ensure that transactional style commits can be performed elsewhere.

Alternatively we could add another flag to S3File that controls this


mariusvniekerk:
Blegh, guess its not as simple as i hoped.

martindurant:
Should there be a test that the is/isn't a MPU for autocommit off/on cases?

PR #246, opened 2019-10-17
Declare support for Python 3.8

Jamim:
Hello,

Python 3.8 is there, so I believe that it would be nice to declare support and add tests for it on CI.

Python 3.8.0 release announcement:
https://discuss.python.org/t/python-3-8-0-is-now-available/2478

These changes also:
 - make tests use [tox](https://tox.readthedocs.io/en/latest/)
 - enable testing for Python 3.5
 - simplify `.travis.yml`
 - remove redundant `test_requirements.txt`
 - add `.venv/` to `.gitignore`

Best regards!


TomAugspurger:
We install test dependencies from conda. We’ll probably wait for conda packages and then just add a 3.8 job.

Jamim:
> We install test dependencies from conda.

Is there a reason to use `conda` instead of `pip` (PyPI) on CI?

TomAugspurger:
It's what we use now, and changing things up in the CI is always risky. I'd
prefer to just wait for conda-forge to package 3.8, which should be done in
the next couple days.

martindurant:
(in some cases, it would be reasonable to have CI runs with both conda and pip; I don't think this is one)

mrocklin:
Most of our users use Conda, so we should probably test with conda.

Jamim:
> Most of our users use Conda

It doesn't look so.

[PyPI stats](https://pypistats.org/packages/s3fs):
> Downloads **last month**: 2,414,836

[Conda stats](https://anaconda.org/conda-forge/s3fs):
> 596,802 **total** downloads

mrocklin:
I wonder how that relates to our user survey

martindurant:
Of course, s3fs is not necessarily used with dask; but also, s3fs is included in the Anaconda installer, so we don't know how many people never update. On top of that, some downloads are caused by mirrors, and sometimes installs come from mirrors; plus the environment behaviours of conda and venv users is likely quite different. In short, I'm not sure the comparison above is useful. 

I would not be against testing via pip *in addition* to conda, except if it makes the CI scripts too complex. Conda is the work environment for all of the devs I think.

TomAugspurger:
Fixed in https://github.com/dask/s3fs/pull/264.

Sorry for the lost effort @Jamim! We'll have a release out in a day or two.

PR #254, opened 2019-11-01
Propagate req_kw from the filesystem to opened files.

sretlawd:
This allows files opened from a filesystem created with
request_payer=True to also be accessed with the appropriate
RequestPayer headers.

Currently, the requester_pays kwarg to S3FileSystem only works for listing keys, etc., but has no effect when you actually open a file to read data. The following will thus result in an access denied error:

```
fs = s3fs.S3FileSystem(requester_pays=True)
with fs.open("bucket/key") as f:
   print(f.read())
```

There doesn't seem to be any way to specify that the file object should also be requester_pays, and I think it makes sense that this should be the default.

In the meantime, for anyone else running into this issue, here's a monkey patch that I'm currently using which accomplishes the same thing.  Probably also illustrates why it'd be much easier to have this implemented upstream!

```
def _patch_fs_request_payer(fs: s3fs.S3FileSystem):
    """
    HACK: Patches an S3FileSystem instance to use RequestPayer: requester on opened files, not
    just directory listings, if the filesystem is initialized with requester_pays=True.
    """

    import types
    from s3fs.core import _fetch_range

    def fetch_range_with_req_kw(self, start, end):
        return _fetch_range(self.fs.s3, self.bucket, self.key, self.version_id, start, end, req_kw=self.fs.req_kw)

    def patched_open(self, *args, **kwargs) -> s3fs.S3File:
        f = original_open(*args, **kwargs)
        f._fetch_range = types.MethodType(fetch_range_with_req_kw, f)
        if hasattr(f, "cache"):
            f.cache.fetcher = f._fetch_range
        return f

    original_open = fs._open
    fs._open = types.MethodType(patched_open, fs)
```

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -256,7 +256,7 @@ class S3FileSystem(AbstractFileSystem):
                 'token': cred['SessionToken'], 'anon': False}
 
     def _open(self, path, mode='rb', block_size=None, acl='', version_id=None,
-              fill_cache=None, cache_type=None, autocommit=True, **kwargs):
+              fill_cache=None, cache_type=None, autocommit=True, requester_pays=None, **kwargs):
         """ Open a file for reading or writing
 
         Parameters
@@ -285,6 +285,9 @@ class S3FileSystem(AbstractFileSystem):
         cache_type : str
             See fsspec's documentation for available cache_type values. Set to "none"
             if no caching is desired. If None, defaults to ``self.default_cache_type``.
+        requester_pays : bool (optional)
+            If RequesterPays buckets are supported.  If None, defaults to the
+            value used when creating the S3FileSystem (which defaults to False.)
         kwargs: dict-like
             Additional parameters used for s3 methods.  Typically used for
             ServerSideEncryption.
@@ -293,6 +296,8 @@ class S3FileSystem(AbstractFileSystem):
             block_size = self.default_block_size
         if fill_cache is None:
             fill_cache = self.default_fill_cache
+        if requester_pays is None:
+            requester_pays = bool(self.req_kw)
 
         acl = acl or self.s3_additional_kwargs.get('ACL', '')
         kw = self.s3_additional_kwargs.copy()
@@ -307,7 +312,7 @@ class S3FileSystem(AbstractFileSystem):
         return S3File(self, path, mode, block_size=block_size, acl=acl,
                       version_id=version_id, fill_cache=fill_cache,
                       s3_additional_kwargs=kw, cache_type=cache_type,
-                      autocommit=autocommit)
+                      autocommit=autocommit, requester_pays=requester_pays)
 
     def _lsdir(self, path, refresh=False, max_items=None):
         if path.startswith('s3://'):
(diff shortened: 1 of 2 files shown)


martindurant:
Yes, this is fair enough. Do you think it would make sense to have the argument instead in `S3File.__init__`, passed in `_open`, so that it could be overridden for individual files?

sretlawd:
Sure, more flexible is definitely better - I'll update the PR.

sretlawd:
OK, this moves it up to S3File.__init__() and initializes in S3FileSystem._open().

I'm struggling with the documentation - I'm not really sure how to explain the difference between s3_additional_kwargs and req_kw - as is, it's pretty confusing.  Any ideas?

martindurant:
Would it be clearer to rename to open_kw everywhere (this fits the convention of the other kwarg-like dicts, I think). @mariusvniekerk might have an opinion here.

s3fs/core.py
```diff
@@ -272,7 +272,7 @@ def get_delegated_s3pars(self, exp=3600):
                 'token': cred['SessionToken'], 'anon': False}
 
     def _open(self, path, mode='rb', block_size=None, acl='', version_id=None,
-              fill_cache=None, cache_type=None, autocommit=True, **kwargs):
+              fill_cache=None, cache_type=None, autocommit=True, req_kw=None, **kwargs):
```
TomAugspurger: Can we call this `request_options`? I had trouble understanding `req_kw` at a glance.

I'd also prefer that this is added to fsspec as an official part of the API.

TomAugspurger:
> Would it be clearer to rename to open_kw everywhere

Don't have a strong preference on `open_` vs `request_`, though I would prefer `_options` to `_kw`. Maybe `fetch_options` is best? For something like `s3fs` request is appropriate. But for a local filesystem, `open` is more appropriate. `fetch` kinda works for both?

s3fs/core.py
martindurant: It specifically stands for "requestor-pays" because, although it is a general dict of kwargs, that is the only option it is used for (or maybe the only one possible). Is it different from the other kwargs because it is only needed when getting actual data, not for file listings, metadata, etc.

Having a set of arbitrary things things in the file system layer that get used upon `open` is not a bad idea, but will be tricky to arrange.

sretlawd:
> Can you add a test for this, ensuring that things are passed through appropriately from an `open`?

There isn't anything for requester_pays=True yet, so wasn't sure where to start but I'll see if I can figure it out.

s3fs/core.py
sretlawd: Just to throw an idea out there - what about having the public API of S3File match that of S3FileSystem, and just have a requester_pays argument on the constructor, keeping req_kw internal?

sretlawd:
So just to show what it'd look like, here's an update with the requester_pays idea.  Could of course change to open_options or whatever you guys decide.

TomAugspurger:
If `requester_pays` is the only request keyword users will likely need, then let's just go with that. Sound OK @martindurant?

martindurant:
ok

TomAugspurger:
Alrighty, thanks @sretlawd!

PR #257, opened 2019-11-06
Don't call super init until init is done

martindurant:
@TomAugspurger , I wonder if you think this might finally fix the auth problems. Now, the super init is only called after connect. This is important, because the super init sets the cached flag, which then skips the rest of the init for a new invokation - this means that it might have been possible for one invocation of the same instance to be done with init while another call of the same thing is still undergoing connect().

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -161,7 +161,6 @@ class S3FileSystem(AbstractFileSystem):
 
         if self._cached:
             return
-        super().__init__()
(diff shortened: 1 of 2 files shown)


TomAugspurger:
Mmm, that seems reasonable. Would be more confident if you had a test case proving it :)

That's tricky of course. Have you tried making a bunch of connection attempts using a thread pool?

Issue #258, opened 2019-11-07
isdir can be very slow when the parent directory contains many other directories


nateyoder:
From my testing it appears as though current calls to `isdir` end up calling `info` which then ends up calling `self.ls(self._parent(path), detail=True, **kwargs)`.  This can end up being an unexpectedly long call if the parent directory contains many subdirectories.

TomAugspurger:
Closed by https://github.com/dask/s3fs/pull/259

PR #259, opened 2019-11-07
Override isdir to not include parent directory ls #258

nateyoder:
This can help avoid potentially unexpected long calls when the parent directory contains many subdirectories which must all be listed. This also uses self._ls under the hood so the dircache still gets filled and used.

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -497,14 +497,29 @@ class S3FileSystem(AbstractFileSystem):
         return super().info(path)
 
     def isdir(self, path):
-        path = self._strip_protocol(path).rstrip("/")
-        bucket, prefix = split_path(path)
-        if prefix:
-            # This should only return files or directories within this path
-            return 0 < len(self._ls(path))
-        else:
-            # Send buckets to super
+        path = self._strip_protocol(path).strip("/")
+        # Send buckets to super
+        if "/" not in path:
             return super(S3FileSystem, self).isdir(path)
+
+        if path in self.dircache:
+            for fp in self.dircache[path]:
+                # For files the dircache can contain itself.
+                # If it contains anything other than itself it is a directory.
+                if fp["name"] != path:
+                    return True
+            return False
+
+        parent = self._parent(path)
+        if parent in self.dircache:
+            for f in self.dircache[parent]:
+                if f["name"] == path:
+                    # If we find ourselves return whether we are a directory
+                    return f["type"] == "directory"
+            return False
+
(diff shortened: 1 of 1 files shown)


martindurant:
I think the best behaviour would be to see if the parent already is in the `fs.dircache`, in which case use it, and do the direct list otherwise. That logic can be in an override to `info()` rather than specifically to `isdir()`. Indeed, fsspec-derived instances always have `.dircache`, even if the specific implementation doesn't use it, so it would make sense to move this logic to `fsspec.spec.AbstractFileSystem`, as well as migrate the implementation of `invalidate_cache`, which happens to be the same in s3fs and gcsfs.

nateyoder:
I added the cache check but after looking into the info override I ran into some performance issues using PyArrow with S3Fs (at least the way I did it) where running info on files in a directory ended up using n-files boto calls rather than just one that populated `.dircache` with the entire directory.  While I understand that is a similar case to what I ran into with `isdir`, at least in my application it slowed things down significantly and so I stopped investigating.

s3fs/core.py
```diff
@@ -497,15 +497,30 @@ def info(self, path, version_id=None):
...
+        parent = self._parent(path)
+        if parent in self.dircache:
+            for f in self.dircache[parent]:
+                if f["name"] == path:
+                    # If we find ourselves return whether we are a directory
+                    return f["type"] == "directory"
+            return False
+
+        # This only returns things within the path and NOT the path object itself
+        return 0 < len(self._lsdir(path))
```
martindurant: `bool(self._lsdir(path))` ?

martindurant:
I think this looks right to me (except the small comment).
I suspect it will be factored out, given that we have isfile and info doing similar things, but I'd be ok to merge for now. Can you confirm, @nateyoder , that this speeds things up for the original case, but is now also fast when the parent exists in the cache?

nateyoder:
@martindurant totally makes sense and I figured that would be the case. However, given my lack of familiarity with the codebase and all of it's use cases wasn't sure I was the person best suited to do it given how many optimizations I'm sure you all have tried to do. I tested that the cache is used at least for my cases and that we aren't doing any of the large ls calls of the parent directory any more.

martindurant:
OK, I am +1 on this then, unless you are planning the refactor, @TomAugspurger

martindurant:
Thanks for the work, @nateyoder !

nateyoder:
Thank you all!

Issue #262, opened 2019-11-12
Stop setting boto logging levels?


TomAugspurger:
Generally speaking, libraries shouldn't configure logging. That's up to the application.

Right now, we set the level for the `boto3` and `botocore` loggers.

> It is strongly advised that you do not add any handlers other than NullHandler to your library’s loggers. This is because the configuration of handlers is the prerogative of the application developer who uses your library

from https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library

PR #263, opened 2019-11-12
Cleanup

TomAugspurger:
* Require fsspec>0.5.2 (the next release)
* Stop configuring boto logging
* Remove unused code

Closes https://github.com/dask/s3fs/issues/262

@martindurant is `utils.py` considered public?


martindurant:
> is utils.py considered public

I suppose not any more, with read_block etc having migrated

PR #269, opened 2019-11-26
Add more s3 fs schemes (s3a://, s3n://)

fhoering:
s3a:// is the official schema used by Hadoop (s3 and s3n are deprecated in hadoop-aws jar)
https://cwiki.apache.org/confluence/display/HADOOP2/AmazonS3

Motivation: Be able to keep a consistent fs scheme across different PySpark, s3fs/PyArrow calls

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -31,6 +31,14 @@ except ImportError:
 
 _VALID_FILE_MODES = {'r', 'w', 'a', 'rb', 'wb', 'ab'}
 
+S3_SCHEMES = ['s3://', 's3a://', 's3n://']
+
+def strip_scheme(path):
+    for scheme in S3_SCHEMES:
+        if path.startswith(scheme):
+            path = path[len(scheme):]
+    return path.rstrip('/').lstrip('/')
+
 
 def split_path(path):
     """
@@ -46,9 +54,7 @@ def split_path(path):
     >>> split_path("s3://mybucket/path/to/file")
     ['mybucket', 'path/to/file']
     """
-    if path.startswith('s3://'):
-        path = path[5:]
-    path = path.rstrip('/').lstrip('/')
+    path = strip_scheme(path)
     if '/' not in path:
         return path, ""
     else:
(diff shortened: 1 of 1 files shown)


martindurant:
I am not opposed to this idea, but it should be pointed out that the protocols have specific meanings for the java runtime, so I'm not sure that they should be treated as strictly identical

fhoering:
To be honest I'm not a big expert either but there seems to be some inconsistency somewhere.  

Seems like Amazon always uses s3:// now
https://docs.aws.amazon.com/emr/latest/ManagementGuide/emr-plan-file-systems.html

And Hadoop always uses s3a://. Not sure why 

Are those different meaning still true in 20198 ? 
I have the impression that those meanings are mostly obsolete now ( (hdfs & s3). Both seem to behave the same way - block storage as a backend and access via a filesystem api. 

Obviously I could convert those paths on my side all the time but it would be nice to be able to use Spark and Pyarrow with hdfs/s3fs filesystem in a consistent way (whic hcoudl also mean s3 all the time but implies fixing hadoop s3 connectors)

s3fs/core.py
```diff
@@ -31,6 +31,14 @@
 
 _VALID_FILE_MODES = {'r', 'w', 'a', 'rb', 'wb', 'ab'}
 
+S3_SCHEMES = ['s3://', 's3a://', 's3n://']
```
martindurant: You can instead set the S3FileSystem protocol attribute and use the `._strip_protocol()` class method. This nicely removes duplicate code which you have found in this PR.

(@TomAugspurger , did the change to a `.protocols` property go in?)

fhoering: OK.
I Get the idea. Change .protocol to a list. Indeed it would be better. It would also permit to support `viewfs` for example which is currently missing.
Currently this is not merged in https://github.com/intake/filesystem_spec  and there is no PR.
Are you currently working on that ? Otherwise I might do it. Seems pretty simple.

fhoering: One question about the code factorization.

What about `split_path` function ? The best would be to move this inside S3FileSystem and also use ._strip_protocol but this would break backward compatibility.

martindurant: I thought @TomAugspurger has written this code somewhere already; I am not working on it.

`split_path` is a little different, since it cares about bucket / key, but principled no opinion on whether it is  a function or method of S3FileSystem. It would make sense for it to make use of `._strip_protocol`, though.

fhoering: Thanks for the feedback.
Turns out setting a list is actually working on `.protocol` (I was looking for the  `.protocols` property before). I pushed the review again. Also removed s3n as I'm not using this and it is deprecated.

martindurant:
lgtm

fhoering:
Thanks

Issue #276, opened 2020-01-05
s3fs does not work with aiobotocore in dask-cloudprovider


gvelchuru:
Hi,

I am trying to run the following code:
```
from dask.distributed import Client
from dask_cloudprovider import FargateCluster
from dask import dataframe as df

if __name__ == "__main__":
cluster = FargateCluster()
cluster.adapt()
client = Client(cluster)

HOLDINGS = df.read_parquet(HOLDINGS_URL)
```
HOLDINGS_URL is my personal s3 bucket, and this code works without using the FargateCluster.

When i do use it, I get the following error:
```
❯ python main.py
/home/ubuntu/miniconda3/envs/dev/lib/python3.7/contextlib.py:119: UserWarning: Creating your cluster is taking a surprisingly long time. This is likely due to pending resources on AWS. Hang tight!
  next(self.gen)
Traceback (most recent call last):
  File "main.py", line 81, in <module>
    HOLDINGS = df.read_parquet(HOLDINGS_URL)
  File "/home/ubuntu/miniconda3/envs/dev/lib/python3.7/site-packages/dask/dataframe/io/parquet/core.py", line 224, in read_parquet
    **kwargs
  File "/home/ubuntu/miniconda3/envs/dev/lib/python3.7/site-packages/dask/dataframe/io/parquet/fastparquet.py", line 178, in read_metadata
    fs, paths, gather_statistics, **kwargs
  File "/home/ubuntu/miniconda3/envs/dev/lib/python3.7/site-packages/dask/dataframe/io/parquet/fastparquet.py", line 119, in _determine_pf_parts
    elif fs.isdir(paths[0]):
  File "/home/ubuntu/miniconda3/envs/dev/lib/python3.7/site-packages/s3fs/core.py", line 511, in isdir
    return bool(self._lsdir(path))
  File "/home/ubuntu/miniconda3/envs/dev/lib/python3.7/site-packages/s3fs/core.py", line 334, in _lsdir
    for i in it:
  File "/home/ubuntu/miniconda3/envs/dev/lib/python3.7/site-packages/aiobotocore/paginate.py", line 14, in __iter__
    "{self} is an AsyncIterable: use `async for`".format(self=self)
NotImplementedError: <aiobotocore.paginate.AioPageIterator object at 0x7fd87c7b3610> is an AsyncIterable: use `async for`
distributed.client - ERROR - Failed to reconnect to scheduler after 10.00 seconds, closing client
distributed.utils - ERROR -
[...]

PR #277, opened 2020-01-10
Switch to botocore

jacobtomlinson:
Given the discussion in #276 I wanted to do a quick POC to see what it would be like to switch s3fs to `botocore` instead of `boto3`.

Most of the changes here are just switching library names and slightly tweaking the way the sessions are constructed as there are some minor differences (keys are given to the client, not the session).

All calls to S3 have remained exactly the same as `boto3` is just handing off to `botocore` underneath with one notable exception.

The `client('s3').copy()` method is not available in `botocore` and after having a dig through the `boto3` codebase it seems quite involved to implement as it enables multi-threaded multi-part transfers between buckets. However I did notice there was an unused `copy_basic` method already implemented which uses the API implementation of `copy_object` and so switched to use that instead.

The result of this is that bucket-bucket transfers may be slower.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -3,12 +3,12 @@ s3fs
 
 |Build Status| |Doc Status|
 
-S3FS builds on boto3_ to provide a convenient Python filesystem interface for S3.
+S3FS builds on botocore_ to provide a convenient Python filesystem interface for S3.
 
 View the documentation_ for s3fs.
 
 .. _documentation: http://s3fs.readthedocs.io/en/latest/
-.. _boto3: https://boto3.readthedocs.io/en/latest/
+.. _botocore: https://botocore.readthedocs.io/en/latest/
 
 .. |Build Status| image:: https://travis-ci.org/dask/s3fs.svg?branch=master
     :target: https://travis-ci.org/dask/s3fs
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,4 +1,4 @@
 numpydoc
-boto3
+botocore
 sphinx
 sphinx-rtd-theme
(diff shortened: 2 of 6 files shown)


martindurant:
Shorter than I would have thought!

@mariusvniekerk , your use cases always seem broader than most, want to review and test this?

TomAugspurger:
Indeed, this isn't too bad. @jacobtomlinson do you know if there will be any user facing changes in `S3FileSystem.__init__`, w.r.t. `client_kwargs` and `kwargs`? Or does botocore have the same behavior of boto with the keywords a user might pass through?

Should we go forward with this?

jacobtomlinson:
AFAIK things should be pretty much the same.

Boto3 is just a thin wrapper around botocore to try and take some of the sting out of the fact that it is swagger generated with a few convenience methods, so the API is pretty much identical. 

As I said above the only downside I see with this is the lacy of the multithreaded copy method. If some folks are up for testing this patch and ensuring it isn't causing too many performance issues then I would propose we go for it. But if performance is impacted then maybe we should go back to the drawing board.

jacobtomlinson:
https://github.com/dask/s3fs/issues/276#issuecomment-574377042 seems to put a little weight behind moving forwards with this.

martindurant:
Does this change mean that s3fs could be run async, then, or is that unrelated (i.e., we're just bipassing the monkey-patch)?

jacobtomlinson:
Unrelated I'm afraid. Just bypassing th monkey-patch.

martindurant:
OK, got it. You can probably drop the "POC", but I would like to get more opinions before moving ahead with this.

martindurant:
OK, swinging back to this: were there any reasons NOT to do this? I am thinking to merge this. Will need to remember to update the conda forge recipe at release time (although, requiring boto3 does bring in botocore automatically, so no one would notice).

Issue #275, opened 2020-01-04
info() return does not include `type` when `version_aware=True`


yzq89:
I believe this results in incorrect behaviour when `version_aware` is enabled (though I personally have not used version-aware buckets myself). Also causes `isfile()` to return false when called on a file.
Version: 0.4.0
https://github.com/dask/s3fs/blob/85b863170f76063b270671442448b15e821f92b6/s3fs/core.py#L467-L476

Example:
```
s3 = s3fs.S3FileSystem(version_aware=True)
path_file = '<bucket>/<prefix>/test.txt'
path_dir = '<bucket>/<prefix>/'
for p in [path_file, path_dir]:
    print(f'isfile={s3.isfile(p)}, isdir={s3.isdir(p)}')
    print(s3.info(p))
```

Output:
```
isfile=False, isdir=False
{'ETag': '"..."', 'Key': '.../test.txt', 'LastModified': datetime.datetime(2019, 10, 25, 20, 29, 56, tzinfo=tzutc()), 'Size': 5, 'size': 5, 'path': '.../test.txt', 'StorageClass': 'STANDARD', 'VersionId': None}
isfile=False, isdir=True
{'Key': '...', 'Size': 0, 'StorageClass': 'DIRECTORY', 'type': 'directory', 'size': 0, 'name': '...'}
```

TomAugspurger:
Thanks for the report. Not including `type` looks like a bug. Interested in submitting a PR to fix it?

yzq89:
Hey @TomAugspurger, certainly happy to do so. I'm not very familiar with the s3fs/fsspec internals though - does this code block apply only for files?
https://github.com/dask/s3fs/blob/85b863170f76063b270671442448b15e821f92b6/s3fs/core.py#L467-L476

TomAugspurger:
I believe so, since s3 doesn't have actual directories. So if we've gotten an object at that point we know it's a file? I could be wrong though.

martindurant:
Yes, that thing must be a file ( cc @mariusvniekerk )

PR #278, opened 2020-01-14
Fix missing type in info() for version-aware buckets

yzq89:
Fixes https://github.com/dask/s3fs/issues/275

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -464,6 +464,7 @@ class S3FileSystem(AbstractFileSystem):
                     'Size': out['ContentLength'],
                     'size': out['ContentLength'],
                     'path': '/'.join([bucket, key]),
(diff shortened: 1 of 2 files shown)


TomAugspurger:
Thanks. Are you able to add a test for this? It may not be possible, if moto doesn't support creating versioned files.

yzq89:
I'm fairly new to tests, hopefully I got it in the right place? pytest showed the expected results.

TomAugspurger:
Thanks, that looks good.

FYI, I edited the text of your GitHub comment to say "Fixes #275" rather than addresses, so that the issue gets auto-closed.

martindurant:
Thank you @yzq89 !

yzq89:
Awesome! Glad to help

PR #283, opened 2020-02-10
Only configure the logger if S3FS_LOGGING_LEVEL is set

### LINKED ISSUE #282: Stop setting logger handlers and levels
This came up re: boto3 and botocore loggers in #262 and was resolved for them, but the `s3fs.core` module still configures a handler and sets the level of the `s3fs` logger:

https://github.com/dask/s3fs/blob/ab6ebbeeaa0cf97d5c71d8556ca6b29082560945/s3fs/core.py#L16-L23

Logging config is something that should generally be owned solely by the application and not interfered with by libraries. The [Python documentation](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library) describes this principle as:

> It is strongly advised that you do not add any handlers other than NullHandler to your library’s loggers. This is because the configuration of handlers is the prerogative of the application developer who uses your library.

Currently, my application's normal logging strategy is interfered with when I use fsspec to handle s3:// URLs via s3fs, and there's no good way to work around it since logging config is loaded very early in the application (as is generally advisable) but s3fs is loaded much later by fsspec on demand.

I think the best solution would be to remove the handler-adding code entirely, but realize that may not abide by backwards compatibility policies for existing behaviour.

As an alternative, adding the handler could be moved into the conditional which sets the level, essentially making it opt-in by setting `S3FS_LOGGING_LEVEL` in the environment.

A final alternative might be to consider calling `logging.basicConfig()` instead, which will do nothing if a root logger is already configured, and otherwise sets up a basic handler like `s3fs.core` is already doing. I think this would be a reasonable compromise, but I'm not entirely certain it wouldn't have other unintended side-effects, such as load-order-dependent quirks.

I'd be happy to provide the patch for any of these options, if it would be merged and released.

tsibley:
Allows people to opt-in to the basic logger config which outputs s3fs
messages while by default avoiding interfering with application-level
logging config.

The premise for piggy-backing on S3FS_LOGGING_LEVEL is that anyone
setting it expects to see some log output.  Anyone who doesn't want to
see log output can either not set it themselves or ensure it is unset
before s3fs is loaded.

Resolves #282.

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -81,9 +81,14 @@ The following are known current omissions:
 Logging
 -------
 
-The logger ``s3fs.core.logger`` provides information about the operations of the
-file system. To see messages, set its level to ``DEBUG``. You can also achieve this via
-an environment variable ``S3FS_LOGGING_LEVEL=DEBUG``.
+The logger named ``s3fs`` provides information about the operations of the file
+system.  To quickly see all messages, you can set the environment variable
+``S3FS_LOGGING_LEVEL=DEBUG``.  The presence of this environment variable will
+install a handler for the logger that prints messages to stderr and set the log
+level to the given value.  More advance logging configuration is possible using
+Python's standard `logging framework`_.
+
+.. _logging framework: https://docs.python.org/3/library/logging.html
 
 Credentials
 -----------
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -14,12 +14,13 @@ from s3fs.errors import translate_boto_error
 from s3fs.utils import ParamKwargsHelper
 
 logger = logging.getLogger('s3fs')
-handle = logging.StreamHandler()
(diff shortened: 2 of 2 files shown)


TomAugspurger:
Thanks, this looks nice. I think the "Logging" section in `index.rst` could be updated to explain how things work now.

martindurant:
Agreed, looks good, but needs documenting.

tsibley:
Ah, sorry, totally spaced on docs. Re-pushed with a doc update included.

martindurant:
Thank you.

tsibley:
Any chance of a new release soon? I'd love to be able to stop hacking around this logging issue. :-)

martindurant:
Yes, coming once I've finalised some things in fsspec

tsibley:
Thank you!

PR #288, opened 2020-02-21
Multipart copy for big files

### LINKED ISSUE #287: Regression caused in move to botocore by removal of copy_managed
This was added in #117 to deal with copying files larger than 5GB within S3.

This is now no longer working since we have removed the copy_managed

martindurant:
Fixes #287 

@mariusvniekerk , I don't think we can ask moto to test this for us - would you?

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -790,7 +790,6 @@ class S3FileSystem(AbstractFileSystem):
                     try:
                         self.fs._call_s3(
                             self.s3.upload_part_copy,
-                            self.s3_additional_kwargs,
                             Bucket=bucket,
                             Key=key,
                             PartNumber=i + 1,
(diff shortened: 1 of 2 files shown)


mariusvniekerk:
Yeah moto is going to want to allocate it in memory. Doubt that's super feasible.

martindurant:
This should be the rough shape, and I suppose we can test by copying smaller files with smaller, chunks, so long as each chunk is >5MB?

mariusvniekerk:
Yeah the constraint is that you can't copy with a range going from 0 to <5MB

mariusvniekerk:
We can allow specifying the chunk size as a variable and just default it to something biggish

mariusvniekerk:
Alternatively we could use https://github.com/boto/s3transfer directly.  This doesn't seem to need boto3 at all.

martindurant:
@mariusvniekerk , seems like a bad idea to rely on a thread pool for something commonly used in conjunction with Dask. I was thinking, conversely, that it would be nice to provide Dask as an up/download manager for the various file systems - if their implementations are reasonably thread-safe!

mariusvniekerk:
I doubt that most dask operations copy objects within / between buckets :)

Issue #198, opened 2019-07-16
TypeError: multiple values for keyword argument 'read_timeout'


keurcien:
I wanted to increase the value of `read_timeout` parameter (passed to botocore.config.Config) but it seems like it shouldn't be passed from the config_kwargs:

```python
# from s3fs/core.py in connect(self, refresh)
conf = Config(connect_timeout=self.connect_timeout,
    read_timeout=self.read_timeout,
    **self.config_kwargs)
```
Passing `read_timeout` would raise the following error:
```python
TypeError: type object got multiple values for keyword argument 'read_timeout'
```

Here's my MWE:
```python
import s3fs
s3 = s3fs.S3FileSystem(config_kwargs={'read_timeout': 500})
```

As a hacky workaround, I can set another parameter to some value, e.g:
```python
s3 = s3fs.S3FileSystem(config_kwargs={'max_pool_connections': 50})
```

And then I can set the value of `read_timeout` without raising any error:
```python
s3 = s3fs.S3FileSystem(config_kwargs={'read_timeout': 500})
print(s3.config_kwargs)
```

martindurant:
For some reason I can't remember, `read_timeout` is implemented as a class attribute, so you would have to do
```
s3fs.S3FileSystem.read_timeout = 500
s3 = s3fs.S3FileSystem(..)
```

keurcien:
Oh ok, I see, thanks for the very quick reply!

martindurant:
Works, OK to close?

PR #291, opened 2020-02-26
Make S3FileSystem read_timeout and connection_timeout configurable

Bruschkov:
# Description
When initiliazing an S3FileSystem object, it is possible to configure the botocore session as per the parameter "config_kwargs". However, if one tries to configure the config parameters "connect_timeout" or "read_timeout" per config_kwargs, the S3FileSystem can not be created. This is unexpected behaviour in so far as the the config_kwargs parameter suggests that every field of the session config can be configured.

# Analysis
The reason for this is that these two values are hardwired into the boto config object:
```
conf = Config(connect_timeout=self.connect_timeout,
                          read_timeout=self.read_timeout,
                          **self.config_kwargs)
```
This will lead to the following error, if "connect_timeout" or "read_timeout" are part of the config_kwargs: 
```
TypeError: type object got multiple values for keyword argument 'connect_timeout'
```

# Solution
We needed to configure the "connect_timeout" and "read_timeout" and found ourselves unable to do so. There is a suggested workaround which is hack-ish and inconvenient: https://github.com/dask/s3fs/issues/198
We implemented a solution that uses the values from config_kwargs if they are set and defaults to the class attributes if not.

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -189,6 +189,14 @@ class S3FileSystem(AbstractFileSystem):
         else:
             return path.split('/', 1)
 
+    def _prepare_config_kwargs(self):
+        config_kwargs = self.config_kwargs.copy()
+        if "connect_timeout" not in config_kwargs.keys():
+            config_kwargs['connect_timeout'] = self.connect_timeout
+        if "read_timeout" not in config_kwargs.keys():
+            config_kwargs['read_timeout'] = self.read_timeout
+        return config_kwargs
+
     def connect(self, refresh=True):
         """
         Establish S3 connection object.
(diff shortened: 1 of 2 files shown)


Bruschkov:
One thing to discuss in the light of this issue: Is it necessary for "connect_timeout" and "read_timeout" to be class attributes of S3FileSystem?

From only the S3FileSystem code it doesn't seem that way...

If they don't have to be class attributes, this could be solved a lot more elegantly and flexible that it is right now.

martindurant:
Sorry to be slow to respond here.
They two parameters do *not* need to be class attributes, but do need to have sensible defaults. Those defaults could be in the init signature, but of course there are already rather a lot of parameters available there...

PR #295, opened 2020-03-10
Implement checksum for S3FileSystem

eeroel:
- Implement `S3FileSystem.checksum` that exposes the `refresh` parameter
- For files, use the ETag field for this purpose; for directories, use a hash of the output of `ls(path, detail=True)`, similar to the default implementation in `AbstractFileSystem`; if the path contains multiple files/directories, use a hash of the list of checksums of the individual items

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -500,14 +500,19 @@ class S3FileSystem(AbstractFileSystem):
 
         listing = self.ls(path, detail=True, refresh=refresh)
         
+        def get_checksum(item):
+            if item["type"] != 'directory':
+                return item["ETag"].strip('"')
+            else:
+                return tokenize(item)
+             
         if not listing:
             raise FileNotFoundError(path)
         elif len(listing)==1:
-            return int(listing[0]["ETag"].strip('"'), 16)
+            return int(get_checksum(listing[0]), 16)
         else:
-            return int(tokenize(self.info(path)), 16)
-
-        return checksum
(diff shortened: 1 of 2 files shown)


s3fs/core.py
```diff
@@ -478,6 +480,39 @@ def info(self, path, version_id=None):
...
+        Parameters
+        ----------
+        path : string/bytes
+            file or path to get checksum for
+        refresh : bool (=False)
+            if False, look in local cache for file details first
+        
+        """
+
+        listing = self.ls(path, detail=True, refresh=refresh)
```
martindurant: Why not `info`? I would have thought that this method works for only a single file, and is probably not useful for a directory (because a directory's contents can have changed without the directory itself having changed, right?)

s3fs/core.py
```diff
@@ -478,6 +480,39 @@ def info(self, path, version_id=None):
...
+        
+        def get_checksum(item):
```
martindurant: I wouldn't make a dynamic function here; it could be a static one.
This is especially true if the method only works on one file at a time.

s3fs/core.py
eeroel: Right. I thought it would be tricky to modify `info` to support bypassing the cache, but on a second look it's probably not. I'll try to fix this.

eeroel: OK, changed this to use `info` instead, and removed test case for directories (since that is unspecified behavior anyway).

martindurant:
OK, this looks good, thank you.

PR #302, opened 2020-03-27
Require python>=3.6

TomAugspurger:
xref https://github.com/dask/s3fs/issues/301

PR #307, opened 2020-03-31
Remove f-strings for py35 compat

martindurant:


TomAugspurger:
Ah, making a 3.5-compatible release is even better! You'll need to revert https://github.com/dask/s3fs/pull/302 as well.

martindurant:
Except that I can't make it work :|
I'll have to check the versions of moto/botocore being installed.

martindurant:
OK works, I'll wrap it up - fails on versions (will skip) and sorting of glob (will fix)

Issue #120, opened 2018-03-11
Setting a custom endpoint_url


npezolano:
How do I pass a custom endpoint url to `s3fs.S3FileSystem` ?

I've tried:

    kwargs = {'endpoint_url':"https://s3.wasabisys.com",
              'region_name':'us-east-1'}
    self.client = s3fs.S3FileSystem(key=AWS_ACCESS_KEY_ID, 
                                    secret=AWS_SECRET_ACCESS_KEY,
                                    use_ssl=True,
                                    **kwargs)

However I get the error:

      File "s3fs/core.py", line 215, in connect
        **self.kwargs)
    TypeError: __init__() got an unexpected keyword argument 'endpoint_url'

I've also tried passing kwargs as the parameter `config_kwargs` and `s3_additional_kwargs` with similar errors. The problem with these is that `endpoint_url` gets passed to `botocore.Config` when it shouldn't.

I can verify `boto3` is working with the following:

    client = boto3.client("s3",
            aws_access_key_id=AWS_ACCESS_KEY_ID,
            aws_secret_access_key=AWS_SECRET_ACCESS_KEY,
            endpoint_url="https://s3.wasabisys.com",
            use_ssl=True,
            region_name="us-east-1", 
            api_version=None,verify=None, config=None)

`fs-s3fs` also works

martindurant:
Duplicate of https://github.com/dask/s3fs/issues/119

It's `client_kwargs={'endpoint_url': 'https:...'}` that you want.

npezolano:
Thanks for the quick reply! That needs to get added to the documentation

martindurant:
@npezolano , since you've just looked through the documentation, you may be the best placed to consider *where* this information should appear. Would you be willing to provide a PR?

dapperfu:
My $0.02 after finding this googling, somewhere up higher(?).

We are using a self hosted minio setup. Almost all online documentation starts with the assumption that everyone is using Amazon's S3 cloud. I spent multiple hours trying to figure out just the search terms that I needed to find ```endpoint_url```.

A note about using s3fs with a self hosted or non-AWS server in an example would probably have gone a long way in helping.

martindurant:
This would naturally find a place in the S3FileSystem docstring and the RST documentation (docs/source/ directory). Would you be willing to add extra information in a PR? I have personally never used s3fs with anything other than AWS, and I suppose it shows. We *have* had it used by others with other s3-compatible services.

[...]

PR #310, opened 2020-04-02
Add Self-hosted S3 section to the docs

tumido:
Documentation for #120 

I was looking for a solution to the very same problem as is described in the issue #120. It seems that:
1) I'm not alone with this usecase.
2) Nobody really likes to write the documentation.

Let's save the future users some time. :slightly_smiling_face:

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -119,6 +119,26 @@ Furthermore, ``S3FileSystem.current()`` will return the most-recently created
 instance, so this method could be used in preference to the constructor in
 cases where the code must be agnostic of the credentials/config used.
 
+Self-hosted S3
+--------------
+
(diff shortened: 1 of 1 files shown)


docs/source/index.rst
```diff
@@ -119,6 +119,26 @@ Furthermore, ``S3FileSystem.current()`` will return the most-recently created
 instance, so this method could be used in preference to the constructor in
 cases where the code must be agnostic of the credentials/config used.
 
+Self-hosted S3
+--------------
+
+To use S3Fs against your self hosted S3-compatible storage, like `MinIO`_ or
```
martindurant, line 125: Prefer lowercase `s3fs` (with backticks to get the code style)

tumido: Updated :+1:

PR #315, opened 2020-04-16
Fixing missing boto3 Session.create_client in botocore v. 1.15.39

jesperrix:


jesperrix:
Sorry, my bad. It looks like if you use 
`session = botocore.session.Session(profile='some-profile')
session.create_client  # exists
`

However when using boto3:
`session = boto3.Session(profile='some-profile')
session.create_client  # Does not exist
# but so does
session.client
`

jesperrix:
And i use s3fs by:
```
session = boto3.Session(profile='some-profile')
fs = s3fs.S3FileSystem(session=session)

martindurant:
Yes, sorry, you should now use a botocore session not a boto3 one. They ought to have the same methods, but they do not. This was done to support async via aiobotocore, but there is no async wrapper for boto3.

martindurant:
I think this will become obsolete following the switch to aiobotocore in #336

Issue #321, opened 2020-05-22
S3FileSystem.copy multipart copy error


jonny-wright:
In version 0.4.2, I get this error when copying a 6GB file within the same bucket.

`Write failed: ClientError('An error occurred (InvalidArgument) when calling the UploadPartCopy operation: Range specified is not valid for source object of size: 6403334120')`

If I rollback back to 0.4.0 it works. Look related to the use of multipart copy in later versions.

I have tried different files and different bucket and always get this error.

martindurant:
Please debug the bytes ranges created in [this line](https://github.com/dask/s3fs/blob/master/s3fs/core.py#L916) - perhaps there is an off-by-one error.

jonny-wright:
Yes, it does look like an off-by-one error.

[https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.upload_part_copy](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.upload_part_copy)

CopySourceRange is zero based.

PR #322, opened 2020-05-22
Fix for multipart copy one off error

jonny-wright:
Fix for #321

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -894,6 +894,13 @@ class S3FileSystem(AbstractFileSystem):
             raise ValueError('Copy failed (%r -> %r): %s' % (path1, path2, e))
         self.invalidate_cache(path2)
 
+
+    @staticmethod
+    def _get_brange(size, block):
+        for offset in range(0, size, block):
+            yield offset, min(offset + block - 1, size - 1)
+
+
     def copy_managed(self, path1, path2, block=5 * 2**30, **kwargs):
         """Copy file between locations on S3 as multi-part
 
@@ -909,11 +916,10 @@ class S3FileSystem(AbstractFileSystem):
             self.s3.create_multipart_upload,
             Bucket=bucket, Key=key, **kwargs)
         parts = []
-        for i, offset in enumerate(range(0, size, block)):
+        for i, (brange_first, brange_last) in enumerate(self._get_brange(size, block)):
             for attempt in range(self.retries + 1):
                 try:
-                    brange = "bytes=%i-%i" % (
-                            offset, min(offset + block - 1, size - 1))
+                    brange = "bytes=%i-%i" % (brange_first, brange_last)
                     out = self._call_s3(
                         self.s3.upload_part_copy,
                         Bucket=bucket,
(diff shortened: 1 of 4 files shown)


martindurant:
Would you please mark the py35 tests as allowed failure (this is because boto/moto are no longer being released for py35)?
Can you please make a test for multipart copy: make a file with random data >5MB, copy it, and check that the new file has the same contents as the old one?

jonny-wright:
Where do I mark allowed test failures (I had a quick look in travis)?

The preexisting test `test_copy_managed` already copies a 36MB file in 5MB blocks. Which makes me think this could be a blind spot in moto?

martindurant:
I think you move the relevant env block from include into an allow_failures block
https://docs.travis-ci.com/user/customizing-the-build. (where travis has renamed matrix-> jobs, but that doesn't matter for us).

> The preexisting test test_copy_managed already copies a 36MB file in 5MB blocks. Which makes me think this could be a blind spot in moto?

That's a little bit worrying! Can you think of any way to test that the copy is correct? Perhaps it would be better to use random data rather than repetitive data?

s3fs/core.py
```diff
@@ -894,6 +894,13 @@ def copy_basic(self, path1, path2, **kwargs):
             raise ValueError('Copy failed (%r -> %r): %s' % (path1, path2, e))
         self.invalidate_cache(path2)
 
+
+    @staticmethod
+    def _get_brange(size, block):
```
martindurant, line 899: Sorry, what is this?
Since it's a function, I would rather it didn't clutter the spec namespace. It could be in utils, ideally with a leading underscore.

jonny-wright: Agreed, I've moved it. But still need to figure out why `test_copy_managed` passes before and after my change to `copy_managed`.

martindurant: Have you tried with totally random data?

jonny-wright: In `test_copy_managed` I just replaced:

```data = b'abc' * 12*2**20```
with
```data = os.urandom(3*12*2**20)```

And, as of yet, I still can't get the test to fail (on master).

For reference here are the two byte ranges (the only difference being the very last digit).

Master branch 
```
37748736
bytes=0-5242879
bytes=5242880-10485759
bytes=10485760-15728639
bytes=15728640-20971519
bytes=20971520-26214399
bytes=26214400-31457279
bytes=31457280-36700159
bytes=36700160-37748736
```

This branch
```
37748736
bytes=0-5242879
bytes=5242880-10485759
bytes=10485760-15728639
bytes=15728640-20971519
bytes=20971520-26214399
bytes=26214400-31457279
bytes=31457280-36700159
bytes=36700160-37748735
```

martindurant: and if you run this against real S3, the master fails but this branch passes?

jonny-wright: That's right.  The master gives:
 
```OSError: Write failed: ClientError('An error occurred (InvalidArgument) when calling the UploadPartCopy operation: Range specified is not valid for source object of size: 6403334120')```

The branch completes the copy without error (although I have not verified the content of the new file).

martindurant: Also, could you please check the difference when the total data does/doesn't divide exactly by the block size?

Issue #320, opened 2020-05-20
fs.info() can trigger a bad performance if object  is missing and cause unexpected list_object request.


philipzhaoTS:
Reading through the [fs.info](https://github.com/dask/s3fs/blob/master/s3fs/core.py#L555) codebase. I think the following behavior is undesired if a particular file path is missing.

Examples:
fs.info("s3://mybucket/missingDir/missingFile"). 
This will result the following request
head_object (mybucket/missingDir/missingFile) => 404, [line here](https://github.com/dask/s3fs/blob/master/s3fs/core.py#L538)

list_object(mybucket/missingDir, "/" ) => result an empty [], [line here](https://github.com/intake/filesystem_spec/blob/master/fsspec/spec.py#L538)

list_object(mybucket/, "/") => result all the data [line here](https://github.com/dask/s3fs/blob/master/s3fs/core.py#L630)

list_object(mybucke/missingDir/missingFile, "/") => result empty [] [line here](https://github.com/intake/filesystem_spec/blob/master/fsspec/spec.py#L542)

list_object(mybucket/missingDir, "/") => result an empty [] [line here](https://github.com/dask/s3fs/blob/master/s3fs/core.py#L630)

Given the following example:
fs.info("s3://mybucket/mydir/missingFile")
head_object (mybucket/mydir/missingFile) => 404
list_object(mybucket/mydir, "/" ) => result some files
list_object(mybucket/mydir/missingFile, "/") => result empty []
list_object(mybucket/mydir, "/") => result some files

Now image if mybucket or mybucket/mydir has a lot of files. Then this list_object will trigger a pagination call and result a lot of list operation.

I think a better implementation should be like:
head_object(mybucket/my_key) # if return 404 then continue
list_object(mybucket/my_key, "/", max_key=1) # to determine whether it is a directory or now.

In my opinion, we don't need to call its parent directory to fetch all its data. I understand there is caching within the fs that it will utilize it. But I think that caching should be used when user explicitly call ls() rather than calling info().

martindurant:
Good point, `info` should give FileNotFound immediately, if the HEAD call fails. The only counterargument I can think for this, is that it's possible that a user only has list and not get access to the bucket - can we detect that case?

> `list_object(mybucket/, "/")`

Couldn't figure out why this call is made.

philipzhaoTS:
S3 will give your accessDenied exception, you can pass that exception to caller. Or you can translate into one of fileSystem's exception.

[...]

PR #323, opened 2020-05-22
Address info() performance issue

pzhao5:
Address performance issue discuss in https://github.com/dask/s3fs/issues/320.

Update info() method to use list_objects_v2(maxKey=1) to determine whether a particular path is a directory not.
    
It will try to avoid calling super.info(), this will result pretty expensive operations.

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -556,8 +556,9 @@ class S3FileSystem(AbstractFileSystem):
             except ParamValidationError as e:
                 raise ValueError('Failed to head path %r: %s' % (path, e))
 
-            # We could look for a prefix.
             try:
+                # We check to see if the path is a directory by attempting to list its
+                # contexts. If anything is found, it is indeed a directory
                 out = self._call_s3(
                     self.s3.list_objects_v2,
                     kwargs,
@@ -581,7 +582,7 @@ class S3FileSystem(AbstractFileSystem):
             except ClientError as e:
                 raise translate_boto_error(e)
             except ParamValidationError as e:
-                raise ValueError("Failed to head path %r: %s" % (path, e))
+                raise ValueError("Failed to list path %r: %s" % (path, e))
 
         return super().info(path)
     
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -533,7 +533,8 @@ class S3FileSystem(AbstractFileSystem):
                                  "filesystem is not version aware")
         bucket, key, path_version_id = self.split_path(path)
         version_id = _coalesce_version_id(path_version_id, version_id)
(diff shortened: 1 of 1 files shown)


s3fs/core.py
```diff
@@ -551,12 +551,38 @@ def info(self, path, version_id=None, refresh=False):
...
+                        "Size": 0,
+                        "size": 0,
+                        "StorageClass": "DIRECTORY",
+                    }
+
+                raise FileNotFoundError(path)
+            except ClientError as e:
+                raise translate_boto_error(e)
+            except ParamValidationError as e:
+                raise ValueError("Failed to head path %r: %s" % (path, e))
```
martindurant, line 584: This was no longer a HEAD operation, so this message should be updated.

s3fs/core.py
```diff
@@ -551,12 +551,38 @@ def info(self, path, version_id=None, refresh=False):
...
                 # This could have failed since the thing we are looking for is a prefix.
-                if isinstance(ee, FileNotFoundError):
-                    return super(S3FileSystem, self).info(path)
-                else:
+                if not isinstance(ee, FileNotFoundError):
                     raise ee
             except ParamValidationError as e:
                 raise ValueError('Failed to head path %r: %s' % (path, e))
+
+            # We could look for a prefix.
```
martindurant, line 559: Please move the comment within the block and expand it along the lines of: "we check to see if the path is a directory by attempting to list its contexts; if anything is found, it is indeed a directory"

pzhao5: Done

s3fs/core.py
pzhao5: Done. Let me know if it address your concern.

pzhao5:
In term of self._ls_from_cache(path) check. It has been checked by this line:
https://github.com/dask/s3fs/blob/master/s3fs/core.py#L536
And I believe this line should cover the case whether it is a file or dir. Should i check it again?

martindurant:
> Should i check it again

Right; that line gives the condition where we certainly want to (re)check HEAD; but if the listing *is* in the cache, your new version doesn't give a way to retrieve it.

pzhao5:
I believe that super.info(path) will use the cache information to retrieve the information.

martindurant:
Yes you are right, but this only happens now after you try to list to see if the path is a directory.

pzhao5:
I don't follow the comment for changing. Would you mind pesocode how you want to change it based on current version. Thanks.

martindurant:
On second thoughts, I think `should_fetch_from_s3` does exactly what I had in mind - so all looks good!

Issue #245, opened 2019-10-11
`mkdir` silently fail to create parent dir after version 0.2.2


JackWindows:
After version 0.2.2, the logic of `mkdir` changed and it can only be called when creating a top level dir (a bucket, e.g. `mkdir('bucket')`).
Creating anything other than a bucket will lead to silent failure (the command returns without exception but nothing has been done).

Here is a illustration of the issue:
```
$ python3
Python 3.6.8 (default, Oct  7 2019, 12:59:55)
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import s3fs
>>> s3fs.__version__
'0.3.5'
>>> fs = s3fs.S3FileSystem()
>>> # Creating a top level dir is fine
... fs.mkdir('my-test-bucket')
>>> fs.exists('my-random-test-bucket-dkaf')
True
>>> fs.rmdir('my-random-test-bucket-dkaf')
>>> # Creating a 2nd level folder does nothing
... fs.mkdir('my-random-test-bucket-dkaf/abc/')
>>> fs.exists('my-random-test-bucket-dkaf/abc')
False
>>> fs.exists('my-random-test-bucket-dkaf')
False
```
While in 0.2.2 an exception will be thrown if bucket doesn't exist yet
```
$ python3
Python 3.6.8 (default, Oct  7 2019, 12:59:55)
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import s3fs
>>> s3fs.__version__
'0.2.2'
>>> fs = s3fs.S3FileSystem()
>>> fs.mkdir('my-random-test-bucket-dkaf/abc/')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.6/dist-packages/s3fs/core.py", line 865, in mkdir
    self.touch(path, acl=acl, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/s3fs/core.py", line 1044, in touch
    Bucket=bucket, Key=key, ACL=acl)
  File "/usr/local/lib/python3.6/dist-packages/s3fs/core.py", line 195, in _call_s3
    return method(**additional_kwargs)
  File "/home/jackwindows/.local/lib/python3.6/site-packages/botocore/client.py", line 357, in _api_call
    return self._make_api_call(operation_name, kwargs)
[...]

PR #325, opened 2020-06-10
mkdir/rmdir behaviour

martindurant:
@jorisvandenbossche , was it something like this that you had in mind?


jorisvandenbossche:
Not fully sure if I read the diff correctly, but this ensures to at least always create the bucket right? (but not any placeholder for the other "directories", which is what you are advocating for in #245)

It might not necessarily enable writing generic tests using `mkdir` and `rmdir` (would need to check that) like the ones I was testing it with, but in the end the primary concern should be correct / expected behaviour in practice, of course (and for that I don't have enough experience with S3 filesystems to do any judgement). And this seems a useful change anyway.

PR #328, opened 2020-06-12
Fix exception causes in core.py

cool-RR:


cool-RR:
@martindurant This is a continuation of #326

martindurant:
Thanks!

PR #331, opened 2020-06-24
Implement fsspec makedirs

fhoering:
- mkdirs is called by https://github.com/apache/arrow/blob/master/python/pyarrow/filesystem.py#L336

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -421,9 +421,6 @@ class S3FileSystem(AbstractFileSystem):
             return files
         return self.dircache[path]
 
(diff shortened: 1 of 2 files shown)


martindurant:
Note that this does exist, but is a no-op by default. Your implementation I think amounts to creating the bucket if it doesn't exist? 
It should have a test.

martindurant:
def: https://github.com/intake/filesystem_spec/blob/master/fsspec/spec.py#L939

fhoering:
Yes I would like it to create the bucket is not existing. Added a test

fhoering:
Thanks

PR #335, opened 2020-07-18
Fix checksum for multipart uploads

mohtar:
checksum() fails if the object is created using the multipart upload feature of S3. This is because the resulting ETag is in a slightly different format: _hash_-_num_parts_. This PR addresses the problem by parsing the _hash_ part only and discarding the rest.

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -616,7 +616,7 @@ class S3FileSystem(AbstractFileSystem):
         info = self.info(path, refresh=refresh)
 
         if info["type"] != 'directory':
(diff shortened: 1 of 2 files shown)


martindurant:
Would you mind adding a test which shows this in action?

mohtar:
Sorry what else should I test? The test code I added fails without the fix.

martindurant:
oops, sorry, the tab hadn't expanded :|

mohtar:
Thanks! 👍

PR #336, opened 2020-07-20
Async

martindurant:
Bringing async to s3fs, with concurrent bulk operations (cp, rm, get/put, cat/pipe).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
-aiobotocore
+aiobotocore>=1.0.1
 fsspec>=0.7.4
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
-botocore>=1.12.91
-fsspec>=0.6.0
+aiobotocore
+fsspec>=0.7.4
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -205,8 +205,8 @@ class S3FileSystem(AsyncFileSystem):
             try:
                 return await method(**additional_kwargs)
             except S3_RETRYABLE_ERRORS as e:
-                err - e
-                time.sleep(1.7**i * 0.1)
+                err = e
+                await asyncio.sleep(min(1.7**i * 0.1, 15))
             except Exception as e:
                 err = e
                 break
@@ -430,7 +430,6 @@ class S3FileSystem(AsyncFileSystem):
         bucket, key, _ = self.split_path(path)
         if not key or create_parents:
             # catch error if creating existing bucket?
-            # catch error if creating existing bucket?async def
             if acl and acl not in buck_acls:
                 raise ValueError('ACL not in %s', buck_acls)
             try:
@@ -1135,7 +1134,8 @@ class S3FileSystem(AsyncFileSystem):
         if len(pathlist) > 1000:
             raise ValueError("Max number of files to delete in one call is 1000")
         delete_keys = {'Objects': [{'Key': self.split_path(path)[1]} for path
-                                   in pathlist]}
+                                   in pathlist],
+                       "Quiet": True}
         for path in pathlist:
             self.invalidate_cache(self._parent(path))
         await self._call_s3(
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -988,7 +988,7 @@ class S3FileSystem(AsyncFileSystem):
             the number of seconds this signature will be good for.
         """
         bucket, key, version_id = self.split_path(path)
-        return self.s3.generate_presigned_url(
+        return sync(self.loop, self.s3.generate_presigned_url,
             ClientMethod='get_object',
             Params=dict(Bucket=bucket, Key=key, **version_id_kw(version_id), **kwargs),
             ExpiresIn=expires)
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -12,6 +12,7 @@ from fsspec.utils import infer_storage_options, tokenize
 from fsspec.asyn import AsyncFileSystem, sync, sync_wrapper
 
 import aiobotocore
+import botocore
 import aiobotocore.session
 from aiobotocore.config import AioConfig
 from botocore.exceptions import ClientError, ParamValidationError, BotoCoreError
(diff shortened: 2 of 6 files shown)


s3fs/core.py
```diff
@@ -160,45 +161,58 @@ def __init__(self, anon=False, key=None, secret=None, token=None,
...
         logger.debug("CALL: %s - %s - %s" % (method.__name__, akwarglist, kw2))
         additional_kwargs = self._get_s3_method_kwargs(method, *akwarglist,
                                                        **kwargs)
-        return method(**additional_kwargs)
+        for i in range(self.retries):
+            try:
+                return await method(**additional_kwargs)
+            except S3_RETRYABLE_ERRORS as e:
+                err - e
+                time.sleep(1.7**i * 0.1)
```
mariusvniekerk, line 209: should this be an async sleep?

mariusvniekerk: additionally i'd also put something like `min(15, 1.7**i * 0.1)` just to avoid the sleeping time exploding in weird cases

s3fs/core.py
```diff
@@ -421,10 +425,12 @@ def _lsdir(self, path, refresh=False, max_items=None):
             return files
         return self.dircache[path]
 
-    def mkdir(self, path, acl="", create_parents=True, **kwargs):
+    async def _mkdir(self, path, acl="", create_parents=True, **kwargs):
         path = self._strip_protocol(path).rstrip('/')
         bucket, key, _ = self.split_path(path)
         if not key or create_parents:
+            # catch error if creating existing bucket?
+            # catch error if creating existing bucket?async def
```
mariusvniekerk, line 433: cleanup

s3fs/tests/test_s3fs.py
```diff
@@ -49,77 +48,81 @@
...
-                    "Condition": {
-                        "StringNotEquals": {
-                            "s3:x-amz-server-side-encryption": "aws:kms"
-                        }
+    import shlex
+    import subprocess
+
+    proc = subprocess.Popen(shlex.split("moto_server s3 -p %s" % port))
+
+    timeout = 5
```
mariusvniekerk, line 63: Given this timeout, i'd highly recommend using the latest dev version of moto to run this with.

s3fs/core.py
martindurant: Good poin!

s3fs/tests/test_s3fs.py
martindurant: What has changed?

s3fs/core.py
```diff
@@ -644,7 +763,7 @@ def isdir(self, path):
             return False
 
         # This only returns things within the path and NOT the path object itself
-        return bool(self._lsdir(path))
+        return bool(sync(self.loop, self._lsdir, path))
 
     def ls(self, path, detail=False, refresh=False, **kwargs):
```
mrocklin, line 768: I'm curious, so for something like `ls` is there an async equivalent?

s3fs/core.py
```diff
@@ -941,126 +1064,125 @@ def copy_managed(self, path1, path2, block=5 * 2**30, **kwargs):
...
+    async def _clear_multipart_uploads(self, bucket):
+        """Remove any partial uploads in the bucket"""
+        out = await self._call_s3(self.s3.list_multipart_uploads, Bucket=bucket)
+        await asyncio.gather(*[
+            self._call_s3(self.s3.abort_multipart_upload,
+                          Bucket=bucket,
+                          Key=upload['Key'],
+                          UploadId=upload['UploadId'])
+            for upload in out['Contents']
+        ])
```
mrocklin, line 1116: I've run into situations before where AWS will cry foul if I do too many operations at a time.  If this patterns repeats itself frequently then we might consider some sort of `batched_gather` that runs at most 20 at a time or something.

s3fs/core.py
```diff
@@ -298,9 +298,13 @@ def connect(self, refresh=True):
...
                              in client_kwargs.items() if key not in drop_keys}
             config_kwargs["signature_version"] = UNSIGNED
-        conf = Config(**config_kwargs)
-        self.s3 = self.session.create_client('s3', config=conf, **init_kwargs, **self.client_kwargs)
-        return self.s3
+        conf = AioConfig(**config_kwargs)
+        self.session = aiobotocore.get_session(**self.kwargs)
+        s3creator = self.session.create_client('s3', config=conf, **init_kwargs, **client_kwargs)
+        self._s3 = await s3creator.__aenter__()
+        return self._s3
```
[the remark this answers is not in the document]
martindurant: All the examples use the client in an async-with context, but we want to keep it around to reuse the connections, so I think there's no other way.

s3fs/core.py
martindurant: For the filepath management, I only plumbed in as much as seemed required to make the bulk operations work. So _info is async (calls head_object, or gets from cache) because copy needs to know how big the file is. I want to try to steer clear of writing async versions of all of the default implementations in AbstractFileSystem for now.

s3fs/core.py
martindurant: I am thinking that the exponential backoff retries are enough. I have no trouble pipe-ing and cat-ing 1300 small files in one go.

mrocklin: Ah!  So each of the sub-functions have their own backoff strategy.  Neat.

martindurant: Anything that goes through _call_s3, yes. There may still be a couple of exceptions to this, but fewer than before, so this rewrite has helped with that.

requirements.txt
```diff
@@ -1,2 +1,2 @@
-botocore>=1.12.91
-fsspec>=0.6.0
+aiobotocore
```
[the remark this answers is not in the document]
martindurant: (note that the fsspec version requirement will be updated following its release, maybe call it 0.8.0)

franz101:
added ability to use different sessions (profiles):
https://github.com/dask/s3fs/pull/339

hellocoldworld:
Hey! I reworked an entire microservice using this new feature! Many many thanks to everyone involved in this. Would you consider releasing a new version to pypi?

martindurant:
@hellocoldworld - release is currently blocked on https://github.com/pandas-dev/pandas/pull/35655 , since we don't want to break the pandas test suite :|

martindurant:
(PS: would love to see your use-case, if it can be public)

PR #339, opened 2020-08-01
Added session handling in S3FileSystem

franz101:
When trying to use an aws profile I ran into problems.
Looking into it the session argument is not used in the initialisation.
Therefore if the argument is provided it should be used


martindurant:
Thanks for this - something that got lost in the wash :)

martindurant:
Actually, I think that `self.session` gets overwritten in `connect` - in the case that a session was passed, only the client creation at the end of that function should need to be called.

PR #344, opened 2020-08-13
Update index.rst

hellocoldworld:
Text and code are in contradiction here.. I think the code is valid and there was a typo in the text. Also shall this section be expanded? It's not very clear to a newcomer like me.


martindurant:
Ooops, you are quite right.
Honestly, using the async mode is not meant for anyone that isn't an expert in asyncio (it's even almost beyond me!). Any help making what is really going on clearer would, of course be appreciated.
Meanwhile, thanks for this fix.

PR #345, opened 2020-08-14
add sign to filesystem class

CJ-Wright:
This will enable downstream classes to specify how to sign urls, providing short term URLs for situations where the URL is needed but permissioning needs to be tightly controlled.

Note that currently this adds a boto3 dep, which I'd like to discuss.

xref: https://github.com/intake/filesystem_spec/pull/376

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -1209,13 +1209,9 @@ class S3FileSystem(AsyncFileSystem):
             # This path is a bucket or folder, which do not currently have a modified date
             raise IsADirectoryError
         return info['LastModified'].replace(tzinfo=None)
-    def sign(self, path, expiration=100):
-        import boto3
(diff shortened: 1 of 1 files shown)


CJ-Wright:
Do you want me to make a alias for `sign`? The behavior between s3fs and gcsfs doesn't seem to agree.

martindurant:
> Do you want me to make a alias for sign

Probably the best, yes; and indeed sorry that "url" has a different meaning for gcsfs.

CJ-Wright:
Ok I'll update this PR.

I have an issue over [on GCS](https://github.com/dask/gcsfs/pull/277) for this, mostly I think that is a license discussion.

CJ-Wright:
@martindurant can this go in? (Along with the `fsspec` base class?)

PR #356, opened 2020-09-04
handle BucketAlreadyOwnedByYou when storing xarray Dataset

leifdenby:
When attempting to store an xarray `Dataset` in zarr format into my own
minio object-store I was getting the exception BucketAlreadyOwnedByYou
from botocore even though the `makedirs` called had `exist_ok=True`.
Looking at the s3fs source-code it appears this situation wasn't handled
before, but it is now.

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -443,7 +443,6 @@ class S3FileSystem(AsyncFileSystem):
         path = self._strip_protocol(path).rstrip('/')
         bucket, key, _ = self.split_path(path)
         if not key or (create_parents and not await self.exists(bucket)):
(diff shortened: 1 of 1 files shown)


martindurant:
Does this happen also on moto? It would be good to have a test, for existing bucket: 
- mkdir errors
- makedirs(exist_ok=True) doesn't error
- makedirs(exist_ok=False) errors

martindurant:
Fixes https://github.com/intake/filesystem_spec/issues/397 ?

martindurant:
Looking at the code, it seems straight-forward, putting this in.

PR #361, opened 2020-09-09
Maybe sync

martindurant:
Allow use of async everywhere


martindurant:
ref https://github.com/intake/filesystem_spec/pull/399

PR #362, opened 2020-09-09
add passthrough extras for boto3 and awscli

### LINKED ISSUE #357: Dependency on aiobotocore renders packaging s3fs no longer possible
I don't suppose there's anything that can be done about this. Nevertheless, I'd like to point out that due to dependency on aiobotocore we are no longer able to package new versions s3fs on Gentoo. FWICS the few other distributions that used to package s3fs are having the same problem.

The problem is that aiobotocore is a huge hack on internals of botocore, and therefore it requires very specific version of botocore. With the daily release cycle of botocore and rare releases of aiobotocore, we can't package it unless we repeatedly force old versions of botocore on our users.

graingert:
Fixes #357

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,4 +23,8 @@ setup(name='s3fs',
       packages=['s3fs'],
       python_requires='>= 3.7',
       install_requires=[open('requirements.txt').read().strip().split('\n')],
+      extras_require = {
+          'awscli': ['aiobotocore[awscli]'],
+          'boto3': ['aiobotocore[boto3]'],
+      },
       zip_safe=False)


graingert:
I'm not :100:% sure this will work - upload it to test.pypi.org first?

martindurant:
> upload it to test.pypi.org first?

What do I need to do?

graingert:
@martindurant https://packaging.python.org/guides/using-testpypi/

https://github.com/aio-libs/aiobotocore/pull/830/files

jrbourbeau:
@graingert you should be able to try out the `extras_require`s by installing directly from GitHub

```
pip install "git+https://github.com/graingert/s3fs@patch-1#egg=s3fs[boto3]"
```

graingert:
hmm it doesn't seem to work with pip-compile or pip 20.2 :(

https://gist.github.com/graingert/b8c9fa1c3a018310d35146ff24587b3d

martindurant:
This isn't right?
```
botocore==1.17.44 \
    --hash=sha256:1b46ffe1d13922066c873323186cbf97e77c137e08e27039d9d684552ccc4892 \
    --hash=sha256:1f6175bf59ffa068055b65f7d703eb1f748c338594a40dfdc645a6130280d8bb \
    # via aiobotocore
```

graingert:
it's supposed to install `boto3==1.14.44`

martindurant:
...but it's botocore that actually matters :)

graingert:
sure but it's only a problem if boto3 is added without a pin:

```
pip install s3fs boto3
```

graingert:
@martindurant https://gist.github.com/graingert/1f50abf7a145f6217d12e30dd5b69baf#file-pip-compile-out-L4

graingert:
@martindurant ok so it turns out delegated/passthrough extras don't work without `--use-feature=2020-resolver`, so this doesn't really help

thehesiod:
ya I've wanted to use this at work as well and found this bug: https://github.com/pypa/pip/issues/3516

martindurant:
Do you want to reboot this PR now?

graingert:
yep these work for me on pip21

graingert:
it's also probably worth yanking any of the old releases without `.whl` files, this way the 2020-resolver won't try to download and execute them - which is pretty slow, and in the future will be able to use `--use-feature=fast-deps` to download only the metadata portions of `.whl` files

PR #371, opened 2020-09-22
Cache results of find, and (optionally) include directories

martindurant:


martindurant:
Enables, for instance, much faster `du` after being called.
`find()` does *not* check for the cache, so it itself is not sped up by this. Can optionally add to check whether directories in the tree are *all* in the cache, and skip re-listing if yes.

PR #372, opened 2020-09-29
fixed resource leak in S3FileSystem._get_file()

sbj-ss:
S3FileSystem._get_file() currently doesn't close the S3 response stream - such behavior leaks resources and can render subsequent calls to this method unresponsive after opening the output file failed.

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -722,13 +722,16 @@ class S3FileSystem(AsyncFileSystem):
             self.s3.get_object, Bucket=bucket, Key=key,
             **version_id_kw(version_id or vers),
         )
-        with open(lpath, "wb") as f0:
-            body = resp['Body']
-            while True:
-                chunk = await body.read(2**16)
-                if not chunk:
-                    break
-                f0.write(chunk)
+        body = resp['Body']
+        try:
+            with open(lpath, "wb") as f0:
+                while True:
+                    chunk = await body.read(2**16)
+                    if not chunk:
+                        break
+                    f0.write(chunk)
+        finally:
+            body.close()
 
     async def _info(self, path, bucket=None, key=None, kwargs={}, version_id=None):
         if bucket is None:


martindurant:
Thanks for the change.
Were you seeing resource warnings, that prompted you to do this? I would have thought close (which, apparently, is not an async method) get called by the garbage collector when leaving the function. I think there are similar constructs elsewhere in the package.

sbj-ss:
I've just copied the line from [!another method](https://github.com/dask/s3fs/blob/16436dc006f27bb84d6d5e0ee93f6639cf28a373/s3fs/core.py#L626).
Calling `_get_file()` asynchronously with non-existing `lpath` directory caused it to hang on the next call; closing `Body` fixed things for me.

martindurant:
Fair enough!
That sounds like it could be put into a nice test function.

martindurant:
My test against this looks like the following, but it does *not* hang for me, even without the fix
```python
@pytest.mark.skipif(sys.version_info < (3, 7), reason="no asyncio.run in py36")
def test_async_close(s3):
    async def _():
        s3 = S3FileSystem(anon=False,
                          asynchronous=True,
                          loop=asyncio.get_running_loop(),
                          client_kwargs={"region_name": "eu-central-1",
                                         "endpoint_url": endpoint_uri})

        fn = test_bucket_name + "/nested/file1"
        s3.connect()  # creates client
        await s3._pipe_file(fn, b'data')

        with pytest.raises(OSError):
            await s3._get_file(fn, '/nonexistent')
        with pytest.raises(OSError):
            # previously would hang https://github.com/dask/s3fs/pull/372
            await s3._get_file(fn, '/nonexistent')

        await s3._s3.close()

    asyncio.run(_())
```

sbj-ss:
Heh. I've written a test that's closer to my code and looks like I can't repro the issue, too.
```python
import asyncio
import sys

import pytest
from s3fs import S3FileSystem

@pytest.mark.skipif(sys.version_info < (3, 7), reason="no asyncio.run in py36")
def test_async_close():
    loop = asyncio.get_event_loop()
    s3 = S3FileSystem(anon=False,
                      asynchronous=True,
                      loop=loop,
                      client_kwargs={"region_name": "eu-west-1"})
    loop.run_until_complete(s3._connect())

    fn = '<put your bucket and file name here>'

    async def async_wrapper():
        coros = [asyncio.ensure_future(s3._get_file(fn, '/nonexistent/a/b/c'), loop=loop) for _ in range(3)]
        completed, pending = await asyncio.wait(coros)
        for future in completed:
            with pytest.raises(OSError):
                future.result()
        print('async wrapper complete')

    for _ in range(2):
        loop.run_until_complete(async_wrapper())
    loop.run_until_complete(s3._s3.close())```
So it's up to you whether to accept the changes - but friendly speaking I'm afraid that the bug may strike again one day. You're closing `Body` in another `s3fs` method anyway - and in a synchronous manner, too, so this doesn't make any harm at least :)

martindurant:
Agreed, let's put it in, but I think your test is worth having too, as a rather more complicated async workflow than the test suite has so far.

martindurant:
(added the test in a new PR rather than pushing to your master branch - your original commit is retained in the history)

PR #374, opened 2020-10-02
Xarray test via dask threads 

martindurant:
A version of the bug shown in https://github.com/pydata/xarray/issues/4478


martindurant:
Fixes pydata/xarray#4478

Issue #330, opened 2020-06-17
Weird behavior With Versioned Files


jvwilbur:
I have a process that runs minutely. It Runs SQL queries, loads them into a dataframe, converts the dataframe into Parquet and uploads to S3 (in a versioned bucket), then uses the copy command to insert the data from S3 into Redshift. This is the first time I've ever left a bug report for any python library, so forgive me if it's written poorly.

```import s3fs

import boto3

s3 = s3fs.S3FileSystem(key=S3_KEY, secret=S3_SECRET, version_aware=True)
client = boto3.client('s3')

TRY S3FS TO RETRIEVE VERSION INFO
versions = s3.object_version_info('s3://bucket/foldert/sc_create_Job.parquet.gz')

TRY BOTO3 TO RETRIEVE VERSION INFO
response = client.list_object_versions(
        Bucket='bucket',
        Prefix='folder/sc_create_Job.parquet.gz'
    )
```

Attached are screenshots of the results, but here is the copy/paste text of them:
boto3 - 
```No output (True)```
s3fs -
```ClientError                               Traceback (most recent call last)

<ipython-input-4-914288398c97> in <module>
----> 1 versions = s3.object_version_info('s3://bucket/folder/sc_create_Job.parquet.gz')
      2 

~\AppData\Roaming\Python\Python36\site-packages\s3fs\core.py in object_version_info(self, path, **kwargs)
    534         while out['IsTruncated']:
    535             out = self._call_s3(self.s3.list_object_versions, kwargs,
--> 536                                 Bucket=bucket, Prefix=key, **self.req_kw)
    537             versions.extend(out['Versions'])
    538             kwargs['VersionIdMarker'] = out.get('NextVersionIdMarker', '')

~\AppData\Roaming\Python\Python36\site-packages\s3fs\core.py in _call_s3(self, method, *akwarglist, **kwargs)
    187         additional_kwargs = self._get_s3_method_kwargs(method, *akwarglist,
    188                                                        **kwargs)
--> 189         return method(**additional_kwargs)
    190 
    191     def _get_s3_method_kwargs(self, method, *akwarglist, **kwargs):

~\AppData\Local\Continuum\anaconda3\lib\site-packages\botocore\client.py in _api_call(self, *args, **kwargs)
    355                     "%s() only accepts keyword arguments." % py_operation_name)
    356             # The "self" in this scope is referring to the BaseClient.
--> 357             return self._make_api_call(operation_name, kwargs)
    358 
[...]

PR #375, opened 2020-10-05
Add key marker for paginated version info calls

ajkerrigan:
Hello s3fs maintainers! I ran into the same issue raised in #330. It pops up when I try to use `object_version_info` for a prefix with more versions than can fit in a single API response. This change resolves that issue.

Long-time happy user / first-time contributor here... so let me know if you have any questions/comments/etc about the change. Thanks!

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -887,7 +887,8 @@ class S3FileSystem(AsyncFileSystem):
             out = self.call_s3(self.s3.list_object_versions, kwargs,
                                Bucket=bucket, Prefix=key, **self.req_kw)
             versions.extend(out['Versions'])
-            kwargs['VersionIdMarker'] = out.get('NextVersionIdMarker', '')
+            kwargs.update({'VersionIdMarker': out.get('NextVersionIdMarker', ''),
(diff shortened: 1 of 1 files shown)


martindurant:
Thanks for the PR!
Do you know how many entries it would take to trigger the pagination - would it be reasonable to write a test against this?

ajkerrigan:
Hmm, good call :). By default it's 1000, but we should be able to force a lower number for testing. At a glance, I'm thinking we can add something like `fs.req_kw['MaxKeys'] = 1` inside the test? I'll take a look.

ajkerrigan:
Ah sorry, scratch that. There's already a [test](https://github.com/dask/s3fs/blob/d1b3efc/s3fs/tests/test_s3fs.py#L1252-L1263) listing 1200 versions of an object, but it seems to pass as-is because of differences between moto and s3. So without hitting a live s3 endpoint I'm not sure how to effectively test the change.

martindurant:
cc @mariusvniekerk

mariusvniekerk:
Yeah 1000 is the default limit

PR #389, opened 2020-11-02
Allow cat_file to fetch only specified range

martindurant:


martindurant:
Also fixes a bug in `_cat_file`, which was not respecting requester_pays.

PR #390, opened 2020-11-02
Add a simple test suite for doing downstream tests

mariusvniekerk:


martindurant:
Does this need to be merged before the first tests would run?

mariusvniekerk:
Yep it does

martindurant:
Can we please do this in two stages, one to add the CI, and then another to add the tests?

jorisvandenbossche:
@mariusvniekerk do the tests actually pass at the moment?

mariusvniekerk:
Rebased to put it into the ci.yml to see if they would trigger.

mariusvniekerk:
They don't :(

martindurant:
The run is here: https://github.com/mariusvniekerk/s3fs/actions/runs/410872762

but apparently there is a syntax problem

mariusvniekerk:
There we go

martindurant:
That's a lot of "❌"!

mariusvniekerk:
How big do we want the test matrix to be?

martindurant:
Er, it does seem like a lot. It would be enough to be python versions on linux and one version for win. I wouldn't even bother with osx... 

Note that fsspec runs the s3fs test suite on all CI runs, and dask does indeed have a couple of s3 tests.

PR #398, opened 2020-11-18
Clean up description of async methods

drwelby:
I think this reads clearer

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -87,8 +87,7 @@ Async
 -----
 
 ``s3fs`` is implemented using ``aiobotocore``, and offers async functionality.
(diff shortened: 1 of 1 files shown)


martindurant:
I am not certain; the point is that the `async def _*` are the actual definitions, and what you will find in the code, but sync versions without the "_" are generated at runtime *if* an explicit sync version isn't given.

drwelby:
This relates to #399 so this change is not a correct interpretation. Once I have more experience with the async behavior I'll give this section another shot. Closing.

martindurant:
Thanks @drwelby . I'm certain the docs could be more complete and clear.

PR #403, opened 2020-12-08
Return version_id even if file is empty

### LINKED ISSUE #402: version_id is None when writing empty object to S3
**What happened**:
When I use s3fs to write a file to s3, it doesn't return its version id if the object is empty. The object is correctly written.

It should return its version id, similar to what happens with similar call done with boto3

```python
import s3fs

s3 = s3fs.S3FileSystem(
    default_cache_type="none",
    requester_pays=True,
    default_fill_cache=False,
    version_aware=True
)

with s3.open("s3://bucket/test1.json", "wb") as f:
    #f.write(b'{"a":1}\n') <= no problem if I write some content
    f.write(b'') # <= if I write this, f.version_id is None

print(f.version_id)
# I tried to invalidate the cache...no difference both for success and failure scenarios
s3.invalidate_cache()
print(f.version_id)
```

**Environment**: 

- Python version: 3.7.5
- Operating System: tried MacOS and Linux
- Install method (conda, pip, source): tried pip 0.4.0, 0.5.0, 0.5.1 and source (latest commit)

matteofigus:
Fixes #402 

I did some debugging and managed to resolve the issue. I hope this helps.

Any thoughts?

Thanks

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -718,7 +718,9 @@ class S3FileSystem(AsyncFileSystem):
         if not truncate and self.exists(path):
             raise ValueError("S3 does not support touching existent files")
         try:
-            self.call_s3(self.s3.put_object, kwargs, Bucket=bucket, Key=key)
+            write_result = self.call_s3(self.s3.put_object, kwargs, Bucket=bucket, Key=key)
+            if self.version_aware:
+                self.version_id = write_result.get("VersionId")
         except ClientError as ex:
             raise translate_boto_error(ex) from ex
         self.invalidate_cache(self._parent(path))
@@ -1708,6 +1710,9 @@ class S3File(AbstractBufferedFile):
                 logger.debug("Empty file committed %s" % self)
                 self._abort_mpu()
                 self.fs.touch(self.path)
+                if self.fs.version_aware and self.fs.version_id:
+                    self.version_id = self.fs.version_id
+
         elif not self.parts:
             if self.buffer is not None:
                 logger.debug("One-shot upload of %s" % self)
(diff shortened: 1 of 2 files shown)


martindurant (commented):
Thanks for having a go at fixing the problem before I've even had a chance to look into it!
I think there's a slightly simpler way to do this, see the comment in the code. Thanks also for providing a comprehensive test case.

matteofigus:
> Thanks for having a go at fixing the problem before I've even had a chance to look into it!
> I think there's a slightly simpler way to do this, see the comment in the code. Thanks also for providing a comprehensive test case.

No problem at all, happy to help! I don't see the comment in the code?

martindurant:
Hmm, seems to have disappeared - but I did write it, honest! To repeat...

I think it would b cleaner to have `touch` return the variable `write_result`, rather than set an attribute on the filesystem instance (it's an attribute of a specific implicit file, not of the whole file system). If `touch` gives this information to the caller, then you can handle it directly in `commit`.

matteofigus:
I think you're right. After returning the `write_result`, I noticed we were actually repeating the `version_aware` check 3 times, so I just managed to move the check at the end which simplifies the flow a bit. What do you think?

martindurant:
It does look like a good simplification

matteofigus:
Thanks a lot for the quick merge, and thanks for working on this awesome library.

Can I ask you what's the process to get the change released in a new version available in Pypi? Anything I can do to help there?

martindurant:
> Anything I can do to help there?

Keep prodding me until I et around to doing it. I will have to check, a new release of fsspec might be needed first, and unfortunately we haven't migrated its CI to GH-actions yet, so that might hold things up.

matteofigus:
> Keep prodding me until I get around to doing it

Hi! Any news? 
Thanks :)

martindurant:
Done.

matteofigus:
Awesome! Thanks! 👍

Issue #404, opened 2020-12-16
Exists check fails starting from version 0.5.1 on OpenStack Object Storage ("Swift")


totycro:
**What happened**:
When using s3fs 0.5.1 or 0.5.2, I get the following error when triggering an "exists" check for example using `in` when accesing the Object Storage of the Open Telekom Cloud, which is an OpenStack deployment. It seems to not feature the key `KeyCount`, and i'm not sure if the problem is rather in s3fs or in the server implementation.  
However I found it noteworthy that this problem doesn't seem to occur with s3fs 0.5.0. 
```
  File "read_bucket.py", line 7, in <module>
    print('a' in store)
  File "/home/mallinger/tmp/s3fs-test/lib/python3.8/site-packages/fsspec/mapping.py", line 169, in __contains__
    return self.fs.exists(path) and self.fs.isfile(path)
  File "/home/mallinger/tmp/s3fs-test/lib/python3.8/site-packages/fsspec/asyn.py", line 121, in wrapper
    return maybe_sync(func, self, *args, **kwargs)
  File "/home/mallinger/tmp/s3fs-test/lib/python3.8/site-packages/fsspec/asyn.py", line 100, in maybe_sync
    return sync(loop, func, *args, **kwargs)
  File "/home/mallinger/tmp/s3fs-test/lib/python3.8/site-packages/fsspec/asyn.py", line 71, in sync
    raise exc.with_traceback(tb)
  File "/home/mallinger/tmp/s3fs-test/lib/python3.8/site-packages/fsspec/asyn.py", line 55, in f
    result[0] = await future
  File "/home/mallinger/tmp/s3fs-test/lib/python3.8/site-packages/s3fs/core.py", line 566, in _exists
    await self._info(path, bucket, key, version_id=version_id)
  File "/home/mallinger/tmp/s3fs-test/lib/python3.8/site-packages/s3fs/core.py", line 740, in _info
    if out["KeyCount"] > 0:
KeyError: 'KeyCount'
```

**What you expected to happen**:
The example should print True or False.

**Minimal Complete Verifiable Example**:

```python
import s3fs
s3 = s3fs.S3FileSystem(
    client_kwargs={"endpoint_url": "https://obs.eu-de.otc.t-systems.com"}
)
store = s3fs.S3Map("s3://REDACTED/REDACTED", s3=s3)
print('a' in store)
```

**Anything else we need to know?**:

**Environment**:

[...]

PR #405, opened 2020-12-16
Handle list_object_v2 calls with missing KeyCount

totycro:
Some s3 server implementations such as OpenStack Object Storage
("Swift") don't provide the field KeyCount in the response to list_object_v2.

This change handles this gracefully and reverts to checking other keys
("Contents" and "CommonPrefixes").

See https://github.com/dask/s3fs/issues/404 for more details.

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -890,7 +890,7 @@ class S3FileSystem(AsyncFileSystem):
                 MaxKeys=1,
                 **self.req_kw,
             )
-            if out["KeyCount"] > 0:
+            if out.get("KeyCount", 0) > 0 or out.get("Contents", []) or out.get("CommonPrefixes", []):
                 return {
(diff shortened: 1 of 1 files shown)


s3fs/core.py
```diff
@@ -890,7 +890,7 @@ async def _info(self, path, bucket=None, key=None, kwargs={}, version_id=None):
                 MaxKeys=1,
                 **self.req_kw,
             )
-            if out["KeyCount"] > 0:
+            if out.get("KeyCount", 0) > 0 or out.get("Contents", []) or out.get("CommonPrefixes", []):
```
martindurant, line 893: Note that this line is probably too long for the linter, sorry :|

martindurant:
Sorry, forgot to press the button here.
Thanks for the fix!

Issue #317, opened 2020-04-23
Cannot delete directory


sweco:
I'm unable to delete an S3 directory, i.e. a path ending with `/` using `rm` nor `bulk_delete`. There is no error raised but the actual directory does not get deleted.

Code to reproduce:
```python
# Create a folder in an S3 bucket via S3 console
s3fs = S3FileSystem()
s3fs.rm('bucket/folder/')
# Observe the directory remained in S3
```

I guess this is the result of calling the `S3FileSystem.split_path(path)` which calls the `AbstractFileSystem._strip_protocol(path)` which also strips the trailing slash.

mariusvniekerk:
S3 doesn't actually have directories, only files

swecooo:
You're right, I basically meant a path ending with `/` which then appears as a directory in the S3 console.

martindurant:
Agree, stripping of trailing "/" should only happen at list time, not for rm/open. Uncertain about info.

martindurant:
Will port https://github.com/dask/gcsfs/issues/273 to fix this

abulka:
In the current `s3fs 0.4.2` I still cannot delete a 

>  path ending with / which then appears as a directory in the S3 console.

even though `s3fs.exists('bucket/folder/')` reports `True`.  I'm having to work around it by going to native `boto3`

```python
s3resource = boto3.resource('s3', ...)
s3resource.Object('bucket', 'folder/').delete()
```

## Repro

Create the offending dir (yeah I know there are really no such things as folders in S3, only keys with a trailing `/`)
```python
BUCKET = 'bucket'
s3client = boto3.client('s3', ...)
s3client.put_object(Bucket=BUCKET, Key=('folder/'))
```

s3fs can see it but can't delete it

```python
print(fs.exists(os.path.join(BUCKET, 'folder/')))  # True
fs.rm('bucket/folder/') # FileNotFoundError
fs.rm('bucket/folder/', recursive=True)  # FileNotFoundError
```

The workaround
```python
s3resource.Object(BUCKET, 'folder/').delete()
```

martindurant:
Does the placeholder key show up with `find`, and does it have the "/" on the end? Does it show up as a file or a directory?

rom1504:
I'm hitting a similar issue when trying to use fs.rm(folder, recursive=True) : it works (everything is deleted) but fails with FileNotFoundError for the folder

martindurant:
Thanks for the ping, I'll look into it.

[...]

PR #407, opened 2021-01-04
Support for S3 folder in s3fs

### LINKED ISSUE #401: Unable to create a directory
**What happened**:
Trying to create a directory in a bucket and it silently fails.

Expect a key created in a bucket.

```python
from s3fs import S3FileSystem
s3 = S3FileSystem(anon=False)
s3.makedirs('s3://my-bucket-name/testdir', exist_ok=False) #fails to create 
s3.makedirs('my-bucket-name/testdir', exist_ok=False) #fails to create
```
How do I create a directory in a bucket? Am I doing something wrong or is it a bug?

s3fs version == 0.5.0

**Environment**:

- Dask version: 
- Python version: 3.7.7
- Operating System: Ubuntu Linux
- Install method (conda, pip, source): conda

spsoni:
A comprehensive solution for creating a folder on S3 to resolve #401 

AWS has a concept of S3 folders. Check https://docs.aws.amazon.com/AmazonS3/latest/user-guide/using-folders.html

> The Amazon S3 console treats all objects that have a forward slash "/" character as the last (trailing) character in the key name as a folder, for example examplekeyname/. You can't upload an object that has a key name with a trailing "/" character using the Amazon S3 console. However, you can upload objects that are named with a trailing "/" with the Amazon S3 API by using the AWS CLI, AWS SDKs, or REST API.

Cleared few TODO's and enriched test cases.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ __pycache__
 dist/
 *.egg-info
 build/
+venv*
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -1023,18 +1023,12 @@ class S3FileSystem(AsyncFileSystem):
         if "/" not in path:
             return super(S3FileSystem, self).isdir(path)
 
-        path = path.rstrip("/") + "/"
-
         if path in self.dircache:
             for fp in self.dircache[path]:
                 # For files the dircache can contain itself.
                 # If it contains anything other than itself it is a directory.
                 if fp["name"] != path:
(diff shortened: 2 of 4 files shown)


spsoni:
Fixes #156 too

spsoni:
Potentially fixes #317 too

spsoni:
I will leave it open for someone interested in the future for reference

PR #409, opened 2021-01-07
README.rst update to point to aiobotocore documentation i

spsoni:
instead of botocore documentation

requirements.txt no longer points to botocore as its core dependency


martindurant:
Thanks

PR #411, opened 2021-01-13
fixes #410

### LINKED ISSUE #410: BUG: recursive find (with maxdepth=None) alters directories as side-effect
Debugging the failures with s3fs 0.5.2 (compared to 0.4.2) in the pyarrow filesystem tests, I noticed that a recursive `find` "alters" the directories, in the sense that the info about the directory fails, and also no longer is found by a next non-recursive find.

Output from debugging to illustrate:

```
(Pdb) s3.info("test/dir")
{'Key': 'test/dir', 'Size': 0, 'StorageClass': 'DIRECTORY', 'type': 'directory', 'size': 0, 'name': 'test/dir'}
(Pdb) infos2 =  s3.find(test_bucket_name, maxdepth=None, withdirs=True, detail=True)
(Pdb) s3.info("test/dir")
{'name': 'test/dir', 'size': 0, 'type': 'directory'}
```

So the first time and the second time I ask for the info of "test/dir" gives a different result, caused by the recursive find (`maxdepth=None`) in the middle between both calls.

As a result, all those "altered" directories are not found anymore when doing a non-recursive `find(.., maxdepth=1)` afterwards (and those were correctly found before that).

A test that can be added to `test_s3fs.py` that reproduces the issue:

```
def test_find_no_side_effect(s3):
    infos1 =  s3.find(test_bucket_name, maxdepth=1, withdirs=True, detail=True)
    infos2 =  s3.find(test_bucket_name, maxdepth=None, withdirs=True, detail=True)
    infos3 =  s3.find(test_bucket_name, maxdepth=1, withdirs=True, detail=True)
    assert infos1.keys() == infos3.keys()
```

martindurant:


martindurant:
@jorisvandenbossche - included your test verbatim

jorisvandenbossche:
Cool, thanks!

martindurant:
Please wait - not *quite* right :)

PR #412, opened 2021-01-14
Avoid adding parent dir in find

martindurant:
Fixes #394

PR #413, opened 2021-01-15
Add test against #393

### LINKED ISSUE #393: BUG: the find() method is now also returning the specified directory to search
Compared to s3fs 0.4, the new 0.5 release changed the behaviour of `S3FileSystem.find()`. This has been partly fixed already on master I think (eg https://github.com/dask/s3fs/issues/378, to respect again the `maxdepth` keyword (to allow non-recursive finds), and the fix for this also seems to have fixed other things for pyarrow, i.e. to include sub-directories in the output, and not only the files). 
But from testing it with pyarrow, I noticed another difference that is still present on master: it now also returns the "current" directory that is passed to `find(..)`.

So when doing `fs.find(base_dir)`, this gives a list of files/directories in "base_dir". But with s3fs 0.5/master, this now also includes "base_dir" itself. 
That behaviour is in contrast with s3fs0.4, but also with eg fsspec's LocalFileSystem.

I have a reproducible test script included below, but this is using a slimmed down and adapted version of a pyarrow test, so using pyarrows test machinery (with a fixture to set up minio to test), so this will need some edits to be includable in the s3fs tests.

<details>

```python
import pathlib
import sys
import os
import subprocess
from tempfile import TemporaryDirectory

import pytest

from pyarrow.util import find_free_port


@pytest.fixture
def tempdir(tmpdir):
    # convert pytest's LocalPath to pathlib.Path
    return pathlib.Path(tmpdir.strpath)


@pytest.fixture(scope='session')
def s3_connection():
    host, port = 'localhost', find_free_port()
    access_key, secret_key = 'arrow', 'apachearrow'
    return host, port, access_key, secret_key


@pytest.fixture(scope='session')
def s3_server(s3_connection):
    host, port, access_key, secret_key = s3_connection

    address = '{}:{}'.format(host, port)
    env = os.environ.copy()
    env.update({
        'MINIO_ACCESS_KEY': access_key,
        'MINIO_SECRET_KEY': secret_key
    })

    with TemporaryDirectory() as tempdir:
        args = ['minio', '--compat', 'server', '--quiet', '--address',
                address, tempdir]
        proc = None
        try:
            proc = subprocess.Popen(args, env=env)
        except OSError:
            pytest.skip('`minio` command cannot be located')
        else:
            yield proc
        finally:
            if proc is not None:
                proc.kill()


@pytest.fixture
def fsspec_localfs(request, tempdir):
    fsspec = pytest.importorskip("fsspec")
    fs = fsspec.filesystem('file')
    return fs, lambda p: (tempdir / p).as_posix()


@pytest.fixture
def fsspec_s3fs(request, s3_connection, s3_server):
    s3fs = pytest.importorskip("s3fs")

    host, port, access_key, secret_key = s3_connection
    bucket = 'pyarrow-filesystem/'

    fs = s3fs.S3FileSystem(
        key=access_key,
        secret=secret_key,
        client_kwargs=dict(endpoint_url='http://{}:{}'.format(host, port))
    )
    try:
        fs.mkdir(bucket)
    except IOError:
        # BucketAlreadyOwnedByYou on second test
        pass

    return fs, bucket.__add__


@pytest.fixture(params=[
    pytest.param(
        pytest.lazy_fixture('fsspec_localfs'),
        id='localfs'
    ),
    pytest.param(
        pytest.lazy_fixture('fsspec_s3fs'),
        id='s3fs)'
    ),
])
def filesystem_config(request):
    return request.param


def test_get_file_info_with_selector(filesystem_config):
    fs, pathfn = filesystem_config

    base_dir = pathfn('selector-dir/')
    file_a = pathfn('selector-dir/test_file_a')
    file_b = pathfn('selector-dir/test_file_b')
    dir_a = pathfn('selector-dir/test_dir_a')
    file_c = pathfn('selector-dir/test_dir_a/test_file_c')

    try:
        fs.mkdir(base_dir)
        with fs.open(file_a, mode="wb"):
            pass
        with fs.open(file_b, mode="wb"):
            pass
        fs.mkdir(dir_a)
        with fs.open(file_c, mode="wb"):
            pass

        infos = fs.find(base_dir, maxdepth=None, withdirs=True, detail=True)
        assert len(infos) == 4

        for info in infos.values():
            if info["name"].endswith(file_a):
                assert info["type"] == "file"
            elif info["name"].endswith(file_b):
                assert info["type"] == "file"
            elif info["name"].endswith(file_c):
                assert info["type"] == "file"
            elif info["name"].rstrip("/").endswith(dir_a):
                assert info["type"] == "directory"
            else:
                raise ValueError('unexpected path {}'.format(info["name"]))
    finally:
        fs.rm(base_dir, recursive=True)
```

</details>

martindurant:
Add test from the linked issue, which was already fixed in #412

Closes #393

### DIFF
--- a/s3fs/tests/test_s3fs.py
+++ b/s3fs/tests/test_s3fs.py
@@ -1839,11 +1839,11 @@ def test_find_no_side_effect(s3):
 
 def test_get_file_info_with_selector(s3):
     fs = s3
-    base_dir = 'selector-dir/'
-    file_a = 'selector-dir/test_file_a'
-    file_b = 'selector-dir/test_file_b'
-    dir_a = 'selector-dir/test_dir_a'
-    file_c = 'selector-dir/test_dir_a/test_file_c'
+    base_dir = "selector-dir/"
+    file_a = "selector-dir/test_file_a"
+    file_b = "selector-dir/test_file_b"
+    dir_a = "selector-dir/test_dir_a"
+    file_c = "selector-dir/test_dir_a/test_file_c"
 
     try:
         fs.mkdir(base_dir)
@@ -1868,6 +1868,6 @@ def test_get_file_info_with_selector(s3):
             elif info["name"].rstrip("/").endswith(dir_a):
                 assert info["type"] == "directory"
             else:
-                raise ValueError('unexpected path {}'.format(info["name"]))
+                raise ValueError("unexpected path {}".format(info["name"]))
     finally:
         fs.rm(base_dir, recursive=True)
--- a/s3fs/tests/test_s3fs.py
+++ b/s3fs/tests/test_s3fs.py
@@ -1835,3 +1835,39 @@ def test_find_no_side_effect(s3):
     s3.find(test_bucket_name, maxdepth=None, withdirs=True, detail=True)
     infos3 = s3.find(test_bucket_name, maxdepth=1, withdirs=True, detail=True)
     assert infos1.keys() == infos3.keys()
+
+
+def test_get_file_info_with_selector(s3):
+    fs = s3
+    base_dir = 'selector-dir/'
+    file_a = 'selector-dir/test_file_a'
+    file_b = 'selector-dir/test_file_b'
+    dir_a = 'selector-dir/test_dir_a'
+    file_c = 'selector-dir/test_dir_a/test_file_c'
+
+    try:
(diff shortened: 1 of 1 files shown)


martindurant:
Thanks again for the reproducer, @jorisvandenbossche

Issue #421, opened 2021-02-17
Unexplainable size limit for switching to multipart upload


andreasg123:
This isn't an observed bug. However, when looking at the code, I found it odd that `_put_file` and `_pipe_file` switch to multipart at 5 MB but use 50 MB chunks by default. In my opinion, the limit should be the specified chunk size.

https://github.com/dask/s3fs/blob/38bceeed4bc7de6db6dc06ce67149aa9a8f49699/s3fs/core.py#L789-L796

martindurant:
If possible, we prefer the simple upload, because it requires one call instead of three. The 5MB limit is set by AWS as the maximum size for PUT. I cannot now find a reference for this, it is possible that the limit has been removed.

andreasg123:
I found that AWS sets a 5 **GB** limit for a PUT: https://aws.amazon.com/s3/faqs/

> The largest object that can be uploaded in a single PUT is 5 gigabytes

Multipart upload is recommended for objects larger than 100 MB.

martindurant:
You'll have to trust me that it was definitely 5**MB** in the past :|

You would be most welcome to make a PR in which is given blocksize is used to decide between single PUT and multi-part (although we should force the blocksize to now exceed 5GB, if that even should happen).

PR #425, opened 2021-02-17
PUT S3 objects up to 5 GB

andreasg123:
This addresses #421. For data of more than twice the chunk size specified for `_pipe_file` and `_put_file`, a multipart upload will be used. The hard limit of 5 GB is also enforced.

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -752,7 +752,8 @@ class S3FileSystem(AsyncFileSystem):
     async def _pipe_file(self, path, data, chunksize=50 * 2 ** 20, **kwargs):
         bucket, key, _ = self.split_path(path)
         size = len(data)
-        if size < 5 * 2 ** 20:
+        # 5 GB is the limit for an S3 PUT
+        if size < min(5 * 2 ** 30, 2 * chunksize):
             return await self._call_s3(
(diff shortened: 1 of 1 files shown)


s3fs/core.py
```diff
@@ -752,7 +752,8 @@ async def _cat_file(self, path, version_id=None, start=None, end=None):
     async def _pipe_file(self, path, data, chunksize=50 * 2 ** 20, **kwargs):
         bucket, key, _ = self.split_path(path)
         size = len(data)
-        if size < 5 * 2 ** 20:
+        # 5 GB is the limit for an S3 PUT
+        if size < min(5 * 2 ** 30, 2 * chunksize):
```
martindurant, line 756: Why 2x the chunksize? I think it is reasonable to split an upload into 1-and-a-bit parts.
Are you thinking, because the MPU route takes a minimum of three calls?

martindurant:
Closes #421

s3fs/core.py
andreasg123: I used that as a heuristics to make the overhead for the multipart upload worthwhile. Also, the AWS recommendation was to use multipart upload for files starting at 100 MB that happens to be twice the default chunk size.

martindurant: OK, I'll take it :)

Issue #422, opened 2021-02-17
Read invalid file content, due to file-size "caching"


anetbnd:
Bug reports that follow these guidelines are easier to diagnose, and so are often handled much more quickly.
-->

**What happened**:
If a `S3File` object is read, it can happen that invalid content is read, due to internal file-size caching in the variable `S3File.size`, if between opening the file and reading it the file in the S3 has changed. This happens, because the file size if stored into the object variable when opening the file, while the actual reading from the file happens only when the `read()` method of the `S3File` object is called. If the "all" cache-type is used for this file, this triggers a `get_object` request on S3 with a range attribute of `0-self.size`.

Example (see code below):
1.) The file is written with 3 bytes to the bucket.
2.) We can read the file without issues.
3.) Then we open the file again (size is setup with value 3)
4.) But now we change the content to have more bytes
5.) When reading now, we just read 3 bytes instead the full content

**What you expected to happen**:

It is clear, that s3fs cannot handle those race-conditions well. There will always be some kind of caching which could lead to invalid behavior when other processes change data in the background. However, at the moment s3fs just silently delivers any content and if the higher layer application cannot check if the content makes sense, it could happen that the read error is not noticed and lead to many other errors in the application. 

My expectation would be that in those cases s3fs notifies the caller with a clear exception, which states that the file has been changed after opening it, thus is cannot be read anymore in a safe manner. This can easily achieved by adding the `IfMatch` attribute to the `get_object` call of boto3 and use here the last read ETag, which is already stored in the directory `S3File.details`. In this case boto3 would generate an exception, telling the user that the file has changed. 

The simplest solution would be to add it in the `__init__` method of `S3File`:

```
        # ....

        if 'r' in mode:
            self.req_kw['IfMatch'] = self.details['ETag']
``` 

Then the application of the user can catch this exception and decide what to do. For example read the file another time.
To make it specially clean, it would also be possible to translate the exception from a boto3 exception into a exception, which tells the user, that the file has changed between opening and reading. 

**Minimal Complete Verifiable Example**:

In this example, I simulate the concurrent write with a direct boto3 call, to not accidentally change an inner state of the s3fs object (any caches or something like that). 

```python
import s3fs
import boto3
import botocore.config


PROXY = None
AWS_KEY = '...'
AWS_SECRET = '...'
USE_SSL = False
BUCKET = '...'
FILE = 'test.file'
FILE_PATH = f's3://{BUCKET}/{FILE}'

[...]

PR #427, opened 2021-02-18
Bugfix/fix tests for windows

anetbnd:
In my preparation for the fix of issue https://github.com/dask/s3fs/issues/422 I first fixed some issues with the test-suite, which I noticed on my windows 10 system. More details are in the comments.

I happy to receive your feedback.

### DIFF
--- a/s3fs/tests/test_s3fs.py
+++ b/s3fs/tests/test_s3fs.py
@@ -1189,11 +1189,19 @@ def _get_s3_id(s3):
 
 
 @pytest.mark.skipif(sys.version_info[:2] < (3, 7), reason="ctx method only >py37")
-@pytest.mark.parametrize("method",
-                         [
-                             "spawn",
-                             pytest.param("forkserver", marks=pytest.mark.skipif(sys.platform.startswith("win"), reason="'forserver' not available on windows"))
-                         ])
+@pytest.mark.parametrize(
+    "method",
+    [
+        "spawn",
+        pytest.param(
+            "forkserver",
+            marks=pytest.mark.skipif(
+                sys.platform.startswith("win"),
+                reason="'forserver' not available on windows",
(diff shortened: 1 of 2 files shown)


martindurant (commented):
Excellent, thank you!
I have the one query on botocore's version (note that we actually call aiobotocore)

anetbnd:
@martindurant You are welcome. I don't know, if I understand you correctly. You have concerns about the `botocore` request from the `test_requirements.txt`. Do you want to delete it completely or just do you want me to change it to a `aiobotocore` requirement (which might probably also limit the `botocore` version to the right range). Just tell me and I can adapt it.

martindurant:
aoibotocore is already in the requirements, so I don't think you need it here. Did you fail to install/test without it?
Pinning the version so tightly is a separate problem, because we would have to continually update both aoibotocore and botocore versions as new versions come it.

martindurant:
Since aiobotocore is in requirements.txt, I don't think you need it here; in any case, setting a precise pin like that would mean having to constantly update it.

martindurant:
(sorry, system reset meant you got two versions of this comment!)

anetbnd:
@martindurant No problem, I can delete it.

Yes, I had issues when installing the environment. I first installed the `requirements.txt` which lead to the right version of `botocore`. Then I installed the `test-requirements.txt`, which automatically updated the version to the latest one (due to missing restrictions in that file). However the new pip installer is afterwards showing a warning about the incompatible version with already installed packages. 

However it does not lead to any errors in the test-cases, I think it is just a formal error due to too tight version requirements from `aiobotocore` (I have exactly the same option about too tight version requirements like you). 

So I will delete it from the `test-requirements.txt`.

martindurant:
Thanks for this!

PR #428, opened 2021-02-19
Butfix/issue 422 read invalid data in race condition duo to outdated filesize

anetbnd:
**Attention: Branch is based on PR https://github.com/dask/s3fs/pull/427, please merge this before**

This pull-request addresses the issue https://github.com/dask/s3fs/issues/422.

It introduces a pre-condition for file-reads to check, if the ETag has been changed since reading the file-size. It works with S3 buckets, but not with the S3 Bucket mock `moto` version 1.3.16, since `moto` does not provide a pre-condition check in this version.

The pre-condition check is already on the master banch of moto and when installed the development version from moto locally, the corresponding test passes. However we still need to wait until the next release of moto.

See also the request for the next release here: https://github.com/spulec/moto/issues/3712

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -17,7 +17,7 @@ from aiobotocore.config import AioConfig
 from botocore.exceptions import ClientError, ParamValidationError
 
 from s3fs.errors import translate_boto_error
-from s3fs.utils import ParamKwargsHelper, _get_brange
+from s3fs.utils import ParamKwargsHelper, _get_brange, FileExpired
 
 logger = logging.getLogger("s3fs")
 
@@ -448,6 +448,16 @@ class S3FileSystem(AsyncFileSystem):
         kwargs: dict-like
             Additional parameters used for s3 methods.  Typically used for
             ServerSideEncryption.
+
+        Exceptions
+        ----------
+        FileExpired(OSError) :
+            Is raised, when the file content has been changed from a different process after
+            opening the file. Reading the file would lead to invalid or inconsistent output.
+            This can also be triggered by outdated file-information inside the directory cache.
+            In this case ``S3FileSystem.invalidate_cache`` can be used to force an update of
+            the file-information when opening the file.
+
         """
         if block_size is None:
             block_size = self.default_block_size
@@ -1583,6 +1593,9 @@ class S3File(AbstractBufferedFile):
                 self.append_block = True
             self.loc = loc
 
+        if "r" in mode:
+            self.req_kw['IfMatch'] = self.details['ETag']
+
     def _call_s3(self, method, *kwarglist, **kwargs):
         return self.fs.call_s3(method, self.s3_additional_kwargs, *kwarglist, **kwargs)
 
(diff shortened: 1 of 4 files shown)


martindurant:
You could mark the test with `@pytest.mark.xfail(condition)` where the condition would be for the *next* release of moto. Then this would show up as Xfail until moto is released, and then turn to PASS. Also, it would allow testing against docker minio, should we get around to implementing that, or against real AWS.

s3fs/core.py
```diff
@@ -448,6 +449,16 @@ def _open(
         kwargs: dict-like
             Additional parameters used for s3 methods.  Typically used for
             ServerSideEncryption.
+
+        Exceptions
+        ----------
+        FileExpired(OSError) :
```
martindurant, line 455: I'm not sure the long description is needed here since FileExpired has its own docstring and message

s3fs/utils.py
```diff
@@ -9,6 +10,15 @@ def ignoring(*exceptions):
         pass
 
 
+class FileExpired(IOError):
+    def __init__(self, filename: str, e_tag: str):
+        super().__init__(
+            errno.EBUSY,
```
martindurant, line 16: Maybe EBADFD?

martindurant (commented):
Looking good - some comments,

s3fs/core.py
anetbnd: I can move this description here to the docstring of the exception. This would make sense. In general it should be documented, that this error can happen in two circumstances:

1. On concurrent write/read accesses
2. When directory cache is outdated

Especially the second case is not straight forward an could lead to very long debugging sessions, if you don't know it.

s3fs/utils.py
anetbnd: Yes this was exactly the error code I chose first. However at python 3.8 (windows) it does not exist inside the errno. If you know a better one, which is available all different platforms, please suggest one.

martindurant: I did not know that! I don't think it matters too much, if the message is clear.

anetbnd:
@martindurant Thanks for your feedback. I updated both PRs with your requested changes. Keep in mind to first merge PR https://github.com/dask/s3fs/pull/427, before merging this one, since this one builds on the branch of the other PR (due to windows testing). If you have further requests, just tell me.

s3fs/tests/test_s3fs.py
```diff
@@ -1871,3 +1890,32 @@ def test_get_file_info_with_selector(s3):
                 raise ValueError("unexpected path {}".format(info["name"]))
     finally:
         fs.rm(base_dir, recursive=True)
+
+
+@pytest.mark.xfail(
+    skip=version.parse(moto.__version__) > version.parse("1.3.16"),
```
martindurant, line 1896: Should be `condition=`, and I think this is the wrong way around: this test is expected to fail is the current version is 1.3.16 or less. It should pass for the next moto.

anetbnd: Good catch! Thanks, I corrected it.

PR #434, opened 2021-03-02
Allow for zero start in _cat_file

### LINKED ISSUE #433: _cat_file cannot be called with zero start
I discovered this by reading the code. The test `if start:` in `_cat_file` will be falsy for zero, preventing a range request with a zero start and a non-zero end. The test should be `if start is not None:` instead.

andreasg123:
Fixes #433

I created the test first and verified that it failed without the fix in `_cat_file`.

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -734,7 +734,7 @@ class S3FileSystem(AsyncFileSystem):
         bucket, key, vers = self.split_path(path)
         if (start is None) ^ (end is None):
             raise ValueError("Give start and end or neither")
-        if start:
+        if start is not None:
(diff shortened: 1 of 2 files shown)


martindurant:
Perfect, thanks!

andreasg123:
@martindurant, do you plan another release soon? Without this fix, `_cat_file` cannot be used to get the first N bytes of an S3 object.

Use of the GitHub version of `s3fs` unfortunately requires the use of the GitHub version of `fsspec`, too, because of the import of `fsspec.utils.setup_logging` (that is immediately redefined in `core.py`).

martindurant:
s3fs and fsspec just went onto pypi. conda-forge releases will follow.

PR #435, opened 2021-03-05
setup: pin moto to under 2.0

isidentical:
I didn't had chance to debug it fully though moto>2.0 is incompatible with the tests as is (2 version check tests are failing) intake/filesystem_spec/pull/557

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,7 +33,7 @@ jobs:
       - name: Install dependencies
         shell: bash -l {0}
         run: |
-          conda install -c conda-forge pip botocore aiobotocore moto pytest flake8 black -y
+          conda install -c conda-forge pip botocore aiobotocore "moto<2.0" pytest flake8 black -y
           pip install git+https://github.com/intake/filesystem_spec --no-deps
           conda list
           conda --version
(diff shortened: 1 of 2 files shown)


martindurant:
( cc @mariusvniekerk , in case you know how versioning is different in moto 2.0 - no need for any action here)

.github/workflows/ci.yml
```diff
@@ -33,7 +33,8 @@ jobs:
       - name: Install dependencies
         shell: bash -l {0}
         run: |
-          conda install -c conda-forge pip botocore aiobotocore moto pytest flake8 black -y
+          conda install -c conda-forge pip botocore aiobotocore flake8 black -y
```
martindurant, line 36: why not include the moto version pin in the conda install line? pip could in theory step on conda packages, so we should keep conda-based tests and pip-based tests separate.

isidentical: I just thought maintaining the same list in 2 different places was redundant, though can revert.

martindurant: True, but I don't see a way around it when I trust conda more than pip

isidentical: alright dropped

martindurant:
Thanks!

PR #441, opened 2021-03-16
copy: implement preserve_etag mode

### LINKED ISSUE #440: Ability to copy big files while preserving etag
If a file is uploaded in multiple parts (multipart upload) the current s3fs strategy to copy is just using a static chunksize (5GB). Which might result in fewer number of API calls and the ETag being different than usual. Example scenerio;
```py
test_file1 = test_bucket_name + "/test/multipart-upload.txt"
test_file2 = test_bucket_name + "/test/multipart-upload-copy.txt"

with s3.open(test_file1, "wb", block_size=5 * 2 ** 21) as stream:
    for _ in range(5):
        stream.write(b"b" * (stream.blocksize + random.randrange(200)))
```
In the code above `test_file1` is created in 5 different parts which all are in different sizes. In that case, the etag looks something like this `b3da0a2caaab0a4e4d81b91f8e80762d-5`. Though if we copy this over (via managed copy) it will copy the whole thing in one operation due it uses a static block size that is bigger than the total size of the file and this would result with an etag that looks like this `96a4c244831bd2b4898f8b014d9c128a-1`. 

DVC needs this use case temporarily (until we revisit our internals). The implementation works simply by determining the block size on the fly by matching each copied part's block size with the part size on the source blob. 

I think we can extend the `copy()` function with a flag like `preserve_etag: bool = False` and then have this functionality independently so that no behavior changes for normal use cases.

isidentical:
Resolves #440

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -1311,6 +1311,49 @@ class S3FileSystem(AsyncFileSystem):
             raise ValueError("Copy failed (%r -> %r): %s" % (path1, path2, e)) from e
         self.invalidate_cache(path2)
 
+    async def _copy_etag_preserved(self, path1, path2, size, total_parts, **kwargs):
+        """Copy file between locations on S3 as multi-part while preserving
+        the etag (using the same part sizes for each part"""
+
+        bucket1, key1, version1 = self.split_path(path1)
+        bucket2, key2, version2 = self.split_path(path2)
+
+        mpu = await self._call_s3(
+            self.s3.create_multipart_upload, Bucket=bucket2, Key=key2, **kwargs
+        )
+
+        parts = []
+        brange_first = 0
+        for i in range(1, total_parts + 1):
+            part_info = await self._call_s3(
+                self.s3.head_object, Bucket=bucket1, Key=key1, PartNumber=i
+            )
+            part_size = part_info["ContentLength"]
+            brange_last = brange_first + part_size - 1
+            if brange_last > size:
+                brange_last = size - 1
+
+            part = await self._call_s3(
+                self.s3.upload_part_copy,
+                Bucket=bucket2,
+                Key=key2,
+                PartNumber=i,
+                UploadId=mpu["UploadId"],
+                CopySource=path1,
+                CopySourceRange="bytes=%i-%i" % (brange_first, brange_last),
+            )
+            parts.append({"PartNumber": i, "ETag": part["CopyPartResult"]["ETag"]})
+            brange_first += part_size
+
+        await self._call_s3(
+            self.s3.complete_multipart_upload,
+            Bucket=bucket2,
+            Key=key2,
+            UploadId=mpu["UploadId"],
+            MultipartUpload={"Parts": parts},
+        )
+        self.invalidate_cache(path2)
+
     async def _copy_managed(self, path1, path2, size, block=5 * 2 ** 30, **kwargs):
         """Copy file between locations on S3 as multi-part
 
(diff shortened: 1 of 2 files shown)


s3fs/core.py
```diff
@@ -1351,14 +1394,23 @@ async def _copy_managed(self, path1, path2, size, block=5 * 2 ** 30, **kwargs):
         )
         self.invalidate_cache(path2)
 
-    async def _cp_file(self, path1, path2, **kwargs):
-        gb5 = 5 * 2 ** 30
+    async def _cp_file(self, path1, path2, preserve_etag=None, **kwargs):
+        threshold = kwargs.pop("managed_threshold", 5 * 2 ** 30)
```
isidentical, line 1398: This is mostly for testing, since I didn't want to create a giant 5GB file on the memory and write it somewhere on the disk.

martindurant: It feels like this should be a module-level constant, so you can easily change it with a pytest fixture.

s3fs/core.py
```diff
@@ -1311,6 +1311,49 @@ async def _copy_basic(self, path1, path2, **kwargs):
...
+        for i in range(1, total_parts + 1):
+            part_info = await self._call_s3(
+                self.s3.head_object, Bucket=bucket1, Key=key1, PartNumber=i
+            )
+            part_size = part_info["ContentLength"]
+            brange_last = brange_first + part_size - 1
+            if brange_last > size:
+                brange_last = size - 1
+
+            part = await self._call_s3(
```
martindurant, line 1336: Can these be done concurrently with `asyncio.gather`?

s3fs/core.py
```diff
@@ -1351,14 +1394,23 @@ async def _copy_managed(self, path1, path2, size, block=5 * 2 ** 30, **kwargs):
...
         path1 = self._strip_protocol(path1)
         bucket, key, vers = self.split_path(path1)
-        size = (await self._info(path1, bucket, key, version_id=vers))["size"]
-        if size <= gb5:
+
+        info = await self._info(path1, bucket, key, version_id=vers)
+        size = info["size"]
+        if size <= threshold:
             # simple copy allowed for <5GB
             await self._copy_basic(path1, path2, **kwargs)
```
martindurant, line 1406: Does basic preserve the ETag?

s3fs/core.py
```diff
@@ -1351,14 +1394,23 @@ async def _copy_managed(self, path1, path2, size, block=5 * 2 ** 30, **kwargs):
...
+        elif preserve_etag:
+            # etag preserving multipart copy
+            _, _, parts_suffix = info["ETag"].strip('"').partition("-")
```
martindurant, line 1409: It's new to me. Is there a URL explaining this, and do you think it might change in the future?

s3fs/core.py
```diff
@@ -1311,6 +1311,49 @@ async def _copy_basic(self, path1, path2, **kwargs):
...
+
+        mpu = await self._call_s3(
+            self.s3.create_multipart_upload, Bucket=bucket2, Key=key2, **kwargs
+        )
+
+        parts = []
+        brange_first = 0
+        for i in range(1, total_parts + 1):
+            part_info = await self._call_s3(
+                self.s3.head_object, Bucket=bucket1, Key=key1, PartNumber=i
```
martindurant, line 1329: +1

martindurant (commented):
Thanks, @isidentical . I have just a few minor questions.

s3fs/core.py
isidentical: > Does basic preserve the ETag?

For basic it is just the md5, which is always the same.

s3fs/core.py
isidentical: It seems like we can initially create futures for `head_object` calls and then `asyncio.gather` them and then another round for creating the upload calls and then gathering. So it seems doable

s3fs/core.py
isidentical: A high level overview can be found here; https://teppen.io/2018/06/23/aws_s3_etags/.

s3fs/core.py
isidentical: We start from range(1, ...) so I assume it is not needed. I would have used +1 if we used `i` only once though we use it 2 times (one for head_object and one for upload part)

martindurant: I mean that as in "I like it" :)

s3fs/core.py
martindurant: Up to you on this one, please let me know if you want to stick with the original version here or if you think you can make `gather` work.

isidentical: I'll give it a shot later today, and let you know!

isidentical: It seems like using asyncio.gather on upload isn't a good option
https://github.com/dask/s3fs/blob/1333a8d46d38fd8d4c85536972cafcba998fd62e/s3fs/core.py#L1327-L1328
so I just did it for head object calls

martindurant: OK, that's fine. My guess would be that this happens because AWS doesn't know which chunk will be last. In any case, each call is quick, since the transfer payload is small.

martindurant:
Can you please document this behaviour in `_cp_file`, with example usage (which would actually go through `cp` perhaps).

PR #443, opened 2021-03-17
 Prevent recursion on error translation 

### LINKED ISSUE #442: Exception translation logic creates recursion in some cases
`raise translate_boto_error(captured_error) from captured_error` is a common pattern in the codebase, though the problem with this is that `translate_boto_error` might return the same exception and it would create a recursion of `__cause__` since `x_exception.__cause__` is itself (`is x_exception`) which would prevent anyone from traversing it easily;
https://github.com/dask/s3fs/blob/1333a8d46d38fd8d4c85536972cafcba998fd62e/s3fs/core.py#L253

This pattern can be replaced via 
```py
new_exc = translate_boto_error(exc)
if new_exc is exc:
     raise exc
else:
    raise new_exc from exc
```
but since it would create a lot of code duplication, I think a better way would be directly using `raise translate_boto_error(exc)` directly and setting the `__cause__` there to the new exception so that we can avoid the recursion and also the code duplication at the same time.

isidentical:
Resolves #442

### DIFF
--- a/s3fs/tests/test_s3fs.py
+++ b/s3fs/tests/test_s3fs.py
@@ -862,24 +862,26 @@ def test_errors(s3):
     with pytest.raises(ValueError):
         s3.find("s3://")
 
+
 def test_errors_cause_preservings(monkeypatch, s3):
     # We translate the error, and preserve the original one
     with pytest.raises(FileNotFoundError) as exc:
         s3.rm("unknown")
-    
-    assert type(exc.value.__cause__).__name__ == 'NoSuchBucket'
 
-    
+    assert type(exc.value.__cause__).__name__ == "NoSuchBucket"
+
     async def head_object(*args, **kwargs):
         raise NoCredentialsError
-    monkeypatch.setattr(type(s3.s3), 'head_object', head_object) 
-    
+
+    monkeypatch.setattr(type(s3.s3), "head_object", head_object)
+
     # Since the error is not translate, the __cause__ would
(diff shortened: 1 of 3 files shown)


s3fs/tests/test_s3fs.py
```diff
@@ -863,6 +863,26 @@ def test_errors(s3):
...
+
+    async def head_object(*args, **kwargs):
+        raise NoCredentialsError
+
+    monkeypatch.setattr(type(s3.s3), "head_object", head_object)
+
+    # Since the error is not translate, the __cause__ would
+    # be None
+    with pytest.raises(NoCredentialsError) as exc:
+        s3.info("test/a.txt")
```
martindurant, line 881: Does this case print the "During handling of the above exception, another exception occurred:" message that generally worries users? Perhaps we should set `ex.__suppress_context__ = True` in every case.

isidentical: > Does this case print the "During handling of the above exception, another exception occurred:" message that generally worries users?

Though in this case (NoCredentialsFound), the exception won't be in the context of anything will it? Considering that it is the only exception. For the rest (like the ones s3fs translates), setting `__cause__` would implicitly set `__suppress_context__` true.

martindurant:
OK, I think this is an improvement. It certainly gives us a single point to improve messaging in the future, should we wish to do so.
Thanks, @isidentical

PR #456, opened 2021-04-05
Rename _connect to set_session

### LINKED ISSUE #454: Implement S3FileSystem.set_session
For consistency with other `fspec` async implementations, could you implement `S3FileSystem.set_session`, returning an object with an async `close` method?
https://filesystem-spec.readthedocs.io/en/latest/async.html#using-from-async

Currently, `_connect` performs the function of `set_session`. However, its return value and the need to call `await close()` on it isn't documented.
https://s3fs.readthedocs.io/en/latest/#async

andreasg123:
Fixes #454 

This renames `_connect` to `set_session` and provides an implementation of `_connect` for backward compatibility. The documentation are tests are updated, too.

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -217,7 +217,7 @@ class S3FileSystem(AsyncFileSystem):
     def s3(self):
         if self._s3 is None:
             if self.asynchronous:
-                raise RuntimeError("please await ``._connect`` before anything else")
+                raise RuntimeError("please await ``.set_session`` before anything else")
             self.connect()
         return self._s3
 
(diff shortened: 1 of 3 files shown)


s3fs/core.py
```diff
@@ -375,7 +377,10 @@ async def _connect(self, refresh=False, kwargs={}):
         self._kwargs_helper = ParamKwargsHelper(self._s3)
         return self._s3
 
-    connect = sync_wrapper(_connect)
+    async def _connect(self, refresh=False, kwargs={}):
+        return await self.set_session(refresh=refresh, **kwargs)
```
martindurant, line 381: is `_connect = set_session` good enough here? Either way, a comment saying that we may deprecate this alias in the future seems appropriate.

martindurant (commented):
Looks good. I have the one comment on the code.

Do you think this needs specific explanation in the s3fs docs (as opposed to just at fsspec)?

andreasg123:
> Do you think this needs specific explanation in the s3fs docs (as opposed to just at fsspec)?

I like the documentation of the async usage in the s3fs docs because it's difficult to find good documentation for that use case.

PR #466, opened 2021-04-14
Docs: Change env variables to upper case

jrieke:
Per the [`boto3` docs](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html#environment-variables), the environment variables for credentials need to be uppercase. If I set lowercase env variables on my machine (OSX), I get the a `NoCredentialsError: Unable to locate credentials`.

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -146,7 +146,7 @@ A more secure way, not including the credentials directly in code, is to allow
 boto to establish the credentials automatically. Boto will try the following
 methods, in order:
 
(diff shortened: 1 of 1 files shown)


martindurant:
Fair enough. Do you think the code should look for both cases?

jrieke:
I guess this is handled by boto3 directly, not by s3fs, right? Anyway, I would personally only look for uppercase to keep it consistent with the behavior of boto3.

martindurant:
Fair enough. s3fs does look at some variables directly, but you are right that we should be good if it's uppercase everywhere.

PR #467, opened 2021-04-19
fix import error

johnhoman:
`fsspec.utils.setup_logging` doesn't appear to be used anywhere in s3fs.core (it's redefined almost immediately after the import). 

`fsspec==0.9.0` has the function `fsspec.utils.setup_logging` but earlier versions do not (it was renamed from setup_logger->setup_logging in `0.9.0`). The tests for `main` seem to be passing on `fsspec>=0.9.0` but the install requirements have `fsspec>=0.8.0`. It seems like the options are either bump the fsspec version to 0.9.0 or remove the unused import. 

I'm happy to make any necessary changes.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
 aiobotocore>=1.0.1
-fsspec>=0.8.0
+fsspec>=0.9.0
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -8,7 +8,7 @@ from typing import Tuple, Optional
 import weakref
 
 from fsspec.spec import AbstractBufferedFile
-from fsspec.utils import infer_storage_options, tokenize, setup_logging
+from fsspec.utils import infer_storage_options, tokenize, setup_logging as setup_logger
 from fsspec.asyn import AsyncFileSystem, sync, sync_wrapper
 
 import aiobotocore
@@ -20,21 +20,18 @@ from botocore.exceptions import ClientError, ParamValidationError
 from s3fs.errors import translate_boto_error
 from s3fs.utils import ParamKwargsHelper, _get_brange, FileExpired
 
+
 logger = logging.getLogger("s3fs")
 
 
 def setup_logging(level=None):
-    handle = logging.StreamHandler()
-    formatter = logging.Formatter(
-        "%(asctime)s - %(name)s - %(levelname)s " "- %(message)s"
-    )
-    handle.setFormatter(formatter)
-    logger.addHandler(handle)
-    logger.setLevel(level or os.environ["S3FS_LOGGING_LEVEL"])
+
+    setup_logger(logger=logger, level=(level or os.environ["S3FS_LOGGING_LEVEL"]))
 
 
 if "S3FS_LOGGING_LEVEL" in os.environ:
     setup_logging()
+
 
 MANAGED_COPY_THRESHOLD = 5 * 2 ** 30
 S3_RETRYABLE_ERRORS = (socket.timeout,)


martindurant:
It would be better to remove the s3fs version (which was the original) now that it's upstream.

johnhoman:
> It would be better to remove the s3fs version (which was the original) now that it's upstream.

@martindurant I'm not really sure I understand what you're asking, although now that I'm rereading this I'm realizing that I probably didn't explain myself well enough. 

The current [requirements.txt](https://github.com/dask/s3fs/blob/eaca2f6efb88e193eca2f3f2cac2d2e4489048db/requirements.txt#L2) has fsspec>=0.8.0, which produces this error on main
```
➜  s3fs git:(main) mkvirtualenv s3fs
created virtual environment CPython3.8.2.final.0-64 in 124ms
  creator CPython3macOsFramework(dest=/Users/jhoman/.virtualenvs/s3fs, clear=False, no_vcs_ignore=False, global=False)
  seeder FromAppData(download=False, pip=bundle, setuptools=bundle, wheel=bundle, via=copy, app_data_dir=/Users/jhoman/Library/Application Support/virtualenv)
    added seed packages: pip==21.0.1, setuptools==52.0.0, wheel==0.36.2
  activators BashActivator,CShellActivator,FishActivator,PowerShellActivator,PythonActivator,XonshActivator
virtualenvwrapper.user_scripts creating /Users/jhoman/.virtualenvs/s3fs/bin/predeactivate
virtualenvwrapper.user_scripts creating /Users/jhoman/.virtualenvs/s3fs/bin/postdeactivate
virtualenvwrapper.user_scripts creating /Users/jhoman/.virtualenvs/s3fs/bin/preactivate
virtualenvwrapper.user_scripts creating /Users/jhoman/.virtualenvs/s3fs/bin/postactivate
virtualenvwrapper.user_scripts creating /Users/jhoman/.virtualenvs/s3fs/bin/get_env_details
(s3fs) ➜  s3fs git:(main) python -m pip -q install fsspec==0.8.0
(s3fs) ➜  s3fs git:(main) python -m pip install -q .
(s3fs) ➜  s3fs git:(main) python -c "import s3fs.core"
  File "<string>", line 1, in <module>
  File "/Users/jhoman/github.com/johnhoman/s3fs/s3fs/__init__.py", line 1, in <module>
    from .core import S3FileSystem, S3File
  File "/Users/jhoman/github.com/johnhoman/s3fs/s3fs/core.py", line 11, in <module>
    from fsspec.utils import infer_storage_options, tokenize, setup_logging
ImportError: cannot import name 'setup_logging' from 'fsspec.utils' (/Users/jhoman/.virtualenvs/s3fs/lib/python3.8/site-packages/fsspec/utils.py)
(s3fs) ➜  s3fs git:(main)
```

removing setup_logging will fix the `ImportError` or moving the requirement for fsspec to 0.9.0 but since the import isn't used anyway it makes sense to just remove it IMO. 

I don't think that either of those things are the thing that you're saying though so I apologize for my misunderstand but would you mind clarifying so I can make the necessary changes.

martindurant:
In the release to come, the required version of fsspec will be updated (actually to 2021.04.0), so that you cannot get the old version without `setup_logging`. It is fine to update it here, but of course this has no effect until release. If you install s3fs in a new environment, you will already get fsspec 0.9.0 (failures are only coming from selective updates). 

Your current version keeps both the fsspec version as I had intended and calls it with the old semantics, so I am happy with this.

martindurant:
and thank you

PR #468, opened 2021-04-20
Updates for release

### LINKED ISSUE #465: order versions appear broken due to recent fsspec changes
@martindurant I see that `fsspec.asyn` module got rid of `maybe_sync` function, but older versions of `s3fs` still rely on the method which now fails to import. For example `s3fs==0.5.2` now does not seem to work after a clean install. I see that s3fs requirements.txt asked for `fsspec>=0.8.0` and that is now incompatible without an upper boundary.

martindurant:
Closes #465

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -4,7 +4,7 @@ Changelog
 Version 2021.04.0
 -----------------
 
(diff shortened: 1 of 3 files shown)


martindurant:
cc @zklaus and any others

Issue #470, opened 2021-04-28
exists() on non-existent files might take up to 6 seconds


isidentical:
In the following example, the first `exists()` call takes about ~5.4 seconds and the next one takes ~3 seconds. This is probably because when a key doesn't exist we first call `head_object` and then proceed with the `list_objects` to see whether if it is a directory. If neither of those checks out, we end up with waiting 6 seconds.

```py
import time
import s3fs

path = "existing-bucket/something"

fs = s3fs.S3FileSystem()
fs.rm(path)

t0 = time.perf_counter()
fs.exists(path)
t1 = time.perf_counter()
print("exists() on a non-existent file took: ", t1 - t0, "seconds")

fs.touch(path)

t0 = time.perf_counter()
fs.exists(path)
t1 = time.perf_counter()
print("exists() on an existing file took: ", t1 - t0, "seconds")

fs.rm(path)
```

martindurant:
the time this takes will, of course, depend on your connection and proximity to the bucket. 
At least we do cache the listings, so that subsequent calls should be faster.

Do you see any other foolproof way to find out if a file or "directory" exists? It is worth noting that not all users who can access a given key also have permissions to list the bucket.

isidentical:
>  At least we do cache the listings, so that subsequent calls should be faster.

Though this is not a listing, this is a single retrieval so we don't cache it (same goes for single info calls, isdir / isfile calls etc). 

> Do you see any other foolproof way to find out if a file or "directory" exists?

I'm still trying to figure it out a faster way, and hopefully send something (:eyes:)

martindurant:
> this is a single retrieval so we don't cache it

If we list the path itself or the parent, those should get cached and avoid the direct call. The trouble is to tell the difference between an empty listing and a failed one.

Note that this should be part of the larger discussion on how to handle object store listings (e.g., what to do with paths ending in the "/" delimiter, and so on).

isidentical:
> If we list the path itself or the parent, those should get cached and avoid the direct call. The trouble is to tell the difference between an empty listing and a failed one.

Yes, I'm aware of that. But if you have a prefix with millions of keys under it, then listing the parent wouldn't be much efficient. So I was thinking from the point of view where someone tries to call `exists()` repeatedly with the same path. 

> Note that this should be part of the larger discussion on how to handle object store listings (e.g., what to do with paths ending in the "/" delimiter, and so on).

[...]

PR #472, opened 2021-04-29
s3: optimize info/exists calls

isidentical:
Described in https://github.com/dask/s3fs/issues/470#issuecomment-829097599

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -923,6 +923,9 @@ class S3FileSystem(AsyncFileSystem):
             raise FileNotFoundError(path)
     
     async def _version_aware_info(self, path):
+        path = self._strip_protocol(path)
+        bucket, key, _ = self.split_path(path)
+
         out = await self._call_s3(
             "head_object",
             self.kwargs,
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -967,8 +967,12 @@ class S3FileSystem(AsyncFileSystem):
                 if self.version_aware:
                     out = await self._version_aware_info(path, version_id)
                 else:
-                    out = await self._simple_info(path)
-
+                    try:
+                        out = await self._simple_info(path)
+                    except PermissionError:
+                        # If the permissions aren't enough for scanning a prefix
+                        # then fall back to using normal HEAD_OBJECT
+                        out = await self._version_aware_info(path, version_id)
                 if out:
                     return out
             except ClientError as e:
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -921,6 +921,28 @@ class S3FileSystem(AsyncFileSystem):
             or out.get("CommonPrefixes", [])
         ):
             raise FileNotFoundError(path)
+    
+    async def _version_aware_info(self, path):
+        out = await self._call_s3(
+            "head_object",
+            self.kwargs,
+            Bucket=bucket,
+            Key=key,
+            **version_id_kw(version_id),
+            **self.req_kw,
+        )
+        return {
+            "ETag": out["ETag"],
+            "Key": "/".join([bucket, key]),
+            "LastModified": out["LastModified"],
+            "Size": out["ContentLength"],
+            "size": out["ContentLength"],
+            "name": "/".join([bucket, key]),
+            "type": "file",
+            "StorageClass": "STANDARD",
+            "VersionId": out.get("VersionId"),
+        }
+
 
     async def _info(self, path, bucket=None, key=None, refresh=False, version_id=None):
         path = self._strip_protocol(path)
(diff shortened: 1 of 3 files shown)


martindurant:
Please ping me to have a thorough review of this.

First a couple of thoughts:
- loosing version information could be a dealbraker for some, so we would only want to do this when the filesystem is not marked as version-aware
- for the case where the user does not have list permission, we would still want to fall back to HEAD (but not bother trying to list for the directory)

isidentical:
> (but not bother trying to list for the directory)

AFAIK permissions might allow searching for prefixes that ends with trailing slash, so it kinda makes sense to bother to list directory since that time we'll call it with a trailing slash.

martindurant:
There is a separate permission for listing files with a prefix versus listing with a prefix that happens to end with "/"? That would be annoying!

martindurant:
This might be a good place to include a test against https://github.com/dask/s3fs/issues/473 and fix, if necessary.

isidentical:
> There is a separate permission for listing files with a prefix versus listing with a prefix that happens to end with "/"?

I think we hit that before, though @efiop probably has more context than me. It is not a specific permission per se, but rather a bucket-level rule declaration that allows you to list `/bucket/something/*` but not `/bucket/somethingXXX` etc.

isidentical:
> This might be a good place to include a test against #473 and fix, if necessary.

That seems a very different bug, though since it is trivial a fix is included in this PR.

efiop:
> > There is a separate permission for listing files with a prefix versus listing with a prefix that happens to end with "/"?
> 
> I think we hit that before, though @efiop probably has more context than me. It is not a specific permission per se, but rather a bucket-level rule declaration that allows you to list `/bucket/something/*` but not `/bucket/somethingXXX` etc.

Correct, here is a ticket for it https://github.com/iterative/dvc/issues/5130

martindurant:
Can you please also add the test shown in https://github.com/dask/s3fs/issues/475 here? I suspect the changes in this PR will have fixed that.

isidentical:
> Can you please also add the test shown in #475 here? I suspect the changes in this PR will have fixed that.

If I am not missing anything, I already did https://github.com/dask/s3fs/pull/472/commits/0eb3620b788775c1a4badec6f2861dad5e53a748

s3fs/core.py
```diff
@@ -888,6 +889,62 @@ async def _get_file(self, rpath, lpath, version_id=None):
...
+                "LastModified": file["LastModified"],
+                "Size": file["Size"],
+                "size": file["Size"],
+                "name": "/".join([bucket, key]),
+                "type": "file",
+                "StorageClass": "STANDARD",
+                "VersionId": file.get("VersionId"),
+            }
+        elif not (file or out.get("KeyCount", 0) or out.get("CommonPrefixes", [])):
+            raise FileNotFoundError(path)
```
martindurant, line 919: Aren't there `else` cases here:
- we got a file by a different name, so the original file was not found
- we got prefix(es), so the path is a directory (does MaxKeys restrict the number of prefixes? Why use delimiter in the call at all?)
- we got a path which is the same except for an extra "/"

What happens if we fail all the ifs, and return `None`?

martindurant:
Specifically the test
```python
@pytest.mark.parametrize("find", [False, True])
def test_lsdir(s3, find):
    if find:
        s3.find(test_bucket_name)

    d = test_bucket_name + "/test"
    assert d in s3.ls(test_bucket_name)
```

isidentical:
> This might be a good place to include a test against #473 and fix, if necessary.
> Can you please also add the test shown in #475 here? I suspect the changes in this PR will have fixed that.

wait you are talking about different issues?

isidentical:
this PR doesn't interact with ls/find in any way, so I don't think there is any reason to include a fix for #475.

martindurant:
OK, that will have to be dealt with separately.

s3fs/core.py
isidentical: > What happens if we fail all the ifs, and return None?

The `else: ` is the condition where the prefix doesn't exist as is, and there are other prefixes. For example for a directory like this;
```
very/
   a12/
      another
   a1234
```

`very/a123` will be handled in the first condition, `very/bsomething` will be handled in the second condition since no such prefix exists either as a file or a directory and `very/a1` and `very/a2` will fallback to the directory search which would return true for `very/a2` and raise file not found error for `very/a`.

I couldn't find a test case which isn't handled by this logic

isidentical: No, else is a totally different path. For example there might other prefixes matched, so it might be a directory etc

s3fs/core.py
```diff
@@ -908,27 +965,17 @@ async def _info(self, path, bucket=None, key=None, refresh=False, version_id=Non
...
+                    out = await self._version_aware_info(path, version_id)
+                else:
+                    try:
+                        out = await self._simple_info(path)
+                    except PermissionError:
+                        # If the permissions aren't enough for scanning a prefix
+                        # then fall back to using normal HEAD_OBJECT
+                        out = await self._version_aware_info(path, version_id)
+                if out:
+                    return out
```
martindurant, line 978: So, is `out` ever `None` or empty here, or would that have raised by now? This is the end of the function anyway, so the `if` seems redundant. It would be nice to specify what this method can be expected to return - indeed, I should have done this.

martindurant (commented):
> No, else is a totally different path.

You mean there are three branches here: the return statement above (key found), this branch (we know the file doesn't exist) and a third implicit branch where we return `None`?

s3fs/core.py
isidentical: It can be None indeed, if it fall backs to the third (implicit) branch.

martindurant: OK, got it. Thanks for clarifying - for some reason I found it hard to follow that. That all sounds good, maybe you want to add a comment in the code - or not.

isidentical:
Okay, I'll add a few comments

martindurant:
(I'll merge later - going out prospecting for a vaccine)

isidentical:
it should be fine

martindurant:
Done! Thanks for this.

PR #480, opened 2021-05-12
copy: drop requirement on etag

### LINKED ISSUE #477: KeyError: 'ETag'
**What happened**:
```
E       KeyError: 'ETag'
```

files get copied

```python
from s3fs import S3FileSystem
s3 = S3FileSystem()
```

**Environment**:

- s3fs version: 2021.4.0
- Python version: 3.6.10
- Operating System: Mac

isidentical:
resolves #477

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -1518,7 +1518,7 @@ class S3FileSystem(AsyncFileSystem):
         info = await self._info(path1, bucket, key, version_id=vers)
         size = info["size"]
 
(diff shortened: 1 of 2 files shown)


martindurant:
@machielg , perhaps you can provide a form of your workflow as a test function, so we can ensure we are fixing your precise problem?

martindurant:
I'll merge this anyway tomorrow, if no further test case is forthcoming.

PR #483, opened 2021-05-20
allow None etag, and do not set IfMatch as a result.

### LINKED ISSUE #477: KeyError: 'ETag'

gwen-at-nielsen:
This PR addresses the issue elaborated on here: https://github.com/dask/s3fs/issues/477, which was causing s3fs to fail when no ETag was available for an s3 object.

Let me know if you guys want a specific unit test for this -- it's such a minor change that I'm not sure whether it warrants one.

Fixes #477

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -1752,8 +1752,9 @@ class S3File(AbstractBufferedFile):
                 self.append_block = True
             self.loc = loc
 
-        if "r" in mode:
-            self.req_kw["IfMatch"] = self.details["ETag"]
+        e_tag = self.details.get("ETag")
+        if "r" in mode and e_tag is not None:
+            self.req_kw["IfMatch"] = e_tag
 
     def _call_s3(self, method, *kwarglist, **kwargs):
         return self.fs.call_s3(method, self.s3_additional_kwargs, *kwarglist, **kwargs)


martindurant:
Thank for contributing! This solves the specific issue you were experiencing? 

Can you please check: does this ETag condition get used anywhere else?

It may be reasonable to test this by modifying a directory listing in-place to a) remove the ETag (moto will always provide one, I think) or b) modify the ETag, where the latter should cause a failure. Please

gwen-at-nielsen:
Hello! I think we determined why the file that was trying to be read didn't have an ETag -- it's actually a folder, not an individual parquet file, and it had been previously written by a spark session through databricks. In a previous version of Pandas s3 folders were not differentiated from files when they are being read from s3. As a result, we were asking for an ETag that genuinely wasn't there, but I'm not sure the right resolution here is to silently ignore not having an ETag.

We'll get our databricks admin to update pandas.

zijie0:
We also met this issue using s3fs to read parquet files in a folder. The code is:

```
pf = ParquetFile(path, open_with=s3.open)
```

I'm not sure if it is an expected behavior. If so, what is the correct way to read parquet files in a folder?

martindurant:
Sorry, @zijie0 , what exactly is your issue? Are you using a version following #480 ?

zijie0:
Hi @martindurant , we are using the latest version of s3fs == 2021.6.1.
The issue is that we are trying to read parquet files in an s3 directory, for example:
s3://parquet_folder containing part-0000.parquet, part-0001.parquet and so on.

If we use the code:

```
path = 's3://parquet_folder'
pf = ParquetFile(path, open_with=s3.open)
```

It will throw an exception of missing key 'ETag'.

Currently the workaroud is using `s3.glob`:

```
path = 's3://parquet_folder'
file_list = s3.glob(f'{path}/*.parquet')
pf = ParquetFile(file_list , open_with=s3.open)
```

martindurant:
@zijie0 , can you post your traceback? This PR should have solved the problem. Are you able to access files on s3 otherwise?

zijie0:
Hi @martindurant , I found that this issue is gone after I update fastparquet version.
I believe that your previous pull request fix the problem: https://github.com/dask/fastparquet/pull/603

PR #486, opened 2021-06-01
find: support arbitrary prefixes

### LINKED ISSUE #485: s3fs: supporting find() on arbitrary prefixes (like glob('a/b/prefix*')) with better performance
When listing only files that start with a certain prefix `foo/bar/b*` `glob()` is handy though in terms of s3 (and also other clouds) it is not as performant as just doing this on the server-side with not adding the final `/` suffix. Especially when the parent directories size is greater than a single listing call (1024). Here is an example;

```py
import time
import secrets
from s3fs import S3FileSystem
from pprint import pprint
from concurrent.futures import ThreadPoolExecutor, as_completed

bucket = 'temp/cache/00/'

def worker(fs, bucket, prefix=''):
    fs.touch(bucket + prefix + secrets.token_hex())

def make_files():
    fs = S3FileSystem()
    with ThreadPoolExecutor(max_workers=8) as executor:
        futures = [
            executor.submit(worker, fs, bucket)
            for _ in range(9900)
        ]
        futures += [
            executor.submit(worker, fs, bucket, '0')
            for _ in range(100)
        ]
        
        for n, future in enumerate(as_completed(futures), 1):
            future.result()
            print(f'{n}/10000 completed')
```

it creates 10_000 files where 100 of them are guaranteed to start with `0` prefix and 9900 random files that can start with anything. In the following case if I do a `glob('temp/cache/00/0*')` it takes about 22 seconds compared to a single find call with a new option (`use_prefix: bool = True`) it takes about 3~ seconds `find('temp/cache/00/0', use_prefix=True)`.

isidentical:
Resolves #485


s3fs/core.py
```diff
@@ -509,9 +509,15 @@ def _open(
...
 
-    async def _lsdir(self, path, refresh=False, max_items=None, delimiter="/"):
+    async def _lsdir(
+        self, path, refresh=False, max_items=None, delimiter="/", use_prefix=True
+    ):
         bucket, prefix, _ = self.split_path(path)
-        prefix = prefix + "/" if prefix else ""
+        if prefix:
+            if use_prefix:
+                prefix += "/"
```
martindurant, line 518: The name of the argument "use_prefix" seems to have the opposite meaning to its function - at least to how it seems to me. The current default, which matches previous behaviour, is the right one, though.

isidentical: Yeah, I don't quite like it as well. That was the first to come to my mind (use_prefix => use the forward slash in the `Prefix=...`).  Any suggestions?

s3fs/core.py
```diff
@@ -559,10 +562,28 @@ async def _lsdir(self, path, refresh=False, max_items=None, delimiter="/"):
...
+            Only return files that match ``^{path}/{prefix}`` (if there is an
+            exact match ``filename == {path}/{prefix}``, it also will be included)
+        """
+
         bucket, key, _ = self.split_path(path)
         if not bucket:
             raise ValueError("Cannot traverse all of S3")
+        if (withdirs or maxdepth) and prefix:
+            # TODO: perhaps propagate these to a glob(f"path/{prefix}*") call
+            raise ValueError("Can not specify prefix whil")
```
martindurant, line 586: message truncated

martindurant (commented):
Good to go. Any chance you can update that error message in the next hour or so before releas?

s3fs/core.py
isidentical:
```suggestion
            raise ValueError("Can not specify 'prefix' option alongside 'withdirs'/'maxdepth' options.")
```

Issue #423, opened 2021-02-17
Test against minio


martindurant:
Currently we run tests against moto in a separate process.
Maybe we should also test against minio, which is a fully-functioning S3 implementation with some of its own quirks (lacks versioning?). I would propose this as a separate CI run.

https://docs.min.io/docs/minio-docker-quickstart-guide.html

PR #495, opened 2021-06-10
spec: implement bucket region caching

### LINKED ISSUE #494: Speed Up exists()/info() calls up to 3X when no region is specified
After a report on aiobotocore, it has been noticed that the connections `aiobotocore` creates are actually gets terminated after each call due to S3's abnormal behaviour regarding resolving regions. This does not affect users who specify regions, but in our use case and I assume in others specifying regions in the client constructor is a bit restrictive since you might need to do copies/transfers across buckets and creating a new fs instance for that is not just practical. So I propose that we introduce a thin wrapper around the clients, and cache the bucket regions to be as efficient as possible. It would be only activated if user specify's the option (`cache_regions`) and what it does is that, every place you access `self.s3` you'd go through the path where your bucket will be reviewed and if there is a specific client for that bucket, it will be used. If not, it will first try to resolve the region via a `HEAD_BUCKET` call, and if it succeeds it will create a new client. If not, it will return the generic one. 

Script;
```py
import time
import s3fs

path = "temp/batuhan"
fs = s3fs.S3FileSystem(skip_instance_cache=True)

def create_dirs(paths):
    for path in paths:
        fs.touch(path)

paths = [
    path + "/file_1",
    path + "/file_2",
    path + "/file_3",
    path + "/file_4",
    path + "/other/file_5",
    path + "/other/file_6",
    path + "/other/file_7",
    path + "/other/something/file_8",
    path + "/other/something/file_9",
    path + "/file_10",
]

# create_dirs(paths)

# Add the dirs
paths.append(path + "/other/")
paths.append(path + "/other/something/")

# Throw some non-existent files
paths.append(path + "/file_11")
paths.append(path + "/other/something2/")
paths.append(path + "/other/file_12")
paths.append(path + "/other/something/file_13")

fs = s3fs.S3FileSystem(
    cache_regions=True, skip_instance_cache=True
)

t0 = time.perf_counter()
for path in paths:
    fs.exists(path)
t1 = time.perf_counter()
print(f"total time spent: {t1 - t0}, total time per invocation: {(t1 - t0) / len(paths)}")
```

With cache regions:
```
total time spent: 21.582699037000566, total time per invocation: 1.3489186898125354
```

Without cache regions:
```
total time spent: 7.280172572000083, total time per invocation: 0.4550107857500052
```

isidentical:
Resolves #494

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -373,10 +373,24 @@ class S3FileSystem(AsyncFileSystem):
                 if key not in drop_keys
             }
             config_kwargs["signature_version"] = UNSIGNED
+
         conf = AioConfig(**config_kwargs)
         self.session = aiobotocore.AioSession(**self.kwargs)
 
-        if self.cache_regions:
+        for parameters in (config_kwargs, self.kwargs, init_kwargs, client_kwargs):
+            for option in ("region_name", "endpoint_url"):
+                if parameters.get(option):
+                    self.cache_regions = False
+                    break
+        else:
+            cache_regions = self.cache_regions
+
+        logger.debug(
+            "RC: caching enabled? %r (explicit option is %r)",
+            cache_regions,
+            self.cache_regions,
+        )
+        if cache_regions:
             s3creator = S3BucketRegionCache(
                 self.session, config=conf, **init_kwargs, **client_kwargs
             )
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -1687,6 +1687,20 @@ class S3FileSystem(AsyncFileSystem):
     def sign(self, path, expiration=100, **kwargs):
         return self.url(path, expires=expiration, **kwargs)
 
+    async def _invalidate_region_cache(self):
+        if not self.cache_regions:
+            raise ValueError(
+                "This method is only available when cache_regions is turned on"
+            )
+
+        # If the region cache is not initialized, then
+        # do nothing.
+        cache = getattr(self, "_s3creator", None)
+        if cache is not None:
+            await cache.clear()
+
+    invalidate_region_cache = sync_wrapper(_invalidate_region_cache)
+
 
 class S3File(AbstractBufferedFile):
     """
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -1689,9 +1689,7 @@ class S3FileSystem(AsyncFileSystem):
 
     async def _invalidate_region_cache(self):
         if not self.cache_regions:
-            raise ValueError(
-                "This method is only available when cache_regions is turned on"
-            )
+            return None
 
         # If the region cache is not initialized, then
         # do nothing.
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -18,7 +18,7 @@ from aiobotocore.config import AioConfig
 from botocore.exceptions import ClientError, ParamValidationError
 
 from s3fs.errors import translate_boto_error
-from s3fs.utils import ParamKwargsHelper, _get_brange, FileExpired
+from s3fs.utils import S3BucketRegionCache, ParamKwargsHelper, _get_brange, FileExpired
 
 
 logger = logging.getLogger("s3fs")
@@ -125,6 +125,10 @@ class S3FileSystem(AsyncFileSystem):
         Whether to support bucket versioning.  If enable this will require the
         user to have the necessary IAM permissions for dealing with versioned
         objects.
+    cache_regions : bool (False)
+        Whether to cache bucket regions or not. Whenever a new bucket is used,
+        it will first find out which region it belongs and then use the client
+        for that region.
     config_kwargs : dict of parameters passed to ``botocore.client.Config``
     kwargs : other parameters for core session
     session : aiobotocore AioSession object to be used for all connections.
@@ -173,6 +177,7 @@ class S3FileSystem(AsyncFileSystem):
         session=None,
         username=None,
         password=None,
+        cache_regions=False,
         asynchronous=False,
         loop=None,
         **kwargs
@@ -207,6 +212,7 @@ class S3FileSystem(AsyncFileSystem):
         self.req_kw = {"RequestPayer": "requester"} if requester_pays else {}
         self.s3_additional_kwargs = s3_additional_kwargs or {}
         self.use_ssl = use_ssl
+        self.cache_regions = cache_regions
         self._s3 = None
         self.session = None
 
@@ -221,9 +227,16 @@ class S3FileSystem(AsyncFileSystem):
     def _filter_kwargs(self, s3_method, kwargs):
         return self._kwargs_helper.filter_dict(s3_method.__name__, kwargs)
 
+    async def get_s3(self, bucket=None):
+        if self.cache_regions and bucket is not None:
+            return await self._s3creator.get_bucket_client(bucket)
+        else:
+            return self._s3
+
     async def _call_s3(self, method, *akwarglist, **kwargs):
         await self.set_session()
-        method = getattr(self.s3, method)
+        s3 = await self.get_s3(kwargs.get("Bucket"))
+        method = getattr(s3, method)
         kw2 = kwargs.copy()
         kw2.pop("Body", None)
         logger.debug("CALL: %s - %s - %s", method.__name__, akwarglist, kw2)
@@ -362,11 +375,19 @@ class S3FileSystem(AsyncFileSystem):
             config_kwargs["signature_version"] = UNSIGNED
         conf = AioConfig(**config_kwargs)
         self.session = aiobotocore.AioSession(**self.kwargs)
-        s3creator = self.session.create_client(
-            "s3", config=conf, **init_kwargs, **client_kwargs
-        )
+
+        if self.cache_regions:
+            s3creator = S3BucketRegionCache(
+                self.session, config=conf, **init_kwargs, **client_kwargs
+            )
+            self._s3 = await s3creator.get_client()
+        else:
+            s3creator = self.session.create_client(
+                "s3", config=conf, **init_kwargs, **client_kwargs
+            )
+            self._s3 = await s3creator.__aenter__()
+
         self._s3creator = s3creator
-        self._s3 = await s3creator.__aenter__()
         # the following actually closes the aiohttp connection; use of privates
         # might break in the future, would cause exception at gc time
         if not self.asynchronous:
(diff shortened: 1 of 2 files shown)


s3fs/core.py
```diff
@@ -384,7 +405,7 @@ def close_session(loop, s3):
             try:
                 sync(loop, s3.__aexit__, None, None, None, timeout=0.1)
                 return
-            except TimeoutError:
+            except FSTimeoutError:
```
martindurant, line 408: This change, and the import for it, are already in main https://github.com/dask/s3fs/blob/main/s3fs/core.py#L387 , so please merge. This maybe got merged after you posted the PR.

isidentical: Right. Just saw that your PR is merged, so I'll do a rebase. Thanks

isidentical:
I don't really have tests for this at the moment, but I will try to come up with something (at worst just mocking, but I'd rather use some sort of real stuff, like moto).

martindurant:
I see the clear evidence from the linked issue that caching can significantly speed up calls, but can you summarise, please, why this is? My understanding is, that currently botocore will look up the region for the bucket of a given call every time; but once we know the region for some bucket, we should be able to reuse it and avoid that call.  It is surprising (to me) that this could be 3x slower, but maybe it depends on where the region of the bucket is versus the s3 lookup endpoint.

Does this require a separate client for each region (or each bucket?), or could the region be specified at call time. Maybe that's actually what you do. 

The case against making this default or the only path would be that HEAD_BUCKET can fail. Is the bucket region not available via an initial region-less call response's metadata?

martindurant:
I don't think moto handles regions and likely neither does minio - not that we use the latter in tests (yet #423 )

isidentical:
> I see the clear evidence from the linked issue that caching can significantly speed up calls, but can you summarise, please, why this is?

The most basic explanation is that, every time a region is unknown for `HEAD_OBJECT` calls boto fails with 400 first and then goes on the route of resolving these. And when a connection gets `400`, the underlying aiohttp connection gets terminated so for every call you are now paying the cost of the entire transport creation.

>  It is surprising (to me) that this could be 3x slower, but maybe it depends on where the region of the bucket is versus the s3 lookup endpoint

If you check the numbers of the linked issue, there is a huge gap which was netted multiple testers. And when the overall connection is getting faster, the gap gets higher. 

> Does this require a separate client for each region (or each bucket?), or could the region be specified at call time. Maybe that's actually what you do.

That was the first thing that I wanted to do, since client creation costs some time (0.05 seconds or something) also the experience is not great. But from what I can see, you have create a new client for each region you want to support. We do cache those clients, so if you have 6 buckets from 2 different regions, you only create 3 different clients. A generic one (to be used in `HEAD_BUCKET` calls), and 2 clients for 2 different regions. 

> The case against making this default or the only path would be that HEAD_BUCKET can fail. Is the bucket region not available via an initial region-less call response's metadata?

I'd rather separate these. I'll look into whether we can retrieve the region from an initial call, though even so I've experienced some problems on DVC side with pinning the region/signature version, some users' settings might be broken etc. Though if we could retrieve the metadata from an existing call, then we could perhaps make it just a bit more faster for the initial case. Need to look into it.

isidentical:
> I don't think moto handles regions and likely neither does minio - not that we use the latter in tests (yet #423 )

Hmmm.

martindurant:
> Though if we could retrieve the metadata from an existing call, then we could perhaps make it just a bit more faster for the initial case. 

Even if not, it should make for simpler code. When debugging some future issues, it would be great not to have to ask whether region client caching was on, and whether, to their knowledge, the HEAD_BUCKET had succeeded.

By the way, is HEAD_BUCKET typically allowed for publicly accessible buckets?

isidentical:
> By the way, is HEAD_BUCKET typically allowed for publicly accessible buckets?

AFAIK, yes, ACL `public-read`/`public-read-write` both support `HEAD_BUCKET`.

isidentical:
> Even if not, it should make for simpler code. When debugging some future issues, it would be great not to have to ask whether region client caching was on, and whether, to their knowledge, the HEAD_BUCKET had succeeded.

After checking a bit, i don't think it is possible to defer this calls. At least from what I an see, we should make them initially. I'll add a bit of logging to help for better debugging and also a guard around `HEAD_BUCKET` to ensure that we return the generic (region unbound) client if the call fails.

isidentical:
One idea that might make this even faster is that, permanent caching. It might sound a bit odd at the first glance, but if you think that all bucket names are globally unique, and their regions are constant (you can't change a bucket's region without actually deleting the bucket and re-creating it with the same name which is really really unlikely so I don't even see a need for that, but perhaps it might be handled in an EAFP way, like try with the known region from the permanent cache on the system and if that fails go back and resolve it again and invalidate the current cache, which would be a bit less performant for initial call on a really really rare case), this optimization really does make sense. Especially for short running processes (e.g CLI applications) which use s3fs. If they are making only a couple of info calls, then they might shave %50 of their runtime by simply enabling this option.

martindurant:
I would be fine with persisting bucket regions and invalidating on error (this should indeed be rare). Now you have to worry about writing and reading a file!

By the way, I am not advertising 2x speed (or 3x) with this improvement, but reduced latency; for cases where the data volumes are large, users may not notice the difference.

isidentical:
> Now you have to worry about writing and reading a file!

Not really though, this is something totally optional. If the region cache is open, and if we fail to read/write to a file then it is perfectly fine. We can still have the cache in-memory by constructing on demand. 

> By the way, I am not advertising 2x speed (or 3x) with this improvement, but reduced latency; for cases where the data volumes are large, users may not notice the difference.

Right. As I stressed in the original issue, this is something that mainly happens on HeadObject (not even on GetObject), so if you make 2 3 HeadObject calls and upload idk 5 GiB data, then of course it is unlikely that you will notice the difference.

s3fs/core.py
```diff
@@ -1663,6 +1687,20 @@ def modified(self, path, version_id=None, refresh=False):
     def sign(self, path, expiration=100, **kwargs):
         return self.url(path, expires=expiration, **kwargs)
 
+    async def _invalidate_region_cache(self):
+        if not self.cache_regions:
+            raise ValueError(
+                "This method is only available when cache_regions is turned on"
```
martindurant, line 1693: I think it's fine to call invalidate when there is no cache

s3fs/utils.py
```diff
@@ -10,6 +15,100 @@ def ignoring(*exceptions):
...
+    def __init__(self, session, **client_kwargs):
+        self._session = session
+        self._stack = AsyncExitStack()
+        self._client = None
+        self._client_kwargs = client_kwargs
+        self._buckets = {}
+        self._regions = {}
+
+    async def get_bucket_client(self, bucket_name=None):
+        # TODO: raise error when client_kwargs has region
```
martindurant, line 57: This one is important!
If you specify a region and bucket is in a different region, what happens?

martindurant (commented):
As far as I can tell, a call to `list_objects_v2` also gives the bucket region with the key `'x-amz-bucket-region'`, so maybe we can avoid calls to HEAD_BUCKET for cases when `ls` (or related) is the first call on a bucket?

s3fs/utils.py
isidentical: We should decide to either ignore or raise an error when one of these options are specified: `region`, `endpoint_url` and `use_accelerated_endpoint` (or something like that, don't recall now).

martindurant: I would assume that if the user specifies, we should trust that they know what they are doing. This is already possible now, so we would be breaking someone's workflow otherwise.

isidentical: > This is already possible now, so we would be breaking someone's workflow otherwise.

I'm asking for the case where user both specify `cache_regions` and `region` at the same time. Should we just ignore `cache_regions`, or raise like `ValueError("You can't use both region and cache_regions options")`. I'd say that we could simply ignore `cache_regions`, but curious what you'd think?

martindurant: OK, I see what you mean. I would normally say that this is a ValueError for kwargs that don't make sense together. However, I expect caching to  become default one release later, so I think we can imply cache_regions=False so long as the documentation says so.

In the implementation before this PR, what happens if you specify a region and then access a bucket in another region?

s3fs/utils.py
```diff
@@ -10,6 +15,99 @@ def ignoring(*exceptions):
...
+
+        general_client = await self.get_client()
+        if bucket_name is None:
+            return general_client
+
+        try:
+            response = await general_client.head_bucket(Bucket=bucket_name)
+        except ClientError:
+            logger.debug(
+                "RC: HEAD_BUCKET clal for %r has failed, returning the general client",
```
martindurant, line 68: typo

s3fs/core.py
```diff
@@ -1663,6 +1701,18 @@ def modified(self, path, version_id=None, refresh=False):
     def sign(self, path, expiration=100, **kwargs):
         return self.url(path, expires=expiration, **kwargs)
 
+    async def _invalidate_region_cache(self):
+        if not self.cache_regions:
```
martindurant, line 1705: Please add a docstring for this

s3fs/core.py
```diff
@@ -360,13 +373,35 @@ async def set_session(self, refresh=False, kwargs={}):
                 if key not in drop_keys
             }
             config_kwargs["signature_version"] = UNSIGNED
+
         conf = AioConfig(**config_kwargs)
         self.session = aiobotocore.AioSession(**self.kwargs)
-        s3creator = self.session.create_client(
-            "s3", config=conf, **init_kwargs, **client_kwargs
+
+        for parameters in (config_kwargs, self.kwargs, init_kwargs, client_kwargs):
```
martindurant, line 380: Can the region/endpoint really be in any one of these kwarg sets?

martindurant (commented):
OK, I am persuaded. I have a couple of small comments, but let's get this merged before release.

s3fs/core.py
isidentical: from what I can see, all of them can contain `region_name`. And `self.kwargs` can contain `endpoint_url` so I just included it in this logic (since if I flatten all of these `not (config_kwargs.get("xxx") or config_kwargs.get("yyy") or self.kwargs.get("xxx") or ...)` they become really complicated).

martindurant: OK, I suppose these are all typically small dicts, so this won't take long.

isidentical: Also note that we are iterating on container objects themselves, not their contents. So at most we iterate over 4 objects and perform 2 lookups which is the same with 8X dict.get() calls, which is extremely cheap

martindurant:
OK, it's in!
This seems like the kind of thing you might well want to write an article about, since you potentially just increased the efficiency of some workloads by a factor of 3!

PR #497, opened 2021-06-30
cast all non-truthy prefixes to an empty string

isidentical:
Now it supports `.find(prefix=None)`


martindurant:
(please run `black`)

PR #499, opened 2021-07-13
spec: implement rm_file() for single file deletion

isidentical:


martindurant:
This is going in, thanks!

I wonder, when do you find it necessary to call rm_file rather than rm (which would call _bulk delete with one path) ?

isidentical:
> I wonder, when do you find it necessary to call rm_file rather than rm (which would call _bulk delete with one path) ?

DVC always uses `rm_file` whether deleting a single file or multiple files, and for some remotes it is more optimized than `rm()` since it is doing a single job and not trying to expand path / check whether it is a directory etc. And s3fs was the only one which didn't support `rm_file` among the stuff we migrated so far so needed that. But yes, bulk deleting is a potential optimization target for me to use when we do `dvc gc`, but later ofc

PR #502, opened 2021-07-22
Fix mkdir so that no create_bucket is called when one already exists

asclepiusaka:


s3fs/core.py
```diff
@@ -685,7 +685,9 @@ async def _find(self, path, maxdepth=None, withdirs=None, detail=False, prefix="
     async def _mkdir(self, path, acl="", create_parents=True, **kwargs):
         path = self._strip_protocol(path).rstrip("/")
         bucket, key, _ = self.split_path(path)
-        if not key or (create_parents and not await self._exists(bucket)):
+        if await self._exists(bucket):
+            await self._ls(bucket)
```
martindurant, line 689: Why list here?

s3fs/tests/test_s3fs.py
```diff
@@ -575,6 +575,16 @@ def test_mkdir(s3):
     assert bucket in s3.ls("/")
 
 
+def test_mkdir_existing_bucket(s3):
+    # mkdir called on existing bucket should be no-op and not calling create_bucket
+    # creating a s3 bucket
+    bucket = "test1_bucket"
+    s3.mkdir(bucket)
+    # a second call.
+    s3.mkdir(bucket)
```
martindurant, line 584: I would have thought this should raise FileExistsError, no?
The other related methods to check:
- s3.makedirs(bucket, exist_ok=True) # no error
- s3.makedirs(bucket, exist_ok=False) # FileExists error
- s3.makedirs(bucket + "/any/sub/path") # no error
- s3.mkdir(bucket + "/any/other/path", create_parents=True)  # no error

martindurant (commented):
Thanks for contributing!
I have left a couple of comments to try to make sure that we get he behaviour that's appropriate.

s3fs/core.py
asclepiusaka: my bad. I mean to leave it to do nothing when bucket already exist, will update.

s3fs/tests/test_s3fs.py
asclepiusaka: you are right... not sure what i was thinking, let me submit another pr, another question though,
if we have 
`
s3.makedirs(existing_bucket + "/path/to/existing/object", exist_ok=False)
`
The current code is no going to raise any error, is this behavior by design too?

asclepiusaka: done, can you review again?

s3fs/tests/test_s3fs.py
```diff
@@ -866,8 +914,9 @@ def test_errors(s3):
         f.close()
         f.read()
 
-    with pytest.raises(ValueError):
-        s3.mkdir("/")
+    # disabling this for now, mkdir("/") can be a non-op as creating bucket
+    # with pytest.raises(ValueError):
```
martindurant, line 918: Everything els looks good here, but I wonder about this one. Whilst I don't expect anyone to try to create "/", and we should no longer be accidentally doing it ourselves, it should still be an error, right? 
I am prepared to let it go as unimportant.

asclepiusaka: You are right, i don't see any reason anyone would call this. Let me actually fix this.

martindurant: Can you please uncomment this part, so that your value error actually gets called?

PR #510, opened 2021-08-20
Aio session is no longer top level

gadomski:
Since https://github.com/aio-libs/aiobotocore/releases/tag/1.4.0, `AioSession` no longer exists in the top-level namespace. This PR also:

- pins the **aiobotocore** dependency more strictly, since the project demonstrated that it will do backwards-incompatible changes with a minor bump, and
- pins the `extras_require` dependencies; they were previously unpinned.

My local test runs are failing but I _don't_ think its due to this change (they were failing on **main**) so apologies if CI fails.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,6 +3,14 @@
 from setuptools import setup
 import versioneer
 
+with open('requirements.txt') as file:
+    aiobotocore_version_suffix = ''
+    for line in file:
+        parts = line.rstrip().split('aiobotocore')
+        if len(parts) == 2:
+            aiobotocore_version_suffix = parts[1]
+            break
+
 setup(name='s3fs',
       version=versioneer.get_version(),
       cmdclass=versioneer.get_cmdclass(),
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -25,8 +25,8 @@
       packages=['s3fs'],
       python_requires='>= 3.6',
       install_requires=[open('requirements.txt').read().strip().split('\n')],
-      extras_require = {
-          'awscli': ['aiobotocore[awscli]'],
-          'boto3': ['aiobotocore[boto3]'],
+      extras_require={
+          'awscli': ['aiobotocore[awscli] ~= 1.4.0'],
```
martindurant, line 29: Can this be generated from the version in requirements.txt?

gadomski: Sure, I can give that a shot.

martindurant:
Thanks for the fix, all appears to be well.

martindurant:
Fixes #511

PR #517, opened 2021-08-21
made glob raise exception on *

drorspei:
Hej hej!

Currently the expected behavior of `glob("*")` is to throw an exception.

But this behavior currently relies on private implementation details of `fsspec` which are orthogonal to `s3fs`.

This pull request removes this implicit dependency, so that both projects can continue to develop in parallel.

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -602,6 +602,11 @@ class S3FileSystem(AsyncFileSystem):
             return files
         return self.dircache[path]
 
+    async def _glob(self, path, **kwargs):
+        if path.startswith("*"):
+            raise ValueError("Cannot traverse all of S3")
+        return await super()._glob(path, **kwargs)
+
     async def _find(self, path, maxdepth=None, withdirs=None, detail=False, prefix=""):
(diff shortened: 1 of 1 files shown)


martindurant:
> But this behavior currently relies on private implementation details of fsspec which are orthogonal to s3fs.

I think the two are pretty tightly coupled!

I think your original request was to *not* raise in this situation, but use the results of `ls("")`, which would be fine. The proposed change here doesn't make any difference to behaviour, as you say, just codifies the current situation.

martindurant:
(tests will pass if you merge from main)

drorspei:
I thought you wanted to keep the exception :P

If the exception is to be kept, then anyway it is codified by the test that checks it, and then I think that making the raise independent of the fsspec code is the right choice (even if the packages are also tightly coupled in other places.)

If instead we allow matching on bucket in glob, then the test that checks glob("*") raises has to go.

Personally I don't have an opinion on what is really better, and I'm a new contributor so I'll go with whatever you think is best. I can put up the pull request either way ^_^

PR #530, opened 2021-09-10
Fix range header in cat_file

martindurant:


martindurant:
Actually, the tests show that the original was correct, and the `end` parameter should be read as the usual pythonic non-inclusive delimiter. i.e., the calling application needs to add one, if they intend the Nth byte to be included.

PR #533, opened 2021-09-17
Call get_bucket_location before calling list_objects_v2

### LINKED ISSUE #532: Disallowing ListObjectsV2 at the root of the bucket makes s3fs attempt to create a bucket
**What happened**:

It's not uncommon to have a bucket that disallows listing the root, but allows listing a specific prefix. In this case `s3fs` will fail any writes and will attempt to create the bucket, which often fails with a completely different error. 

Falling back to creating a bucket is very strange behaviour. I imagine it's legacy and impossible to change, but I would expect that `s3fs` does not require full list objects permissions over the bucket to perform any writes.

```python
In [1]: import s3fs

In [2]: s3 = s3fs.S3FileSystem(anon=False)

In [5]: s3.mkdirs("s3://s3fs-test-bucket-123/foo/bar")
2021-09-17 17:34:50,222 - s3fs - DEBUG - _call_s3 -- CALL: list_objects_v2 - () - {'MaxKeys': 1, 'Bucket': 's3fs-test-bucket-123'}
2021-09-17 17:34:50,516 - s3fs - DEBUG - _call_s3 -- Nonretryable error: An error occurred (AccessDenied) when calling the ListObjectsV2 operation: Access Denied
2021-09-17 17:34:50,516 - s3fs - DEBUG - _call_s3 -- CALL: create_bucket - () - {'Bucket': 's3fs-test-bucket-123', 'ACL': ''}
2021-09-17 17:34:50,576 - s3fs - DEBUG - _call_s3 -- Nonretryable error: An error occurred (IllegalLocationConstraintException) when calling the CreateBucket operation: The unspecified location constraint is incompatible for the region specific endpoint this request was sent to.
```

The full traceback is like so:

```
File "/home/app/.cache/pypoetry/virtualenvs/x/lib/python3.9/site-packages/dask/dataframe/io/parquet/arrow.py", line 819, in initialize_write
    fs.mkdirs(path, exist_ok=True)
  File "/home/app/.cache/pypoetry/virtualenvs/x/lib/python3.9/site-packages/fsspec/spec.py", line 1159, in mkdirs
    return self.makedirs(path, exist_ok=exist_ok)
  File "/home/app/.cache/pypoetry/virtualenvs/x/lib/python3.9/site-packages/fsspec/asyn.py", line 88, in wrapper
    return sync(self.loop, func, *args, **kwargs)
  File "/home/app/.cache/pypoetry/virtualenvs/x/lib/python3.9/site-packages/fsspec/asyn.py", line 69, in sync
    raise result[0]
  File "/home/app/.cache/pypoetry/virtualenvs/x/lib/python3.9/site-packages/fsspec/asyn.py", line 25, in _runner
    result[0] = await coro
  File "/home/app/.cache/pypoetry/virtualenvs/x/lib/python3.9/site-packages/s3fs/core.py", line 731, in _makedirs
    await self._mkdir(path, create_parents=True)
  File "/home/app/.cache/pypoetry/virtualenvs/x/lib/python3.9/site-packages/s3fs/core.py", line 716, in _mkdir
    await self._call_s3("create_bucket", **params)
```

It seems like it's failing to detect [the bucket exists on this line](https://github.com/dask/s3fs/blob/9e7ea3784cffc4cbec05281e2a689ae4c3ff8517/s3fs/core.py#L833-L836). There are much better methods to detect if a bucket exists, like `get-bucket-location`.

orf:
Closes #532

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -832,14 +832,14 @@ class S3FileSystem(AsyncFileSystem):
                 pass
             try:
                 await self._call_s3(
-                    "get_bucket_location", Bucket=bucket, **self.req_kw
+                    "list_objects_v2", MaxKeys=1, Bucket=bucket, **self.req_kw
                 )
                 return True
             except Exception:
                 pass
             try:
                 await self._call_s3(
-                    "list_objects_v2", MaxKeys=1, Bucket=bucket, **self.req_kw
+                    "get_bucket_location", Bucket=bucket, **self.req_kw
                 )
                 return True
             except Exception:
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -832,6 +832,13 @@ class S3FileSystem(AsyncFileSystem):
                 pass
             try:
                 await self._call_s3(
+                    "get_bucket_location", Bucket=bucket, **self.req_kw
+                )
+                return True
+            except Exception:
+                pass
+            try:
+                await self._call_s3(
                     "list_objects_v2", MaxKeys=1, Bucket=bucket, **self.req_kw
                 )
                 return True
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -838,9 +838,7 @@ class S3FileSystem(AsyncFileSystem):
             except Exception:
                 pass
             try:
-                await self._call_s3(
-                    "get_bucket_location", Bucket=bucket, **self.req_kw
-                )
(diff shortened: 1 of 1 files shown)


martindurant:
Will merge when green, thanks.

orf:
All green! Thanks for the quick response here 💪

isidentical:
@orf do you also want to make a PR about your `MaxKeys=0` suggestion, I think that also makes sense!

PR #545, opened 2021-11-03
Bugfix/issue 544 drop unnecessary list objects v2

ashnyukov:


isidentical (approved):
Tests could stay (beside the permission one), but other than that LGTM. I have not yet found time to run the benchmarks locally, but will do to confirm the behavior. Thanks!

ashnyukov:
@isidentical 
about the tests -- agree + reverted
about the benchmarks -- please double check (don't want to undo your work for no reason)

martindurant:
@isidentical , up to you.
Do you we still want listv2 as a fallback? Is it possible that HEAD is forbidden but LIST allowed?

isidentical:
> about the benchmarks -- please double check (don't want to undo your work for no reason)


Regarding this, unfortunately, I do not have access to my AWS account anymore so can't benchmark. @ashnyukov I see that you already tried the aiobotocore case I have provided, can you also try this with/without your patch? If the results are not changed much, then it is good to go for me.

```py
import time
import s3fs

path = "existing-bucket/something"

fs = s3fs.S3FileSystem()
fs.rm(path)

t0 = time.perf_counter()
fs.exists(path)
t1 = time.perf_counter()
print("exists() on a non-existent file took: ", t1 - t0, "seconds")

fs.touch(path)

t0 = time.perf_counter()
fs.exists(path)
t1 = time.perf_counter()
print("exists() on an existing file took: ", t1 - t0, "seconds")

fs.rm(path)
```

> Do you we still want listv2 as a fallback? Is it possible that HEAD is forbidden but LIST allowed?

No, this was a simple workaround so I don't see any cases where this might be expected/depended upon.

martindurant:
I found very exactly the same performance for existing file, and variable performance for non-existent file with the patch and new aiohttp Vs pre-patch and older aiohttp.

PR #550, opened 2021-11-09
Pass through cache_options to S3File

rjzamora:
Downstream users and libraries may need to pass `cache_options` (along with `cache_type`) to the back-end cache when an `S3File` instance is created. This is especially true for the`"parts"` cache (`KnownPartsOfAFile`) and the new `fsspec.parquet.open_parquet_file` optimization that uses it.

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -481,6 +481,7 @@ class S3FileSystem(AsyncFileSystem):
         cache_type=None,
         autocommit=True,
         requester_pays=None,
+        cache_options=None,
         **kwargs
     ):
         """Open a file for reading or writing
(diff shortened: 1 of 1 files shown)


martindurant:
Fails on something completely unrelated? I'm happy to push it in for the same of dask's release, but someone should dig out why that test started to fail. Perhaps some other package updated.

rjzamora:
>Fails on something completely unrelated? I'm happy to push it in for the same of dask's release, but someone should dig out why that test started to fail. Perhaps some other package updated.

Interesting!  I will try to do a bit of digging today/tomorrow.

rjzamora:
@martindurant - Since the same tests are failing on `main`, I vote that the fix not be a blocker for getting this merged (I'd like to point people to `main` rather than a PR to use "parts" caching with s3).

martindurant:
The tests are failing for you on main? I don't think it's shown up anywhere else yet.

rjzamora:
> The tests are failing for you on main?

I believe so - Just submitted #551 as a simple test

martindurant:
OK, so finding why it fails onto the long todo list...

PR #552, opened 2021-11-16
Don't cache metadata

martindurant:
Attempted fix/debug for CI failures


martindurant:
@isidentical - given that the metadata cache is completely independent of the general file info cache, I think we should not store both, and eat the cost of repeated lookups during set/getxattr.

isidentical:
Agreed. I don't think `set`/`getxattr`/`metadata` is that common anyway, so it should be fine.

isidentical (approved):

PR #556, opened 2021-11-18
Mark ClientPayloadError as a retryable exception

orf:
Occasionally `aiohttp.ClientPayloadError` can be thrown when reading a response from S3, specifically:

```
aiohttp.client_exceptions.ClientPayloadError: Response payload is not completed
``` 

This isn't caught by `s3fs` even though it's definitely a retryable error - it indicates a transient network issue that caused the client to fail to read the complete response. This PR conditionally adds this to the list of retryable exceptions.


martindurant:
Good find

maximsch2:
Is this enough of a fix? I'm hitting ClientPayloadError that's being thrown off here: https://github.com/fsspec/s3fs/blob/main/s3fs/core.py#L916 which is outside of a retry loop.

PR #557, opened 2021-11-18
Fix #477 -- Don't set IfMatch if ETag is missing

### LINKED ISSUE #477: KeyError: 'ETag'

dargueta:
Fixes #477.


martindurant:
Looks good, thanks.

PR #562, opened 2021-12-12
Fix typos

kianmeng:


s3fs/tests/test_s3fs.py
```diff
@@ -1558,7 +1558,7 @@ def test_change_defaults_only_subsequent():
     systems created after the change.
     """
     try:
-        S3FileSystem.cachable = False  # don't reuse instances with same pars
+        S3FileSystem.cacheable = False  # don't reuse instances with same pars
```
isidentical, line 1561: @martindurant any idea whether this might be doing something that it shouldn't do? (is cacheable an fsspec attribute?)

martindurant: Oh no! Indeed this should *not* change, it is defined in AbstarctFileSystem

martindurant: https://github.com/fsspec/s3fs/pull/563

PR #572, opened 2021-12-18
Persist HTTP Header

### LINKED ISSUE #571: Persist HTTP Contents
Hello,

On appending files, it erase older content type / encoding

Br
nicolas

Platob:
Persist Headers on append
Infer Content Type/Encoding on write

Fixes #571

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -1834,10 +1834,14 @@ class S3File(AbstractBufferedFile):
                     )
         elif "w" in mode or append:
             if "ContentType" not in self.s3_additional_kwargs:
-                mimetype, _ = guess_type(self.path)
+                mimetype, encoding = guess_type(self.path)
                 if isinstance(mimetype, str):
                     self.s3_additional_kwargs["ContentType"] = mimetype
-            if "ContentEncoding" not in self.s3_additional_kwargs:
+                    
+                if "ContentEncoding" not in self.s3_additional_kwargs:
+                    if isinstance(encoding, str):
+                        self.s3_additional_kwargs["ContentEncoding"] = encoding
+            elif "ContentEncoding" not in self.s3_additional_kwargs:
                 _, encoding = guess_type(self.path)
                 if isinstance(encoding, str):
                     self.s3_additional_kwargs["ContentEncoding"] = encoding
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -937,14 +937,14 @@ class S3FileSystem(AsyncFileSystem):
 
     async def _put_file(self, lpath, rpath, chunksize=50 * 2 ** 20, **kwargs):
         bucket, key, _ = self.split_path(rpath)
-        
+
         # Infer Content Type/Encoding
         mimetype, encoding = guess_type(key)
         if isinstance(mimetype, str):
             kwargs["ContentType"] = kwargs.get("ContentType", mimetype)
         if isinstance(encoding, str):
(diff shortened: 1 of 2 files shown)


s3fs/core.py
```diff
@@ -936,6 +937,14 @@ async def _pipe_file(self, path, data, chunksize=50 * 2 ** 20, **kwargs):
 
     async def _put_file(self, lpath, rpath, chunksize=50 * 2 ** 20, **kwargs):
         bucket, key, _ = self.split_path(rpath)
+
+        # Infer Content Type/Encoding
+        mimetype, encoding = guess_type(key)
+        if isinstance(mimetype, str):
+            kwargs["ContentType"] = kwargs.get("ContentType", mimetype)
```
martindurant, line 944: I think it's a bad idea to guess this for the user, where octet-stream is the *right* type for almost any opaque remote blob. Furthermore, the guessing is often wrong :|

I don't actually understand how you can guess the encoding!

s3fs/core.py
```diff
@@ -1754,6 +1766,30 @@ class S3File(AbstractBufferedFile):
     retries = 5
     part_min = 5 * 2 ** 20
     part_max = 5 * 2 ** 30
+    # See: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/
+    # services/s3.html#S3.Client.put_object
+    additional_info_keys = [
+        "CacheControl",
```
martindurant, line 1772: I am no expert here, I am not certain it's useful to reflect all of these parameters back at the server. I certainly would be quite surprised to see them show up in an `info` call, since they are not actually information about the file for the most part.

s3fs/core.py
```diff
@@ -1811,18 +1847,39 @@ def __init__(
...
+            for key in self.additional_info_keys:
+                value = info.get(key, None)
+                if value is not None:
+                    self.s3_additional_kwargs[key] = self.s3_additional_kwargs.get(
+                        key, value
+                    )
+        elif "w" in mode or append:
+            # Infer HTTP Contents: Type, Encoding
+            if "ContentType" not in self.s3_additional_kwargs:
+                mimetype, encoding = guess_type(self.path)
```
martindurant, line 1871: Same as before, I am dubious about guessing this

s3fs/tests/test_s3fs.py
```diff
@@ -256,7 +256,8 @@ def test_info(s3):
     linfo = s3.ls(a, detail=True)[0]
     assert abs(info.pop("LastModified") - linfo.pop("LastModified")).seconds < 1
     info.pop("VersionId")
-    info.pop("ContentType")
+    for key in S3File.additional_info_keys:
+        info.pop(key, None)
```
martindurant, line 260: I guess at this point, you really should test that you can a) set the MIME of an upload, b) append to that upload c) recover the original MIME. 
What happens if in b) you specify a different MIME? Does it depend on whether the file is smaller or larger than one binary block?

martindurant:
I support the spirit of this PR, but I would limit it to the append case, to persist the type/encoding headers as found in the original. That is a nicely simple addition, which can be easily tested. I would remove the rest, perhaps open a new issue specifically about the possible usage of `mimetypes`.

isidentical:
Agree with @martindurant, this seems like a very huge PR with a lot of changes in the core but no tests at all. It might be simpler to change the focus on just adding the headers, and making the other improvements in separate fully-tested PRs.

Platob:
Ok thank you

PR #574, opened 2021-12-27
fixes: acl not set for small files

kaktus42:
```
fs = s3fs.S3FileSystem()
with fs.open("s3://my-bucket/my/file", "w", acl="bucket-owner-full-control") as f:
  f.write("test")
```

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -1979,6 +1979,7 @@ class S3File(AbstractBufferedFile):
                     Key=self.key,
                     Bucket=self.bucket,
                     Body=data,
+                    ACL=self.acl,
                     **self.kwargs,
                 )
             else:
--- a/s3fs/tests/test_s3fs.py
+++ b/s3fs/tests/test_s3fs.py
@@ -1045,7 +1045,7 @@ def test_write_small_with_acl(s3):
             "Type": "Group",
         },
     }
-    
+
     with s3.open(filename, "wb", acl="public-read") as f:
         f.write(body)
     assert s3.cat(filename) == body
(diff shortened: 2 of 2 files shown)


martindurant:
Thanks for the one-liner!
Perhaps you could add a test showing that the uploaded file did indeed get the requested ACL?

kaktus42:
I also made a test for large files, but it fails - although when testing it manually, setting acl works.
Maybe someone else can have a look:

```python
def test_write_large_with_acl(s3):
    bucket, key = (test_bucket_name, "test-acl")
    filename = bucket + "/" + key
    public_read_acl = {
        "Permission": "READ",
        "Grantee": {
            "URI": "http://acs.amazonaws.com/groups/global/AllUsers",
            "Type": "Group",
        },
    }
    
    mb = 2 ** 20
    payload_size = int(2.5 * 5 * mb)
    payload = b"0" * payload_size

    with s3.open(filename, "wb", acl="public-read") as f:
        f.write(payload)

    assert s3.cat(filename) == payload
    assert s3.info(filename)["Size"] == payload_size

    print(sync(s3.loop, s3.s3.get_object_acl, Bucket=bucket, Key=key))

    assert (
        public_read_acl
        in sync(s3.loop, s3.s3.get_object_acl, Bucket=bucket, Key=key)["Grants"]
    )
```

but this works:
```python
fs = s3fs.S3FileSystem(profile="saml")

mb = 2 ** 20
payload_size = int(2.5 * 5 * mb)
payload = b"0" * payload_size

with fs.open("s3://mybucket/my/path/test", "wb", acl="bucket-owner-full-control") as f:
  f.write(payload)
```

martindurant:
Let's leave the large file case for a follow-up, since you fix and test for small files here.

martindurant:
Please run `black` to appease the linter

martindurant:
Thank you @kaktus42

PR #575, opened 2022-01-02
persist metadata on append

### LINKED ISSUE #571: Persist HTTP Contents

Platob:
Fixes #571

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -1042,7 +1042,7 @@ class S3FileSystem(AsyncFileSystem):
                     "size": out["ContentLength"],
                     "name": "/".join([bucket, key]),
                     "type": "file",
-                    "StorageClass": "STANDARD",
+                    "StorageClass": out.get("StorageClass", "STANDARD"),
                     "VersionId": out.get("VersionId"),
                     "ContentType": out.get("ContentType"),
                 }
@@ -1814,13 +1814,33 @@ class S3File(AbstractBufferedFile):
         self.append_block = False
 
         if "a" in mode and s3.exists(path):
-            loc = s3.info(path)["size"]
+            head = {
+                k: v for k, v in self._call_s3(
+                    "head_object", self.kwargs, Bucket=bucket, Key=key,
+                    **version_id_kw(version_id), **self.req_kw
+                ).items() if k in {
+                    "CacheControl", "ContentDisposition", "ContentEncoding",
+                    "ContentLanguage", "ContentLength", "ContentMD5",
+                    "ContentType", "Expires", "WebsiteRedirectLocation",
+                    "ServerSideEncryption",
+                    "SSECustomerAlgorithm", "SSEKMSKeyId",
+                    "BucketKeyEnabled", "StorageClass",
+                    "ObjectLockMode", "ObjectLockRetainUntilDate",
+                    "ObjectLockLegalHoldStatus",
+                    "Metadata"
+                }
+            }
+
+            loc = head.pop("ContentLength")
             if loc < 5 * 2 ** 20:
                 # existing file too small for multi-upload: download
                 self.write(self.fs.cat(self.path))
             else:
                 self.append_block = True
             self.loc = loc
+
+            # Reflect head
+            self.s3_additional_kwargs.update(head)
 
         if "r" in mode and "ETag" in self.details:
             self.req_kw["IfMatch"] = self.details["ETag"]
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -1815,19 +1815,34 @@ class S3File(AbstractBufferedFile):
 
         if "a" in mode and s3.exists(path):
             head = {
-                k: v for k, v in self._call_s3(
-                    "head_object", self.kwargs, Bucket=bucket, Key=key,
-                    **version_id_kw(version_id), **self.req_kw
-                ).items() if k in {
-                    "CacheControl", "ContentDisposition", "ContentEncoding",
-                    "ContentLanguage", "ContentLength",
-                    "ContentType", "Expires", "WebsiteRedirectLocation",
+                k: v
+                for k, v in self._call_s3(
+                    "head_object",
+                    self.kwargs,
+                    Bucket=bucket,
+                    Key=key,
+                    **version_id_kw(version_id),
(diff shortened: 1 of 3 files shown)


s3fs/core.py
```diff
@@ -1042,7 +1042,7 @@ async def _info(self, path, bucket=None, key=None, refresh=False, version_id=Non
                     "size": out["ContentLength"],
                     "name": "/".join([bucket, key]),
                     "type": "file",
-                    "StorageClass": "STANDARD",
+                    "StorageClass": out.get("StorageClass", "STANDARD"),
```
martindurant, line 1045: Is this ever missing?

s3fs/core.py
```diff
@@ -1814,14 +1814,49 @@ def __init__(
         self.append_block = False
 
         if "a" in mode and s3.exists(path):
-            loc = s3.info(path)["size"]
+            head = {
+                k: v
+                for k, v in self._call_s3(
```
martindurant, line 1819: The previous version allowed the info to come from cache, but here we explicitly repeat. This might be OK, or maybe we should include the header keys that matter in the normal info, so we can reuse them.

s3fs/core.py
```diff
@@ -1814,14 +1814,49 @@ def __init__(
...
+                    "head_object",
+                    self.kwargs,
+                    Bucket=bucket,
+                    Key=key,
+                    **version_id_kw(version_id),
+                    **self.req_kw,
+                ).items()
+                if k
+                in {
+                    "CacheControl",
```
martindurant, line 1829: Do you please have a source for this list of keys? I would be pretty surprised at some of them if they were meant to be passed back as part of the upload.

martindurant (commented):
Thanks for the pared down PR! It is much more nicely self-contained now.

I have left a few comments for you.

s3fs/core.py
Platob: I saw StorageClass could be different from standard when set on write

martindurant: OK

s3fs/core.py
Platob: It's still something I can't find best way to reflect, on one hand it's nice to preserve it in cache, but heavy to get all keys

Since it's mainly used on append, I think it might be better to reflect it one time at initialization and let user cache it or not later

s3fs/core.py
Platob: Added comments

martindurant: thank you

s3fs/core.py
```diff
@@ -1814,14 +1814,55 @@ def __init__(
...
+            loc = head.pop("ContentLength")
             if loc < 5 * 2 ** 20:
                 # existing file too small for multi-upload: download
                 self.write(self.fs.cat(self.path))
             else:
                 self.append_block = True
             self.loc = loc
 
+            # Reflect head
+            self.s3_additional_kwargs.update(head)
```
martindurant, line 1864: Do you think that the user should be able override the values from HEAD at the time of append?

Platob: I would say yes, user still have choice to let default behavior reflecting head, or updating headers based on new content pushed

martindurant: OK, then this is the wrong way around, the existing values will always win. You would need
```
head.update(self.s3_additional_kwargs)
self.s3_additional_kwargs = head
```

Would be nice to test this too.

Platob: Just excluded current additional kwargs, if it's ok for you ? Or your solution works too

martindurant: OK

martindurant:
@efiop might have thoughts on this

efiop:
We don't use appends in dvc (yet), so this doesn't directly affect us (yet). I haven't dealt with s3fs internals enough yet, but the approach looks good to me 👍

s3fs/core.py
```diff
@@ -1814,14 +1814,58 @@ def __init__(
...
+                        "WebsiteRedirectLocation",
+                        "ServerSideEncryption",
+                        "SSECustomerAlgorithm",
+                        "SSEKMSKeyId",
+                        "BucketKeyEnabled",
+                        "StorageClass",
+                        "ObjectLockMode",
+                        "ObjectLockRetainUntilDate",
+                        "ObjectLockLegalHoldStatus",
+                        "Metadata",
```
efiop, line 1852: Let's put these into some constant to simplify the logic here, as there are currently too many levels.

s3fs/core.py
```diff
@@ -1814,14 +1814,58 @@ def __init__(
...
+                    }
+                    and k not in self.s3_additional_kwargs
+                )
+            }
+
+            loc = head.pop("ContentLength")
```
efiop, line 1858: How about smth like this (untested) to reduce levels and make it a bit more readable:

```suggestion
            # See:
            # put: https://boto3.amazonaws.com/v1/documentation/api/latest
            # /reference/services/s3.html#S3.Client.put_object
            #
            # head: https://boto3.amazonaws.com/v1/documentation/api/latest
            # /reference/services/s3.html#S3.Client.head_object
            head = self._call_s3(
                "head_object",
                self.kwargs,
                Bucket=bucket,
                Key=key,
                **version_id_kw(version_id),
                **self.req_kw,
            )
            
            head = {
                key: value
                for key, value in head.items()
                if key in PRESERVE_KWARGS and key not in self.s3_additional_kwargs
            }

            loc = head.pop("ContentLength")
```

s3fs/tests/test_s3fs.py
```diff
@@ -1226,6 +1228,7 @@ def test_append(s3):
     with s3.open(test_bucket_name + "/nested/file1", "ab") as f:
         assert f.tell() == len(data)  # append, no write, small file
     assert s3.cat(test_bucket_name + "/nested/file1") == data
+
```
efiop, line 1231: Unrelated change, let's not pollute the PR, please
```suggestion
```

martindurant:
@efiop , perhaps you can commit your suggestions here so we can finish this PR off?

martindurant:
thank you!

PR #579, opened 2022-01-15
Set content type on upload

damnMeddlingKid:
This PR adds the content type to uploaded objects by populating the "ContentType" kwarg using mimetypes. I'm not sure how to test this and also unsure about if anything extra is needed for multipart, would love clarity on that.

tested with 

```
pytest s3fs/tests/test_s3fs.py -k test_content_type
```

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -945,10 +945,10 @@ class S3FileSystem(AsyncFileSystem):
                 await self._mkdir(lpath)
         size = os.path.getsize(lpath)
 
-        content_type, _ = mimetypes.guess_type(lpath)
-
-        if content_type is not None:
-            kwargs.setdefault("ContentType", content_type)
+        if "ContentType" not in kwargs:
+            content_type, _ = mimetypes.guess_type(lpath)
+            if content_type is not None:
+                kwargs["ContentType"] = content_type
 
         with open(lpath, "rb") as f0:
             if size < min(5 * 2 ** 30, 2 * chunksize):
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -945,8 +945,10 @@ class S3FileSystem(AsyncFileSystem):
                 await self._mkdir(lpath)
         size = os.path.getsize(lpath)
 
-        content_type = mimetypes.guess_type(lpath)
-        kwargs.update({"ContentType": content_type[0]})
+        content_type, _ = mimetypes.guess_type(lpath)
+
+        if content_type is not None:
+            kwargs.setdefault("ContentType", content_type)
 
         with open(lpath, "rb") as f0:
             if size < min(5 * 2 ** 30, 2 * chunksize):
--- a/s3fs/tests/test_s3fs.py
+++ b/s3fs/tests/test_s3fs.py
@@ -694,6 +694,7 @@ def test_s3_file_info(s3):
     with pytest.raises(FileNotFoundError):
         s3.info(fn + "another")
 
+
 def test_content_type_is_set(s3, tmpdir):
     test_file = str(tmpdir) + "/test.json"
     destination = test_bucket_name + "/test.json"
@@ -701,12 +702,14 @@ def test_content_type_is_set(s3, tmpdir):
     s3.put(test_file, destination)
     assert s3.info(destination)["ContentType"] == "application/json"
 
+
 def test_content_type_is_not_overrided(s3, tmpdir):
     test_file = os.path.join(str(tmpdir), "test.json")
     destination = os.path.join(test_bucket_name, "test.json")
     open(test_file, "w").write("text")
     s3.put(test_file, destination, ContentType="text/css")
     assert s3.info(destination)["ContentType"] == "text/css"
+
 
 def test_bucket_exists(s3):
     assert s3.exists(test_bucket_name)
--- a/s3fs/tests/test_s3fs.py
+++ b/s3fs/tests/test_s3fs.py
@@ -695,8 +695,8 @@ def test_s3_file_info(s3):
         s3.info(fn + "another")
 
 def test_content_type_is_set(s3, tmpdir):
-    test_file = os.path.join(str(tmpdir), "test.json")
-    destination = os.path.join(test_bucket_name, "test.json")
+    test_file = str(tmpdir) + "/test.json"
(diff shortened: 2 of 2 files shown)


s3fs/core.py
```diff
@@ -943,6 +944,10 @@ async def _put_file(self, lpath, rpath, chunksize=50 * 2 ** 20, **kwargs):
             else:
                 await self._mkdir(lpath)
         size = os.path.getsize(lpath)
+
+        content_type = mimetypes.guess_type(lpath)
+        kwargs.update({"ContentType": content_type[0]})
```
isidentical, line 949: Wouldn't this override the user provided one?

isidentical: Also `mimetypes.guess_type` can return `None`, so we will be effectively sending a null header if we can't guess. We need to do this conditionally (and through using `setdefault`).

isidentical:
Thanks for the PR @damnMeddlingKid. Testing should be easy, if the given header is preserved by the moto. One important thing is that, we should not override the user provided one. 

Guessing the mime type automatically seems like something aws-cli is doing when transferring a local file to the s3, so I'm +1 on supporting this in `put_file` as well. Not sure though if there are any performance implications, but at least on Linux it seems pretty-fast (and alongside a network request, it is nothing).

martindurant:
> Guessing the mime type automatically seems like something aws-cli is doing when transferring a local file to the s3, so I'm +1 on supporting this in put_file as well. Not sure though if there are any performance implications

I believe t is a simple lookup in a dictionary, should be negligible. We do, however, need to ass mimetypes to the deps list.

damnMeddlingKid:
> > Guessing the mime type automatically seems like something aws-cli is doing when transferring a local file to the s3, so I'm +1 on supporting this in put_file as well. Not sure though if there are any performance implications
> 
> I believe t is a simple lookup in a dictionary, should be negligible. We do, however, need to ass mimetypes to the deps list.

its part of the standard library

martindurant:
> its part of the standard library

Well ok, then!

isidentical:
AFAIK windows might be the actual problematic case. Before 3.10, the initialization of the mimetypes DB takes about ~130 ms, after it is about ~30 ms ([source](https://bugs.python.org/issue44582#msg397114)). Though I find even this negligible.

s3fs/core.py
damnMeddlingKid: good call, i've added in a check. right now im ignoring the encoding, should i add that in aswell ?.

damnMeddlingKid:
could you point me at an example for testing this change ?.

s3fs/core.py
```diff
@@ -943,6 +944,12 @@ async def _put_file(self, lpath, rpath, chunksize=50 * 2 ** 20, **kwargs):
             else:
                 await self._mkdir(lpath)
         size = os.path.getsize(lpath)
+
+        content_type, _ = mimetypes.guess_type(lpath)
+
+        if content_type is not None:
+            kwargs.setdefault("ContentType", content_type[0])
```
isidentical, line 951: `content_type` is already unpacked, so AFAIK this should be:
```suggestion
            kwargs.setdefault("ContentType", content_type)
```

isidentical:
@damnMeddlingKid you can add a test like the following which uploads a custom file (let's say json) and checks the `ContentType`. https://github.com/fsspec/s3fs/blob/dfa1cd28467541cfafb25056a0402a49bacd219a/s3fs/tests/test_s3fs.py#L687-L695

damnMeddlingKid:
Added two tests. one to check that the content type is set correctly and another to make sure that user's content type isn't overrided if it is set.

s3fs/tests/test_s3fs.py
```diff
@@ -694,6 +694,19 @@ def test_s3_file_info(s3):
     with pytest.raises(FileNotFoundError):
         s3.info(fn + "another")
 
+def test_content_type_is_set(s3, tmpdir):
+    test_file = os.path.join(str(tmpdir), "test.json")
+    destination = os.path.join(test_bucket_name, "test.json")
+    open(test_file, "w").write("text")
+    s3.put(test_file, destination)
+    assert s3.info(destination)["ContentType"] == "application/json"
+
```
martindurant, line 703: I believe the linter will complain if you don't have two lines between functions.

s3fs/tests/test_s3fs.py
```diff
@@ -694,6 +694,19 @@ def test_s3_file_info(s3):
...
+def test_content_type_is_not_overrided(s3, tmpdir):
+    test_file = os.path.join(str(tmpdir), "test.json")
+    destination = os.path.join(test_bucket_name, "test.json")
```
martindurant, line 706: This should be "/" join, because we use posix-style paths for s3 even on windows

s3fs/core.py
```diff
@@ -943,6 +944,12 @@ async def _put_file(self, lpath, rpath, chunksize=50 * 2 ** 20, **kwargs):
             else:
                 await self._mkdir(lpath)
         size = os.path.getsize(lpath)
+
+        content_type, _ = mimetypes.guess_type(lpath)
+
+        if content_type is not None:
```
martindurant, line 950: I would change the logic around so that we don't do the lookup if the user gives a value, and not to bother setting a None:
```
if "ContentType" not in kwargs:
    content_type, _ = mimetypes.guess_type(lpath)
    if content_type is not None:
        kwargs["contentType"] = content_type
```

I assume that the implicit default is octet-stream, perhaps we should mention this in a comment or docs.

martindurant:
Thanks for adding the tests, they do the job!

s3fs/core.py
damnMeddlingKid: good idea, i've added this in.

s3fs/tests/test_s3fs.py
damnMeddlingKid: switched to "/".

s3fs/core.py
martindurant: You can propose this in a separate PR, but I am thinking the encoding is even less useful than MIME, I would need to be convinced that mimetypes actually does a good job at guessing.

martindurant:
Thank you @damnMeddlingKid

PR #581, opened 2022-01-17
drop dup Key and Size returned in _info

raybellwaves:
Closes https://github.com/fsspec/s3fs/issues/580

Welcome to continue the convo here on whether this is a good idea of not. As well as if i'm missing anything.

I'll have to address failing tests: https://github.com/fsspec/s3fs/runs/4843522436?check_suite_focus=true#step:6:467


martindurant:
Are there any objections to this? I intend to merge by day's end if not.

PR #583, opened 2022-01-20
add development page to docs

raybellwaves:
I'm would like to add this page as I have to recently (re)think about create a development environment and running tests.

My goal is this would be copy-pastable to help new contributors.

I imagine this could be expanded in the future e.g. https://docs.dask.org/en/stable/develop.html#install

### DIFF
--- a/docs/source/development.rst
+++ b/docs/source/development.rst
@@ -3,8 +3,7 @@ Development
 
 Create a development environment::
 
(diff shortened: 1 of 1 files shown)


raybellwaves:
If interested you can add a read the docs check here in the CI through some repo settings

martindurant:
Yes please, I would love both the dev page and docs test in the CI.

PR #585, opened 2022-01-24
add skip_instance_cache kwargs to super_kwargs

jinmel:
Seems like skip_instance_cache is missing from super_kwargs :)

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -214,7 +214,8 @@ class S3FileSystem(AsyncFileSystem):
         self.kwargs = kwargs
         super_kwargs = {
             k: kwargs.pop(k)
-            for k in ["use_listings_cache", "listings_expiry_time", "max_paths"]
+            for k in ["use_listings_cache", "listings_expiry_time", "max_paths",
+                      "skip_instance_cache"]
             if k in kwargs
         }  # passed to fsspec superclass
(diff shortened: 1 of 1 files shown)


isidentical (approved):

jinmel:
can you run the workflow? @isidentical

isidentical:
Sure thing!

martindurant:
Although this change should not break anything, it will have no effect, because `skip_instance_class` is swallowed by the metaclass on [this line](https://github.com/fsspec/filesystem_spec/blob/master/fsspec/spec.py#L61) and should never appear in the kwargs. Did you have a case where the kwarg was causing a problem?

jinmel:
@martindurant do you mean that this change is not necessary? I'll close this PR if so

+ Is metaclass instantiated before any class constructor?

martindurant:
>  do you mean that this change is not necessary?

Correct, unless you have some specific case I haven't thought of showing a problem.

> Is metaclass instantiated before any class constructor

Sort of. Specifically, the metaclass's `__call__` method is used to create the new instance. In the case that we use a cached instance, `__init__` isn't called *at all*.

jinmel:
Seems like the skip_instance_cache is working without this change. I am closing this PR. 

@martindurant thanks for your insight 😄

Issue #589, opened 2022-02-07
callbacks not supported in `_get_file` and `_put_file`


ap--:
Hello everyone,

I was playing with `fsspec.callbacks` trying to monitor the local file download when using a CachingFileSystem, and noticed that `get_file` and `put_file` in s3fs currently don't support callbacks.

```python
Traceback (most recent call last):
  File "_cache_test.py", line 70, in <module>
    cfs.get(path, fn, callback=MyCallback(flush=True))
  File "/Users/poehlmann/Environments/envs/pasu/lib/python3.8/site-packages/fsspec/implementations/cached.py", line 402, in <lambda>
    return lambda *args, **kw: getattr(type(self), item).__get__(self)(
  File "/Users/poehlmann/Environments/envs/pasu/lib/python3.8/site-packages/fsspec/spec.py", line 776, in get
    self.get_file(rpath, lpath, **kwargs)
  File "/Users/poehlmann/Environments/envs/pasu/lib/python3.8/site-packages/fsspec/asyn.py", line 91, in wrapper
    return sync(self.loop, func, *args, **kwargs)
  File "/Users/poehlmann/Environments/envs/pasu/lib/python3.8/site-packages/fsspec/asyn.py", line 53, in sync
    coro = func(*args, **kwargs)
TypeError: _get_file() got an unexpected keyword argument 'callback'
``` 

I'll open a PR right away to add the callback keyword argument for both `_get_file` and `_put_file`.
Let me know if I should add tests too.

Cheers,
Andreas 😃

ap--:
Closed via: 6f844d4dfb70ad7f2397220663b907398659ccd6

PR #590, opened 2022-02-07
support callbacks in _get_file and _put_file

ap--:
Hello everyone,

This PR adds the callback keyword argument for both `_get_file` and `_put_file`.
See issue #589

Cheers,
Andreas 😃

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -45,7 +45,7 @@ if "S3FS_LOGGING_LEVEL" in os.environ:
     setup_logging()
 
 
-MANAGED_COPY_THRESHOLD = 5 * 2 ** 30
+MANAGED_COPY_THRESHOLD = 5 * 2**30
 S3_RETRYABLE_ERRORS = (socket.timeout, IncompleteRead)
 
 if ClientPayloadError is not None:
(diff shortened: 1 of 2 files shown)


martindurant:
fsspec testing has a simple callback you can use to make sure that things work as they should

martindurant:
(@isidentical , maybe you could suggest a simple way to test the callbacks)

ap--:
Hi @martindurant 

Thanks for the feedback and the right pointer.
And sorry for the delay, it's been a busy day at work.
I'll push a test soon.

Cheers,
Andreas 😃

martindurant:
Looks like `black` updated and started to reformat things. Please update your local version of `black` to get the formatting as required in CI, or else we could pin the version being used to do the check.

ap--:
I added tests for `_get_file`, `_put_file` and the branching behaviour via `get` and `put`.

There's some inconsistencies regarding the callback call interface in `fsspec.asyn`.
Is `lazy_call` still supposed to be used?
It's not part of the `Callback` base class.

Should I open another issue for that in the filesystem_spec repository?

Cheers,
Andreas

martindurant:
> Should the changes due to the new release of black be included in the main branch first and I rebase my changes on top?

That would make sense, but then you would have to wait around for someone to  do it, so maybe it's easier to do it here.

martindurant:
Thank you

PR #594, opened 2022-02-11
tests: remove lazy_call and call methods from BranchingCallback

ap--:
Not needed anymore with fsspec/filesystem_spec#902 merged.

Cheers,
Andreas 😃

### DIFF
--- a/s3fs/tests/test_s3fs.py
+++ b/s3fs/tests/test_s3fs.py
@@ -912,15 +912,6 @@ def test_get_put_with_callback(s3, tmpdir):
     test_file = str(tmpdir.join("test.json"))
 
     class BranchingCallback(Callback):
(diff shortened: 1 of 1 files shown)


martindurant:
An unrelated test is failing, possibly due to moto (or botocore?) update. Do you have time to investigate or possibly fix, @ap-- ? It seems that setting a metadata attribute with None is causing no update, whereas the intended action is to remove that attribute.

ap--:
I'm on mobile for the rest of the day.
It looks like the tests install `moto>=3`,
Maybe the ci yaml just needs to limit moto's major version to `<3` for now.

But i can have a thorough look tomorrow 😃

ap--:
@martindurant that fixed it :+1: Ready to be merged.

PR #605, opened 2022-03-18
Allow empty etag in _info.

zijie0:
Similar to this pull request: https://github.com/fsspec/s3fs/pull/480 .

We encounter problems when accessing files hosted by Minio which might not be uploaded through Minio API (e.g., Minio on HDFS gateway, local files). This simple fix could solve problems like: https://github.com/iterative/dvc/issues/7161 .


martindurant:
Thanks for the super simple fix.

PR #609, opened 2022-03-23
Bumping version of aiobotocore to 2.2.0

jmleoni:
In a attempt to solve https://github.com/fsspec/s3fs/issues/558  I am following suggestion from the aiobotocore contributors to update its version to solve the issue https://github.com/aio-libs/aiobotocore/issues/901#issuecomment-1069891573

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,7 +33,7 @@ jobs:
       - name: Install dependencies
         shell: bash -l {0}
         run: |
-          conda install -c conda-forge pip botocore aiobotocore==2.1.0 "aiohttp<=4" "moto>=2.0,<3" pytest flake8 black -y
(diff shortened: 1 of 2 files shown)


martindurant:
Please also update in the test runs, e.g., https://github.com/fsspec/s3fs/blob/main/.github/workflows/ci.yml#L36

martindurant:
Ah, if aiobotocore 2.2.0 is not even out on conda(-forge), then we'll definitely have to wait.

jmleoni:
I have just launched a PR for aibotocore to be updated in conda-forge
https://github.com/conda-forge/aiobotocore-feedstock/pull/51

martindurant:
OK, please ping here when that is merged and the new packages ready.

jmleoni:
@martindurant the package is updated on condaforge. Can you try this build again please ?

PR #611, opened 2022-03-25
Refactor tests to use fixtures, add dependabot and move tests out of the module directory

orf:
By leveraging fixtures we remove the need to spin up and down a moto3 server on each test. This greatly improve the speed.

The fixtures also mean we can split tests out by directory and generally be a bit less "hard-codey". There where a few implicit test dependencies that this refactor picked up.

With `pytest-xdist` we can cut total test runtime from ~6-7 minutes to ~1 minute.


.github/workflows/ci.yml
```diff
@@ -33,18 +33,16 @@ jobs:
       - name: Install dependencies
         shell: bash -l {0}
         run: |
-          conda install -c conda-forge pip botocore aiobotocore==2.1.0 "aiohttp<=4" "moto>=2.0,<3" pytest flake8 black -y
+          pip install -r requirements.txt -r test_requirements.txt
```
martindurant, line 36: It's a pretty big change to swap out conda (/mamba) installation for pip. My life usually revolves around conda, so I would like to keep that, but I would understand adding a separate run with pip.

martindurant: (an action stage can be made to use a conda environment file instead of listing dependencies right in the run file, which is probably what you were after)

.github/workflows/ci.yml
```diff
@@ -33,18 +33,16 @@ jobs:
...
           pip install git+https://github.com/fsspec/filesystem_spec --no-deps
-          conda list
```
martindurant, line 38: It's always useful to list the versions that are actually being used in the test. Pip is fine for this.

tests/conftest.py
```diff
@@ -0,0 +1,206 @@
...
+                pass
+            timeout -= 0.1
+            time.sleep(0.1)
+        yield
+        proc.terminate()
+        proc.wait()
+
+
+@pytest.fixture()
+def boto3_client(endpoint_uri):
```
martindurant, line 134: This is a botocore session, not boto3

tests/conftest.py
```diff
@@ -0,0 +1,206 @@
...
+            "nested/file2": b"world",
+            "nested/nested2/file1": b"hello\n",
+            "nested/nested2/file2": b"world",
+        },
+        "glob": {"file.dat": b"", "filexdat": b""},
+    }
+
+
+@pytest.fixture()
+def bucket_names():
```
martindurant, line 71: This and the file_* are not really fixtures, because they don't do any setup/teardown. Is it not simpler to leave them as constants? Is this the only way to avoid importing the values in test modules?

martindurant (commented):
In most dask-related projects and all of the rest of fsspec, the tests directory lives as a package within the main source. This makes the tests and any definitions there importable and more easily allows for test "utility" functions. Also, it means that the test code gets distributed along with the wheel, which might be useful. Is there a good justification for moving the tests into a top-level folder?

martindurant:
@ianthomas23 , there may be some useful work in this PR, since you have been building fixtures anyway. I will close this PR at this point, since there's not much chance of completing it, I think.

ianthomas23:
@martindurant Will do, thanks for the heads-up.

PR #623, opened 2022-05-10
update changelog & release procedure

achimgaedke:
Hey @martindurant - what are your plans for a release?

Here is a little helping hand - but please don't feel too rushed if you have some other business to finish.


martindurant:
This is a fine change

jmleoni:
@martindurant Do you think you can issue a release so this change can be used with installations of s3fs done from pypi ? (the changelog was updated, but the release wasn't issued)

PR #634, opened 2022-08-03
Fix syntax error hiding minio code sample

yuvipanda:
Without this, the code sample is just hidden

From https://s3fs.readthedocs.io/en/latest/index.html?highlight=minio#s3-compatible-storage

image


martindurant:
Oops :)

PR #640, opened 2022-08-22
Fix `pip install --pre s3fs`

barentsen:
It is not possible to run `pip install --pre s3fs` right now because the two latest pre-releases of `aiohttp` are broken (v4.0.0a0 and v4.0.0a1).

This PR proposes excluding these two versions from the s3fs requirements.

More details: https://github.com/fsspec/filesystem_spec/pull/1018


martindurant:
Will gcsfs need this too?

PR #648, opened 2022-08-31
Added a slightly changed version from the awscli find bucket function

saineju:
Added a slightly modified version of the find_bucket_key from [awscli](https://github.com/aws/aws-cli/blob/733f856ebdfc59edad6f4b7242b7027eeecec7d0/awscli/customizations/s3/utils.py#L217) ensure that the bucket name and paths are split properly

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -372,26 +372,25 @@ class S3FileSystem(AsyncFileSystem):
         """
 
         _S3_ACCESSPOINT_TO_BUCKET_KEY_REGEX = re.compile(
-            r'^(?P<bucket>arn:(aws).*:s3:[a-z\-0-9]*:[0-9]{12}:accesspoint[:/][^/]+)/?'
-            r'(?P<key>.*)$'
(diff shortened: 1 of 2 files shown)


martindurant:
Sorry to be so slow to respond here!
This looks great, running tests now. At a glance, not sure that the reference to "access-point" is intentional? Of course, I know nothing about what these URIs ought to look like.

saineju:
Thanks for fixing the linting issue, it's always something I forget. Just commenting for the question:

Not entirely sure what you mean regarding the reference to "access-point" as such, but I would assume those are intentional. AWS offers a possibility to create access points for S3 buckets as one way to manage access to shared s3 buckets. There can be several access points for one bucket and those can have different permissions for access. While the ap's are mostly used similarly to s3 buckets, the bucket URI is either the ARN of the acccess point or the access point alias. 

While the Alias would parse similarly to normal s3 bucket name, it is autogenerated and cannot be changed for the buckets, so using it in automation can be harder than just using the ARN, hence the need for allowing also ARN's in the names.

martindurant:
Thanks for the explanation. I suppose it's the kind of thing that doesn't make that much sense until you need to use it :)

PR #654, opened 2022-09-22
find/ls: use list_object_versions instead of list_objects_v2 for version aware filesystems

pmrowla:
Fixes https://github.com/fsspec/s3fs/issues/653

- Version aware filesystems use `list_object_versions` instead of `list_objects_v2` for listing directories
- `fs.dircache` will only cache entries for the latest file versions (file versions with `IsLatest == True`)
- Individual file lookups (via `_info`) will return the cached version when the requested version is latest (when requested `version_id is None`)
- When `_info` is doing a lookup for a specific (non-`None`) version_id, it will check if the cached version matches and then make the required `HeadObject` call if needed.

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -659,7 +659,13 @@ class S3FileSystem(AsyncFileSystem):
                 logger.debug("Get directory listing page for %s" % path)
                 await self.set_session()
                 s3 = await self.get_s3(bucket)
-                pag = s3.get_paginator("list_objects_v2")
+                if self.version_aware:
+                    method = "list_object_versions"
+                    contents_key = "Versions"
+                else:
+                    method = "list_objects_v2"
+                    contents_key = "Contents"
+                pag = s3.get_paginator(method)
                 config = {}
                 if max_items is not None:
                     config.update(MaxItems=max_items, PageSize=2 * max_items)
@@ -674,10 +680,11 @@ class S3FileSystem(AsyncFileSystem):
                 dircache = []
                 async for i in it:
                     dircache.extend(i.get("CommonPrefixes", []))
-                    for c in i.get("Contents", []):
-                        c["type"] = "file"
-                        c["size"] = c["Size"]
-                        files.append(c)
+                    for c in i.get(contents_key, []):
+                        if not self.version_aware or c.get("IsLatest"):
+                            c["type"] = "file"
+                            c["size"] = c["Size"]
+                            files.append(c)
                 if dircache:
                     files.extend(
                         [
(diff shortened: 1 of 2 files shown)


pmrowla:
Ideally I think we should also be caching the full results of `list_object_versions`, so dircache would actually contain another level of nesting (for files -> versions), but for now this PR maintains current behavior while also fixing the original bug.

I'll most likely send a follow up PR that handles this additional caching at some point (it will probably be needed for downstream DVC eventually) but I don't think I'll be able to get to it right away.

s3fs/core.py
```diff
@@ -1129,12 +1136,24 @@ async def _info(self, path, bucket=None, key=None, refresh=False, version_id=Non
...
+        if not refresh:
             out = self._ls_from_cache(path)
-            out = [o for o in out if o["name"] == path]
-            if out:
-                return out[0]
-            return {"name": path, "size": 0, "type": "directory"}
+            if out is not None:
+                if self.version_aware and version_id is not None:
+                    # If cached info does not match requested version_id,
+                    # fallback to calling head_object
```
efiop, line 1144: So looks like it is kinda defeating the purpose of dircache, right? Since calling head_object for every object from dircache is more expensive than just dropping the dircache and doing a list_object_versions call.

s3fs/core.py
```diff
@@ -645,10 +651,11 @@ async def _lsdir(
                 dircache = []
                 async for i in it:
                     dircache.extend(i.get("CommonPrefixes", []))
-                    for c in i.get("Contents", []):
-                        c["type"] = "file"
-                        c["size"] = c["Size"]
-                        files.append(c)
+                    for c in i.get(contents_key, []):
+                        if not self.version_aware or c.get("IsLatest"):
```
martindurant, line 655: So the only potential downside I see, is that this call might take substantially longer. Is there a way to specify that we do wand the version info, but only for the latest entry at each key?

pmrowla: Unfortunately there is not. `list_objects_v2` always returns the latest entry, but does not include version IDs. `list_object_versions` does not allow you to specify that you only want the latest versions.

We could pass a jmespath query into the paginator to filter results by the `IsLatest` key, but that's still client-side filtering and doesn't actually help us avoid needing to get all of the versions from S3.

martindurant: So for the user who doesn't want this problem and knows they will be working only with latest versions, they should not use the versioned flag on the filesystem instance, even when the bucket is indeed versioned? We should document this.

martindurant: Perhaps in the docstring of S3FileSystem at the argument version_aware?

martindurant:
Thank you

PR #655, opened 2022-09-27
If listing expires between checking it, and using it

### LINKED ISSUE #651: ls() crashes once directory listing has expired
When `listings_expiry_time` is set and the listing expires, `ls` crashes with 

```
[07/Sep/2022 13:28:51] [ERROR] [task]    File "/usr/local/lib/python3.7/site-packages/s3fs/core.py", line 1010, in _info
[07/Sep/2022 13:28:51] [ERROR] [task]      out = [o for o in out if o["name"] == path]
[07/Sep/2022 13:28:51] [ERROR] [task]  TypeError: 'NoneType' object is not iterable
```

because `_ls_from_cache` returns None

martindurant:
Fixes #651 

@Eugeny, can you please test? You seemed to be hitting a corner case where list-from-cache returned None a moment after just having checked that the value was not in fact None. Now we only call once.

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -1158,12 +1158,13 @@ class S3FileSystem(AsyncFileSystem):
         if path in ["/", ""]:
             return {"name": path, "size": 0, "type": "directory"}
         version_id = _coalesce_version_id(path_version_id, version_id)
-        if not refresh and self._ls_from_cache(path) is not None:
+        if not refresh:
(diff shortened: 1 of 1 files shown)


martindurant:
OK, please let me know if all is still well in the next few days!

Eugeny:
Working fine to far, thanks for the fix!

martindurant:
Thanks for reporting back. Merging this.

PR #661, opened 2022-11-08
ls: support listing versions

efiop:


martindurant:
Simple&good.
This made it into the coming release.

Issue #662, opened 2022-11-08
S3FileSystem NotImplementedError


clayms:
Code that worked without any issue a few days ago now fails.  

`s3fs.__version__`:  `'2022.10.0'` 

```python
import s3fs

KEY = os.getenv('AWS_ACCESS_KEY_ID')
SECRET_KEY = os.getenv('AWS_SECRET_ACCESS_KEY')

fs = s3fs.S3FileSystem(key=KEY, secret=SECRET_KEY)
fs.ls("s3://my-bucket")
```

gives:

```python
---------------------------------------------------------------------------
NotImplementedError                       Traceback (most recent call last)
<ipython-input-4-9ec24d2dd6bc> in <cell line: 2>()
      1 fs = s3fs.S3FileSystem(key=KEY, secret=SECRET_KEY)
----> 2 fs.ls("s3://my-bucket")

/data/code-envs/python/polars/lib/python3.9/site-packages/fsspec/spec.py in ls(self, path, detail, **kwargs)
    338         dicts if detail is True.
    339         """
--> 340         raise NotImplementedError
    341 
    342     def _ls_from_cache(self, path):

NotImplementedError: 

```

martindurant:
I can only suppose you have some environment mismatch between fsspec and s3fs -  please check.

```
fs = s3fs.S3FileSystem()
fs.ls
```
->
```
<function s3fs.core.S3FileSystem._ls(path, detail=False, refresh=False)>
```
(**not** AbstractFileSystem.ls with the not-implemented)

clayms:
@martindurant 

What I get in the same session that has the error. 

```python 
fs = s3fs.S3FileSystem()
fs.ls
```
->
```
<bound method AbstractFileSystem.ls of <s3fs.core.S3FileSystem object at 0x7fd314fa5e20>>
```

clayms:
@martindurant 

How to get around this?

clayms:
Versions of both `fsspec` and `s3fs` .

```
fsspec==2022.10.0
s3fs==2022.10.0
```

clayms:
I think it might be `pyarrow==10.0.0` causing the issue.

https://arrow.apache.org/docs/python/filesystems.html

martindurant:
[...]

PR #663, opened 2022-11-09
Allow use of `s3fs` when `fsspec` is mediated by a `LazyLoader`

alexander-beedie:
Hi all, greetings from Polars... 

We had an issue (https://github.com/pola-rs/polars/issues/5451) raised just now, relating to #662 here. I've started debugging this on our side; we make use of `importlib`'s [LazyLoader](https://docs.python.org/3/library/importlib.html#importlib.util.LazyLoader) class to defer loading several modules, with `fsspec` being amongst them.
 
I've spotted a one-liner you can add which fixes the issue at (apparently) no cost, and it will guard against similar gremlins if anyone else lazy-loads `fsspec` - just to be clear, I will keep looking for a lower-level solution on our side, as it shouldn't really be on you to handle somebody else lazy-loading  ;)

As you can tell from the simplicity of the patch, the jump straight to submodule imports appears to bypass the standard `LazyLoader` machinery for detecting module usage and properly loading the root (despite the submodules being lazy-loaded successfully!) Most likely connected to some subtlety of [submodule](https://docs.python.org/3/reference/import.html#submodules) binding/references, but we shall see...

Regards.

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -11,6 +11,7 @@ import re
 
 from urllib3.exceptions import IncompleteRead
 
+import fsspec  # noqa: F401
 from fsspec.spec import AbstractBufferedFile
 from fsspec.utils import infer_storage_options, tokenize, setup_logging as setup_logger
 from fsspec.asyn import (


alexander-beedie:
PS: @martindurant, seems we both have a background in astrophysics - I wonder what the Venn diagram of ex-astrophysicists now involved in pydata projects looks like... :D

martindurant:
I am OK to include this, as it is indeed no cost, but you'll need a flake8 exception:
```
s3fs/core.py:14:1: F401 'fsspec' imported but unused
```

alexander-beedie:
> ```
> s3fs/core.py:14:1: F401 'fsspec' imported but unused
> ```

No problem; just updated accordingly.

martindurant:
> I seem to have triggered this PR from 'main',

That's fine

alexander-beedie:
> That's fine

Ok, synced-up and re-pushed.

alexander-beedie:
FYI: we now have a lower-level fix on our side that will go out in the next release (https://github.com/pola-rs/polars/pull/5459).

martindurant:
OK, thanks

Issue #659, opened 2022-11-03
fs.put() writes the object under a different target path if it is called multiple times


tilan7663:
Hello,

We are seeing very strange behavior with fs.put API after upgrading our s3fs version from 0.4.2 to 2022.10.0. It seems like the object path changes when fs.put() is called multiple times.

### How to reproduce

**Local filesystem**
```
mkdir -p /home/user/foo_dir
echo "text" > /home/user/foo_dir/hello.txt
```

**Python**
```
import s3fs
import uuid

s3 = s3fs.S3FileSystem()
session_id = uuid.uuid4().hex

s3_path = f"s3://bucket/{session_id}/" // s3_path doesn't exist yet in S3

s3.put("/home/user/foo_dir", s3_path, recursive=True) // first attempt

// after the first s3.put() is called, the object is created under s3://bucket/{session_id}/hello.txt which is the expected behavior

s3.put("/home/user/foo_dir", s3_path, recursive=True) // second attempt

// after the second attempt, the object path is changed and it is created under s3://bucket/{session_id}/foo_dir/hello.txt. The old version of library does not produce this behavior
```

After we did some code tracing, it seems like the reason that the object path is changed on the second attempt is because of a [patch ](https://github.com/fsspec/filesystem_spec/commit/576b1a131416d6af56bb1561bdeac86fe29155b5) that was made previously to address a somewhat related [issue](https://github.com/fsspec/s3fs/issues/521). And interestingly, after we revert this patch, we couldn't reproduce the behavior that was described in the original issue and new behavior is same as the the one in s3fs==0.4.2. Wondering if there is something that could be changed since this patch was merged which somehow makes this patch behave differently.

Please let me know if you have any questions.

Thanks,
Tian

ianthomas23:
I cannot reproduce this locally. I've submitted a PR (#666) containing a test for it to see if it passes CI.

ianthomas23:
@tilan7663 Since fsspec/filesystem_spec#1148 and fsspec/filesystem_spec#1163 this is now expected behaviour as it is consistent with command-line `cp` and `scp`. To obtain the behaviour that you would like you just need to append a slash to your source directory, i.e.
```python
s3.put("/home/user/foo_dir/", s3_path, recursive=True)
```

I am going to close this issue as I think no further action is necessary. Feel free to re-open it if you would like to discuss it further.

PR #666, opened 2022-11-30
Add test that s3fs.put of single file in directory is repeatable

ianthomas23:
I cannot reproduce issue #659 locally, which is that an `s3fs.put` of a single file in a directory produces different results the first and second times it is run. So this PR adds an explicit test for this situation, to run it through the full CI suite.

### DIFF
--- a/s3fs/tests/test_s3fs.py
+++ b/s3fs/tests/test_s3fs.py
@@ -2112,6 +2112,26 @@ def test_put_single(s3, tmpdir):
     assert s3.cat(test_bucket_name + "/dir/abc") == b"text"
 
 
+def test_put_single_repeat(s3, tmpdir):
+    # See https://github.com/fsspec/s3fs/issues/659
+    # Potential different behaviour if s3.put called multiple times with
+    # a directory containing a single file
+    directory = os.path.join(str(tmpdir), "issue659_dir")
+    os.mkdir(directory)
+    test_file = os.path.join(directory, "issue659_file")
+    with open(test_file, "w") as f:
+        f.write("some text")
+
+    assert not s3.exists(test_bucket_name + "/issue659_dir")
+
+    for _ in range(2):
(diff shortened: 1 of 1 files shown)


martindurant:
@tilan7663 , can you produce a test similar to the code here that shows your unexpected behaviour?

tilan7663:
@martindurant @ianthomas23 Thanks for taking a look. I can cherry-pick the change and see if it yields the same behaviour. The behaviour we noticed prior the upgrade is similar to the following

```
def test_put_single_repeat_2(s3, tmpdir):
    # See https://github.com/fsspec/s3fs/issues/659
    # Potential different behaviour if s3.put called multiple times with
    # a directory containing a single file
    import uuid

    directory = os.path.join(str(tmpdir), "issue659_dir")
    os.mkdir(directory)
    test_file = os.path.join(directory, "issue659_file")
    with open(test_file, "w") as f:
        f.write("some text")

    session_id = uuid.uuid4().hex
    s3_dir = os.path.join(test_bucket_name, session_id)
    s3_file = os.path.join(test_bucket_name, session_id, "issue659_file")
    s3_dir_not_exist = os.path.join(test_bucket_name, session_id, "issue659_dir")
    s3_file_not_exist = os.path.join(test_bucket_name, session_id, "issue659_dir", "issue659_file")

    assert not s3.exists(s3_path)

    for _ in range(2):
        s3.put(directory, s3_dir, recursive=True)
        assert s3.isdir(s3_dir)
        assert s3.isfile(s3_file)
        assert not s3.exists(s3_dir_not_exist)
        assert not s3.exists(s3_file_not_exist)
        assert s3.cat(s3_file) == b"some text"
```

ianthomas23:
@tilan7663 Thanks for the clarification. My test doesn't accurately represent your problem as I've missed out an extra directory level. I will look into it further.

ianthomas23:
I am closing this PR as it has been superseded by #691.

PR #668, opened 2022-12-05
fix flake8 config for flake8 6.0.0

dtrifiro:
see https://github.com/PyCQA/flake8/issues/1750


dtrifiro:
This was causing CI to fail: https://github.com/fsspec/s3fs/actions/runs/3586175988/jobs/6035055289#step:7:1

PR #676, opened 2022-12-23
Handle temporary errors in HeadBucket

Eugeny:
For freshly created S3 buckets, boto `head_bucket` will fail with `ERR An error occurred (403) when calling the HeadBucket operation: Forbidden` even though ResponseMetadata will contain the correct region header - for up to an hour in my experience.

This PR adds handling for this and continues if the region header is present despite HeadBucket failing.

### DIFF
--- a/s3fs/utils.py
+++ b/s3fs/utils.py
@@ -37,12 +37,18 @@ class S3BucketRegionCache:
 
         try:
             response = await general_client.head_bucket(Bucket=bucket_name)
-        except ClientError:
-            logger.debug(
-                "RC: HEAD_BUCKET call for %r has failed, returning the general client",
-                bucket_name,
+        except ClientError as e:
+            region = (
+                e.response["ResponseMetadata"]
+                .get("HTTPHeaders", {})
+                .get("x-amz-bucket-region")
             )
-            return general_client
+            if not region:
+                logger.debug(
+                    "RC: HEAD_BUCKET call for %r has failed, returning the general client",
+                    bucket_name,
+                )
+                return general_client
         else:
             region = response["ResponseMetadata"]["HTTPHeaders"]["x-amz-bucket-region"]
 


martindurant:
Thanks for digging this up! While we cannot make a test for CI, how would I go about reproducing the error you are fixing, so I can understand it?

Eugeny:
In my experience, just creating a fresh S3 bucket on AWS in non-US region and trying to `.ls()` the files in it already triggers the error as s3fs would fail to resolve the region and fall back to us-east-1.

martindurant:
So there is a latency time between creating the bucket and when the normal method would work?

Note that you are failing on linting errors - consider running pre-commit.

Eugeny:
Thanks, I've just fixed the formatting. 

Yes, there's latency between the bucket getting created and `HeadBucket` stopping returning HTTP redirects (which boto3 interprets as errors, hence head_bucket throwing). I haven't tested this, but I have a hunch that latest boto3 could have fixed this as well by correctly not interpreting redirects as errors.

Can't pull the exact response AWS gives when authorized, but you can see the difference even with an anonymous request:

An older bucket:

```
❯ curl -sI https://aaa.s3.amazonaws.com/     
HTTP/1.1 403 Forbidden
x-amz-bucket-region: eu-central-1
x-amz-request-id: xxx
x-amz-id-2: xxx
Content-Type: application/xml
Date: Thu, 29 Dec 2022 19:18:32 GMT
Server: AmazonS3
```

A fresh bucket:

```
❯ curl -sI https://bbb.s3.amazonaws.com/
HTTP/1.1 307 Temporary Redirect
x-amz-bucket-region: eu-north-1
x-amz-request-id: xxx
x-amz-id-2: xxx
Location: https://bbb.s3.eu-north-1.amazonaws.com/
Content-Type: application/xml
Date: Thu, 29 Dec 2022 19:18:37 GMT
Server: AmazonS3
```

martindurant:
Got it, thank you.

PR #678, opened 2022-12-30
Set default cache type to "readahead"

mrocklin:
See analysis by @sneha5gsm in https://github.com/fsspec/s3fs/issues/677

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -205,10 +205,10 @@ class S3FileSystem(AsyncFileSystem):
     default_fill_cache : Bool (True)
         Whether to use cache filling with open by default. Refer to
         ``S3File.open``.
-    default_cache_type : string ('bytes')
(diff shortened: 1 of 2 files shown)


martindurant:
Breaks one specific test relying on the previous default value of the cache. Could pass "bytes" to the open() call to make it the same test as before, or alter the test. I can do it if you don't have time.

mrocklin:
Oops, I screwed up my local git config and accidentally pushed this to the main fork.  My apologies.

mrocklin:
Force pushed.  Looks like it's coming up green.  Thanks for the suggestion @martindurant

martindurant:
(please delete the branch when you are done with it)

mrocklin:
Done.  Again, my apologies.  

Thanks for merging!

Thanks also to @sneha5gsm whose work this really is.

PR #691, opened 2023-01-24
Add cp, get and put recursive tests

ianthomas23:
This add new tests for the correct asynchronous `cp`, `get` and `put` behaviour added to `fsspec` in PR fsspec/filesystem_spec#1163. The tests are derived from, and almost identical to, similar tests in `fsspec`'s `test_memory.py`.

CI should fail when running against `fsspec` default branch as it is now, and pass after fsspec/filesystem_spec/pull/1163 is merged.


ianthomas23:
@martindurant Please can you approval the CI workflow here and in https://github.com/fsspec/gcsfs/pull/521 and we can see if they both pass the tests.

Issue #693, opened 2023-01-31
API Reference docs no space between attribute name and type


williamjamir:
In the API reference section, the attribute name and type have no space, giving the wrong impression of the attribute name.
![image](https://user-images.githubusercontent.com/5083518/215727926-2454f718-7d34-476f-8d48-6ab5eb70f4ef.png)
![image](https://user-images.githubusercontent.com/5083518/215729112-70c3d1d5-928a-4ea5-9334-2f172ff4823e.png)

ianthomas23:
Thanks @williamjamir. I know what the cause of this is as I fixed the same problem in `fsspec` recently. I'll submit a fix tomorrow.

ianthomas23:
Fixed by #697.

PR #696, opened 2023-02-01
Update docs build

ianthomas23:
This is a WIP to update the docs build to use a Read the Docs v2 config file. I was originally intending just to fix issue #693 but it turns out that RTD is building on Python 3.7 for which quite a few of the dependencies are a bit old and no longer updated. So this is updating the docs build system to follow a similar approach to `fsspec` and to fix a variety of doc-related issues at the same time.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,18 @@
+version: 2
+
+build:
+  os: ubuntu-22.04
+  tools:
+    python: miniconda3-4.7
+
+conda:
+  environment: docs/environment.yml
+
+python:
+  install:
+    - method: pip
+      path: .
+
+sphinx:
+  configuration: docs/source/conf.py
+  fail_on_warning: true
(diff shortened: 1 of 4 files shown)


ianthomas23:
It seems that in order to test this we would need to alter the RTD Advanced Settings to build on PRs and not just commits to main branch.

martindurant:
> it turns out that RTD is building on Python 3.7 

With the conda env option, we can pick our python version, no?

martindurant:
I have enabled PR builds in RTD

ianthomas23:
> With the conda env option, we can pick our python version, no?

Yes, we can specify it in the `docs/environment.yml` that we tell RTD to use.

martindurant:
Oh, but I do remember that conda (as opposed to mamba) was using a lot of memory and crashing RTD builds at some point.

ianthomas23:
> Oh, but I do remember that conda (as opposed to mamba) was using a lot of memory and crashing RTD builds at some point.

OK, we'll see what happens.

ianthomas23:
RTD hasn't automatically built the docs for this, presumably because the PR had already been created. I'll close this and create another PR to see if that works.

PR #710, opened 2023-03-14
bump aiobotocore to 2.5.0

chriszs:
Bumps aiobotocore to 2.5.0 to resolve a dependency conflict I'm seeing with a package that depends on the latest boto3.


shcheklein:
`aiobotocore~=2.4.2` was causing SSO failures in some cases - https://github.com/iterative/dvc/issues/9294 . It's great that it's fixed now. @martindurant is there anything that prevents from making a release?

PR #715, opened 2023-03-28
Set cache_regions to True by default

martindurant:
This seems to solve a number of aiobotocore redirect access issues, and aside from complexity, I don't see a downside.

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -899,6 +899,7 @@ class S3FileSystem(AsyncFileSystem):
     makedirs = sync_wrapper(_makedirs)
 
     async def _rmdir(self, path):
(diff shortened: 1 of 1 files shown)


martindurant:
cc @Eugeny 

One small caveat I have here, is the possibility of HEAD_BUCKET failing. I recebntly discovered that the following always succeeds (could be aiohttp rather than requests):
```
requests.head(f"https://{bucket}.s3.amazonaws.com").headers["x-amz-bucket-region"]
```
Do you think we should switch to this simpler form? I should be faster too, since it involves no auth.

Another specific thing that occurred to me: we should never use region caching together with endpoint_url, right?

martindurant:
(cc @efiop , if interested)

Issue #717, opened 2023-03-31
get() is including extra object path information in the local path after upgrading to version 2023.3.0


gmweaver:
After upgrading from 2022.11.0 to 2023.3.0, I am seeing different behavior when copying a list of remote paths to a local directory via `get`. Specifically, it seems to be including extra object path data in the final local path (looks like only one level up).

As an example, assume I am copying a set of files to local.

```
import os

import s3fs

remote_objs_to_copy = [
    "s3://my-bucket/path/to/some_data/data1.parquet", 
    "s3://my-bucket/path/to/some_data/data2.parquet",
]
target_local_dir = "/tmp/testing_s3fs/"

s3fs.S3FileSystem(...).get(remote_objs_to_copy, target_local_dir)
```

Running `os.listdir(target_local_dir)`.

v2022.11.0 returns: 

`["/tmp/testing_s3fs/data1.parquet", "/tmp/testing_s3fs/data2.parquet"]`

v2023.3.0 returns: 

`[some_data]`

Listing files under `some_data`, `os.listdir(f"{target_local_dir}/some_data")`, I do see the files.

ianthomas23:
I cannot reproduce this, see new PR (https://github.com/fsspec/s3fs/pull/718) that behaves as expected, i.e. does not include an extra directory.

@gmweaver:

1. Can you post your `pip list` or `conda list` to confirm that you have consistent versions of `fsspec` and `s3fs` installed?
2. Your `v2022.11.0` listing in your original post is of the `s3` filesystem not the `local` one. Can you check this please.

gmweaver:
1. We use `pip-compile` and the generated requirements file has:

```
fsspec==2023.3.0
```

2. Apologies, that is a typo, updated initial post with correct file paths.

Can a test be added where there is an ending slash: `target = os.path.join(tmpdir, "target/")`

gmweaver:
Also, tried to run the suggested test above locally, but I run into errors that look related to mocking boto/s3:

```
python -m venv venv
pip install -r requirements.txt -r test_requirements.txt
pytest
```

I see `ERROR s3fs/tests/test_mapping.py::test_simple - botocore.exceptions.EndpointConnectionError: Could not connect to the endpoint URL: "http://127.0.0.1:5555/test"`

ianthomas23:
Ah yes, my mistake. With the trailing slash on the target directory I can reproduce this, the `some_data` directory is created and it shouldn't be.

[...]

PR #718, opened 2023-03-31
Add test_get_two_files

ianthomas23:
Helper to test issue #717. It is equivalent to the previous function `test_cp_two_files` but using `get` to local filesystem rather than `cp`.


ianthomas23:
Closing this. We need to deal with such cases in an organised way, not just one at a time.

PR #727, opened 2023-04-16
Remove duplicate class S3AsyncStreamedFile

makistsantekidis:
S3AsyncStreamedFile seems to be defined twice in line s3fs/core.py#2292-2320

This PR removes the second definition.


martindurant:
Huh, how did that happen!
Thanks for the spot.

PR #731, opened 2023-05-01
Fix bug were SlowDown retries could result in BadDigest failures for small files

MatthijsPiek:
When doing _put_file with small files, a retryable error like `SlowDown` could ultimately result in `BadDigest` failures.
The `botocore` code would read the entire file passed into `put_object`. The `error_wrapper` would then retry with the same file object but nothing would seek back to the start again.

This is now fixed by not using a file object but, similar to the multi-part upload, read the content into bytes first. These bytes can be retried without seeking. This resolves the problem.

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -1107,10 +1107,10 @@ class S3FileSystem(AsyncFileSystem):
 
         with open(lpath, "rb") as f0:
             if size < min(5 * 2**30, 2 * chunksize):
+                chunk = f0.read()
                 await self._call_s3(
-                    "put_object", Bucket=bucket, Key=key, Body=f0, **kwargs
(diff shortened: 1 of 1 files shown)


martindurant:
Related: https://github.com/fsspec/gcsfs/pull/540

martindurant:
(I included he linked PR/issue, because aiohttp specifically complains about large `bytes` objects in payloads. Maybe aiobotocore deals with this internally, since apparently the file was not getting closed between retries)

martindurant:
I don't immediately see what the failure has to do with this PR
```
s3fs/tests/test_s3fs.py::test_xattr
```

MatthijsPiek:
Me neither, sorry. If I understand the test correctly, it doesn't seem to call into any of the changed code?

PR #745, opened 2023-06-05
Let async file know its size after first read

martindurant:


mukhery (approved):
can confirm that both `fsspec.generic.rsync(<s3_uri>, <local_path>)` and `fsspec.generic.GenericFileSystem().cp(<s3_uri>, <local_path>)` now work with this change

PR #746, opened 2023-06-08
s3fs: handle versioned paths in info and exists

efiop:


efiop:
Hm, I'm having a hard time reproducing this within the s3fs test suite. Looks like it depends on the fragile dircache state that we are hitting. We are creating new buckets within our dvc test suit and hitting this reliably.

martindurant:
moto isn't identical to aws s3 on versioning. I wonder how minio does. I was thinking for a while that it would be nice to have a second testing medium for the differences that exist.

efiop:
@martindurant We use moto as well, hence why I'm surprised.

efiop:
I'm not sure I can get to the bottom of it today, but I think this is more in sync with how adlfs does it and I can confirm that this works with our tests so probably still worth considering getting it in as is.

s3fs/core.py
```diff
@@ -978,7 +978,8 @@ async def _exists(self, path):
                 if self._ls_from_cache(path):
                     return True
             except FileNotFoundError:
-                return False
```
efiop, line 981: So we are getting this exception if we've succeded to list path parent, but haven't found matches. In that sense returning `False` makes sense here. But for versioned path we'll fail to match versioned name and will also get FileNotFoundError. Let me take another look...

s3fs/core.py
```diff
@@ -967,18 +967,37 @@ async def _ls(self, path, detail=False, refresh=False, versions=False):
...
         if key:
-            try:
-                if self._ls_from_cache(path):
-                    return True
-            except FileNotFoundError:
-                return False
+            cached_exists = self._exists_from_cache(path, bucket, key, version_id)
+            if cached_exists is not None:
+                return cached_exists
+
```
efiop, line 1000: Ok, so I think adlfs is unreasonably falling back to a real API call in https://github.com/fsspec/adlfs/blob/cdd9513013172ac379bf466abaf10e31fc9da6e9/adlfs/spec.py#L1388 . But at least that works 🙂 In s3fs we'll keep closer to old behaviour and will trust dircache by fixing how we interpret it here for versioned paths (only file could versioned, so we try to match version_id and if we fail to do that we report that file doesn't exist in cache).

efiop:
> Hm, I'm having a hard time reproducing this within the s3fs test suite. Looks like it depends on the fragile dircache state that we are hitting. We are creating new buckets within our dvc test suit and hitting this reliably.

Might be related to the fact that we use `ls(versions=True)` and maybe that causes the state to be different from the one in s3fs tests. That "corruption" of dircache would explain the situation. Taking a look...

efiop:
@martindurant Ok, found the cause and added a test. We should be good to go now. PTAL.

martindurant:
Good fight

efiop:
@martindurant Could you trigger a new s3fs release, please?

PR #757, opened 2023-07-20
For exists(), more precise error handling.

### LINKED ISSUE #750: S3FileSystem.exists returns false instead of error if credentials are wrong.
Short version: When calling `exists("/path/to/file")` on an `s3fs.S3FileSystem` with wrong credentials, instead of throwing an error that the credentials are wrong, it instead responds with False. This is misleading, since the file might actually exist but the credentials are simply wrong.

Tested with version `2023.6.0`, but existed in previous versions at least dating back to 2022 as well.

Was this a conscious decision or is this a bug? In case of the former, I heavily disagree with it...

kasuteru:
This PR improves the error handling logic for `s3fs.S3FileSystem.exists()` and resolves #750 

First change is that there is no longer a generic `except Exception` block, meaning that e.g. `ConnectionError` now correctly raises an error instead of returning `False` when querying a bucket.

Second change is that a warning is now logged when attempting to query the existence of a bucket that the user does not have access to - this is important since the bucket might actually exist. This is in line with Amazon's proposed behavior, see [here](https://docs.aws.amazon.com/AmazonS3/latest/userguide/example_s3_HeadBucket_section.html).

For more information on this, see related issue #750.

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -1019,12 +1019,17 @@ class S3FileSystem(AsyncFileSystem):
                     "list_objects_v2", MaxKeys=1, Bucket=bucket, **self.req_kw
                 )
                 return True
-            except Exception:
+            except (FileNotFoundError, PermissionError):
                 pass
             try:
                 await self._call_s3("get_bucket_location", Bucket=bucket, **self.req_kw)
                 return True
-            except Exception:
+            except FileNotFoundError:
+                return False
+            except PermissionError as e:
+                logger.warning(
+                    "Bucket %s doesn't exist or you don't have access to it.",
+                    bucket)
                 return False
 
     exists = sync_wrapper(_exists)


kasuteru:
@martindurant this implements solution 2) mentioned in #750. I am happy to change it to another solution, as discussed.

martindurant:
I think 2) is the right choice.

martindurant:
Can we make a test which surfaces the ConnectionError? Perhaps just set the endpoint_url to something silly.

kasuteru:
I will try to think of something, and also attempt to add tests for the other changes.

kasuteru:
I created test coverage for two of the three new cases: 

- A bucket that is not existent, but we have access to check (raises `FileNotFoundError`, which we catch)
- A `ConnectionError` as requested, symbolic for all other types of errors where we are actually unable to even check whether a bucket exists. This is what this PR is trying to fix (since currently, `False` is returned instead).

However, I cannot figure out how to create good test coverage for the third path, the `PermissionError` that is returned in case listing of buckets is prohibited by the endpoint. I am unsure on how to create a s3 file system locally that disallows listing of available buckets (which would be required for this test). The best I found is to trigger the permissionerror by providing the wrong credentials:


```
def test_exists_bucket_nonexistent_or_no_access(caplog):
    # Ensure that a warning is raised and False is returned if querying a
    # bucket that might not exist or might be private.
    fs = s3fs.S3FileSystem(key="asdfasfsdf", secret="sadfdsfds")
    assert not fs.exists("s3://this-bucket-might-not-exist/")
    assert caplog.records[0].levelname == "WARNING"
    assert "doesn't exist or you don't have access" in caplog.text
```

If this is good enough, it is already in the PR. But I am happy to improve this one.

martindurant:
Yes, that test is fine - it is, of course, failing on permissions with real AWS in this case (moto does not enforce credentials, which is why you couldn't test with that)

s3fs/tests/test_s3fs.py
```diff
@@ -2464,6 +2465,34 @@ def test_exists_isdir(s3):
     assert not s3.isdir(bad_path)
 
 
+def test_exists_raises_on_connection_error(monkeypatch):
+    # Ensure that we raise a ConnectionError instead of returning False if we
+    # are not actually able to connect to storage, by setting a fake proxy and
+    # then re-creating the S3FileSystem instance.
+    monkeypatch.setenv('https_proxy', 'https://fakeproxy.127.0.0.1:8080')
+    S3FileSystem.clear_instance_cache()
+    s3 = S3FileSystem(anon=False, client_kwargs={"endpoint_url": endpoint_uri})
```
martindurant, line 2474: `skip_instance_cache=True` so as not to affect future instances.
However, isn't it simpler to pass endpoint_url here?

s3fs/tests/test_s3fs.py
```diff
@@ -2464,6 +2465,34 @@ def test_exists_isdir(s3):
...
+    # bucket that might either not exist or be private.
+    fs = s3fs.S3FileSystem(key="asdfasfsdf", secret="sadfdsfds")
+    assert not fs.exists("s3://this-bucket-might-not-exist/")
+    assert caplog.records[0].levelname == "WARNING"
+    assert "doesn't exist or you don't have access" in caplog.text
+
+
+def test_exists_bucket_nonexistent(s3, caplog):
+    # Ensure that NO warning is raised and False is returned if checking bucket
+    # existance when we have full access.
```
martindurant, line 2491: Well, the bucket might still exist, but it's not one we can see. We happen to know for the sake of the test that it doesn't exist.

s3fs/tests/test_s3fs.py
```diff
@@ -2464,6 +2465,34 @@ def test_exists_isdir(s3):
...
+    s3 = S3FileSystem(anon=False, client_kwargs={"endpoint_url": endpoint_uri})
+    s3.invalidate_cache()
+    with pytest.raises(EndpointConnectionError):
+        s3.exists('this-bucket-does-not-exist/')
+
+
+def test_exists_bucket_nonexistent_or_no_access(caplog):
+    # Ensure that a warning is raised and False is returned if querying a
+    # bucket that might either not exist or be private.
+    fs = s3fs.S3FileSystem(key="asdfasfsdf", secret="sadfdsfds")
```
martindurant, line 2483: Comment: this tests against real AWS, might fail due to network issues.

s3fs/tests/test_s3fs.py
kasuteru: Not sure whether this is true for all S3 systems, but at least for our internal systems, I get a "PermissionError" when trying to access a bucket I cannot see (because I don't have permission to see it) but a `FileNotFoundError` like in this test when I would have permission to see, but it simply doesn't exist. This is why in the PR change, the warning would only be raised in case of `PermissionError` but not in case of `FileNotFoundError`. Should I change it to return the warning in both cases?

s3fs/tests/test_s3fs.py
kasuteru: Not sure what you mean with simpler? However, I added `skip_instance_url` as requested.

kasuteru:
I finally managed to set up pre-commit, I ran in SSL errors before. So hopefully the PR should not fail because of that any more. I also added changes as requested.

s3fs/tests/test_s3fs.py
```diff
@@ -2464,6 +2465,40 @@ def test_exists_isdir(s3):
...
+    # Ensure that we raise a ConnectionError instead of returning False if we
+    # are not actually able to connect to storage, by setting a fake proxy and
+    # then re-creating the S3FileSystem instance.
+    monkeypatch.setenv("https_proxy", "https://fakeproxy.127.0.0.1:8080")
+    S3FileSystem.clear_instance_cache()
+    s3 = S3FileSystem(
+        anon=False,
+        client_kwargs={"endpoint_url": endpoint_uri},
+        skip_instance_cache=True,
+    )
```
martindurant, line 2478:
```suggestion
    S3FileSystem.clear_instance_cache()
    s3 = S3FileSystem(
        anon=False,
        client_kwargs={"endpoint_url": endpoint_uri},
        skip_instance_cache=True,
        endpoint_url="https://fakeproxy.127.0.0.1:8080/"
    )
```

kasuteru:
Forgot to run pre-commit, this is now fixed... But I am honestly not sure where the warnings are coming from, I even installed Python 3.8 locally and ran the tests there, they run fine... 

The problem is the `Unclosed client session` warning, but I am unable to reproduce how s3fs.exists() triggers that...

image

martindurant:
The following may remove the warnings
```diff
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -541,6 +541,12 @@ class S3FileSystem(AsyncFileSystem):
     @staticmethod
     def close_session(loop, s3):
         if loop is not None and loop.is_running():
+            try:
+                loop = asyncio.get_event_loop()
+                loop.create_task(s3.__aexit__(None, None, None))
+                return
+            except RuntimeError:
+                pass
             try:
                 sync(loop, s3.__aexit__, None, None, None, timeout=0.1)
```

martindurant:
Your failures look like genuine problems.

For the client close warnings (which are not errors) https://github.com/fsspec/s3fs/pull/760 will fix this.

kasuteru:
If we don't manage to fix the warnings issue, I could also completely remove the warning from the PR. While I think it would be benefitial to warn users, it would already be an improvement if we manage to get rid of the generic `except Exception` clause. Let's see whether #760 changes things.

PR #764, opened 2023-07-31
fix: set default acl to private for r2 compatibility

### LINKED ISSUE #749: R2 compatibility: Header 'x-amz-acl' with value '' not implemented
Attempt to upload any file to [Cloudflare R2](https://developers.cloudflare.com/r2/api/s3/api/) storage leads to the following error

```
botocore.exceptions.ClientError: An error occurred (NotImplemented) when calling the PutObject operation: Header 'x-amz-acl' with value '' not implemented
```

I believe, that's because `s3fs` always passes `ACL` option to botocore methods even when it is just an empty string.

The problem can be reproduced as follows:

```python
with s3fs.open('hello.txt', 'w') as f:
    f.write('Hello!')
```

zanussbaum:
Fixes #749

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -599,7 +599,7 @@ class S3FileSystem(AsyncFileSystem):
         path,
         mode="rb",
         block_size=None,
-        acl="",
+        acl="private",
         version_id=None,
         fill_cache=None,
         cache_type=None,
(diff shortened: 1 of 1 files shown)


martindurant:
I would love some people using s3fs with various security configurations to give this a go. I'm prepared to wait a little while for that.

csqzhang:
Hi @martindurant, any plan to release this change? Currently we are also blocked by this ACL issue. I also see people encountering the same problem in different opened issues. I tested the change in our infra and it works and resolves the our issue.

martindurant:
A round of releases should be coming soon.

PR #771, opened 2023-08-11
Fix issue where empty dir name in s3 key breaks listdir api.

varunmittal91:
Tried writing this test case, but moto_service i think ignores these slashes anyway. Maybe somebody else will have a better luck with this test case.

```python
def test_simple_with_empty_folder_name(s3):
    e_base_path = f"{test_bucket_name}//test"
    e = e_base_path + "/" + "e"

    data = b"e" * (10 * 2**20)

    with s3.open(e, "wb") as f:
        f.write(data)

    with s3.open(e, "rb") as f:
        out = f.read(len(data))
        assert len(data) == len(out)
        assert out == data

    dir_list = s3.listdir(f"s3://{test_bucket_name}//test")
    assert len(dir_list) > 0
```


varunmittal91:
So s3 has a wiered thing, one of our customers have key s3://bucket_name//test//some_data

With lstrip it removes those '/' but without that we can do a listdir.

martindurant:
Note that `.find()` should I think still work in these odder cases.

varunmittal91:
oh okay, I can try and see if that resolves the usecase. Mahor problem is that i am using it with pyarrow and fspec which will do a listdir.

varunmittal91:
okay I will give it a try, thanks for looking into it.

PR #772, opened 2023-08-18
fill dircache on open

monken:
When opening a file for reading, the returned file info is not persisted in the `dircache`. This causes consecutive calls to the same file to trigger additional `HeadObject` requests.

```py
with fs.open(bucket + '/file.csv', 'rb') as f:
    f.read()

with fs.open(bucket + '/file.csv', 'rb') as f:
    f.read()
```

Expected result is that only one HeadObject request is made instead of two.

If you were to do a `fs.listdir(dataset_base_url)` before the `open()` calls, no `HeadObject` request would be triggered all because the cache will be filled from the list operation.

This PR will fill the dircache with the result from the `HeadObject` operation to prevent subsequent redundant requests.

This also fixes a bug where a file's size would not be reported properly if the directory was first listed, and a file was overwritten or pipe_file was used after a directory listing due to missing cache invalidation.

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -1048,7 +1048,7 @@ class S3FileSystem(AsyncFileSystem):
             )
         except ClientError as ex:
             raise translate_boto_error(ex)
-        self.invalidate_cache(self._parent(path))
+        self.invalidate_cache(path)
         return write_result
 
     touch = sync_wrapper(_touch)
@@ -1081,9 +1081,11 @@ class S3FileSystem(AsyncFileSystem):
         size = len(data)
         # 5 GB is the limit for an S3 PUT
         if size < min(5 * 2**30, 2 * chunksize):
-            return await self._call_s3(
+            out = await self._call_s3(
                 "put_object", Bucket=bucket, Key=key, Body=data, **kwargs
             )
+            self.invalidate_cache(path)
+            return out
         else:
 
             mpu = await self._call_s3(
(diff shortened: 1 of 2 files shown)


s3fs/core.py
```diff
@@ -1282,16 +1284,20 @@ async def _info(self, path, bucket=None, key=None, refresh=False, version_id=Non
...
-                    "name": "/".join([bucket, key]),
+                    "Size": out["ContentLength"],
+                    "name": path,
                     "type": "file",
+                    "Key": path,
                     "StorageClass": out.get("StorageClass", "STANDARD"),
                     "VersionId": out.get("VersionId"),
                     "ContentType": out.get("ContentType"),
                 }
+                self.dircache[path] = [info]
```
martindurant, line 1299: This is unfortunately incorrect. The dircache should be `{path: [list-of-info]}` where `path` is the parent containing the list of files with info. Therefore, although your method successfully caches the detail of the one file of interest, it very probably would have consequences for other operations on the same path or the parent.

monken: Do you have a suggestion how to resolve this issue? Assigning the `[info]` to `self._parent(path)` doesn't seem to be a good idea because we did not actually list that directory and it could contain additional files. I actually think there is little harm in doing it in the way I proposed.

martindurant: I am not totally convinced that it's worth the extra effort to prevent multiple HEAD calls. One auth has happened and a session is open, it should always be fast. Are you experiencing a high cost from it?

martindurant: > every range request (e.g., for an individual column) will trigger a head_object request

Ah, this is annoying, and I suppose this is the real problem. When making a range request, it shouldn't need to call `open()` at all, but use `cat`! I don't suppose we can change this upstream.

martindurant: Are you aware of https://developer.nvidia.com/blog/optimizing-access-to-parquet-data-with-fsspec/ , which should let you prospectively download and cache all the pieces you will be needing in one shot, and then arrow effectively loads from memory/local.

monken: I'll take a look. Thanks for sharing! The issue with `blockcache` seems to stem from https://github.com/fsspec/filesystem_spec/blob/2fbe8deff1c8d79f3125def9d930d19c213f03f6/fsspec/spec.py#L1605 where it requires the `size` property to create its cache.

martindurant: As a side issue, and sorry that I indulge myself: fastparquet might do a better job that arrow. I find, after all these years, it's still surprisingly often the better performing library, assuming you want pandas rather than actual arrow structures.

monken: Appreciate your advice! We are actually trying to move away from pandas towards polars so arrow is preferred. I'll give it a shot though!

Issue #781, opened 2023-09-04
[BUG] RuntimeError: Loop is not running


DanielTsiang:
```
Versions:

fsspec                        2023.9.0
s3fs                          2023.9.0
```

On my Sanic app startup, I'm getting the following warning message:
```python3
/opt/python/lib/python3.10/site-packages/s3fs/core.py:546: RuntimeWarning: coroutine 'ClientCreatorContext.__aexit__' was never awaited
  loop.create_task(s3.__aexit__(None, None, None))
```

Which I think is then what is leading to this exception later on:
```python3
RuntimeError: Loop is not running
```

The `RuntimeError` did not appear previously on version `2023.6.0`.

However, the error only seems to occur on app startup. The app seems to still work fine afterwards.

alicederyn:
I think the bug is because `create_task` is called without saving a strong reference to the task: https://github.com/fsspec/s3fs/blob/7dd0e368f3121f44ec215c870ca44206afaf68e6/s3fs/core.py#L570

From the docs for https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task:

> Important: Save a reference to the result of this function, to avoid a task disappearing mid-execution. The event loop only keeps weak references to tasks. A task that isn’t referenced elsewhere may get garbage collected at any time, even before it’s done.

martindurant:
Perhaps 
```
return loop.create_task(s3.__aexit__(None, None, None)) 
```
? It's not obvious where we should stash the task. If only [`call_soon()`](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.call_soon) or its threadsafe variant would handle coroutines.

alicederyn:
> Perhaps
> 
> ```
> return loop.create_task(s3.__aexit__(None, None, None)) 
> ```

Then your caller would need to solve the same problem :/

I'm not 100% sure if it's actually safe to call `loop.create_task` from a finalizer by the way -- I would suggest instead using `asyncio.run_coroutine_threadsafe`. I believe finalizers can occur between any two bytecodes, so you run the risk of corrupting state in your asyncio implementation, or deadlocking it.

How about having a private set in your module that you store tasks in? You can delete them with an `add_done_callback` cleanup step on the future returned by `run_coroutine_threadsafe`. You'd need to protect the set with a lock, to avoid races if users have multiple event loops on separate threads.

martindurant:
Would you like to code that suggestion into a draft PR?

[...]

PR #782, opened 2023-09-04
Await loop.create_task in core.py, fix for #781

DanielTsiang:
Await loop.create_task in core.py, fix for https://github.com/fsspec/s3fs/issues/781:
* https://github.com/fsspec/s3fs/issues/781

This should resolve the following warning message:
```python3
/opt/python/lib/python3.10/site-packages/s3fs/core.py:546: RuntimeWarning: coroutine 'ClientCreatorContext.__aexit__' was never awaited
  loop.create_task(s3.__aexit__(None, None, None))
```

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -543,7 +543,7 @@ class S3FileSystem(AsyncFileSystem):
         if loop is not None and loop.is_running():
             try:
                 loop = asyncio.get_event_loop()
-                loop.create_task(s3.__aexit__(None, None, None))
+                await loop.create_task(s3.__aexit__(None, None, None))
                 return
             except RuntimeError:
                 pass


martindurant:
`close_session` is not a coroutine, so you cannot `await` here. The best you could do is try to run the loop until completion, but this kind of waiting in a blocking context is exactly the problem we can't solve. As far as I understand, create_task should immediately schedule the call, and it will run so long as the loop will do at least one more tick.
If we did want to wait here, we would have to do something like `fsspec.asyn.sync`, where we mess around with thread `event` objects.

martindurant:
```
E     File "/home/runner/work/s3fs/s3fs/s3fs/core.py", line 546
E       await loop.create_task(s3.__aexit__(None, None, None))
E       ^
E   SyntaxError: 'await' outside async function
```

DanielTsiang:
Closing based on reply from @martindurant.

PR #784, opened 2023-09-05
Update core.py to take in charge default value of acl to `private`

yilas:
See ticket https://github.com/fsspec/s3fs/issues/779 for more details

In the [original code](https://github.com/fsspec/s3fs/blob/main/s3fs/core.py#L653) : 

```python
acl = acl or self.s3_additional_kwargs.get("ACL", "")
```

The value `acl` is set by default to `private` (see the arguments of the `_open` function) which means that the `acl` value must be set to something other than `None`. As a result, `self.s3_additional_kwargs.get("ACL", "")` is never evaluated (AFAIK 😄)

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -599,7 +599,7 @@ class S3FileSystem(AsyncFileSystem):
         path,
         mode="rb",
         block_size=None,
-        acl="private",
+        acl=None,
         version_id=None,
         fill_cache=None,
         cache_type=None,
@@ -650,7 +650,7 @@ class S3FileSystem(AsyncFileSystem):
         if requester_pays is None:
             requester_pays = bool(self.req_kw)
 
-        acl = acl or self.s3_additional_kwargs.get("ACL", "")
+        acl = acl or self.s3_additional_kwargs.get("ACL", "private")
         kw = self.s3_additional_kwargs.copy()
         kw.update(kwargs)
         if not self.version_aware and version_id:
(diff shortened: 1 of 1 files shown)


s3fs/core.py
```diff
@@ -599,7 +599,7 @@ def _open(
         path,
         mode="rb",
         block_size=None,
-        acl="private",
+        acl=None,
```
martindurant, line 602: I used False, but same idea. I think the key, though, is not to pass the argument at all when not specified, since S3 should pick whatever it thinks as default for that case.

s3fs/core.py
```diff
@@ -858,7 +858,8 @@ async def _find(self, path, maxdepth=None, withdirs=None, detail=False, prefix="
 
     find = sync_wrapper(_find)
 
-    async def _mkdir(self, path, acl="private", create_parents=True, **kwargs):
+    async def _mkdir(self, path, acl=None, create_parents=True, **kwargs):
+        acl = acl or self.s3_additional_kwargs.get("ACL", "private")
```
martindurant, line 862: The bucket ACLs and key ACLs aren't quite the same sets, so this could be invalid ("private" is fine)

martindurant (commented):
Please also see #785

s3fs/core.py
yilas: Not really the same because `acl` should be a string. `False` is a boolean.

martindurant: None is also not a string.

martindurant:
I think we can close this given #785

s3fs/core.py
martindurant: Updating the doc would be fine, although a falsy value is simply discarded. We are talking about my PR now, not this one, right?

PR #792, opened 2023-09-19
Remove CI warnings

ianthomas23:
Remove warnings in CI:

- Replace deprecated `mamba-org/provision-with-micromamba` with `mamba-org/setup-micromamba`.
- Specify `python-version` in `lint` action.
- Update versions of actions that are using deprecated `node12`.


ianthomas23:
Confirming that this has dealt with all GHA warnings.

PR #793, opened 2023-09-20
Admin: Improve moto usage, simplifying and speeding up tests

bblommers:
Hi there! Moto maintainer here. I had a look at the tests for `s3fs`, and realized that there is a cleaner way to use the Moto library for your usecase.

The ThreadedMotoServer starts an instance of MotoServer in the background, achieving the same result as calling `moto_server s3` in a subprocess. So the behaviour is the same, it just hides the nasty details like calling `subprocess.Popen`, and having to manually call `requests.get` to verify whether the server is ready.

The second change is to turn the `s3_base` fixture into a module-scoped fixture. The ThreadedMotoServer will now only be created once per module (file), meaning that the tests are executed much faster (~3mins vs ~5 mins).

Because we do expect a clean slate in between tests, I added `reset_s3_fixture` - calling the reset-API in Moto deletes all existing state before every test.

### DIFF
--- a/s3fs/tests/test_s3fs.py
+++ b/s3fs/tests/test_s3fs.py
@@ -13,6 +13,7 @@ import time
 import sys
 import pytest
 import moto
+from moto.moto_server.threaded_moto_server import ThreadedMotoServer
 from itertools import chain
 import fsspec.core
 from dateutil.tz import tzutc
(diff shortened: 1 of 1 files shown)


martindurant:
That's great, thanks for getting in touch and contributing, and general thanks for moto's existence :)

PR #795, opened 2023-09-21
Correct rmdir when not just a bucket

### LINKED ISSUE #689: s3fs rmdir command returns error when given a valid s3 path
Issue:
I have s3 uri that points to empty folder, i have verified the uri is valid by checking the path with fs.isdir and fs.info and both return valid responses. However when i go and use the rmdir command i get the following error:

Invalid bucket name: Bucket name must match the regex "^[a-zA-Z0-9.\-_]{1,255}$" or be an ARN matching the regex "^arn:(aws).*:(s3|s3-object-lambda):[a-z\-0-9]*:[0-9]{12}:accesspoint[/:][a-zA-Z0-9\-.]{1,63}$|^arn:(aws).*:s3-outposts:[a-z\-0-9]+:[0-9]{12}:outpost[/:][a-zA-Z0-9\-]{1,63}[/:]accesspoint[/:][a-zA-Z0-9\-]{1,63}$"

How is this function supposed to be used?

Version info:
Python 3.11.1
sf3s 2011.11.0

ianthomas23:
Fixes #689.

This implements the fix proposed in #689 which looks good to me.

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -919,6 +919,13 @@ class S3FileSystem(AsyncFileSystem):
     makedirs = sync_wrapper(_makedirs)
 
     async def _rmdir(self, path):
+        bucket, key, _ = self.split_path(path)
+        if key:
+            if await self._exists(path):
+                # User may have meant rm(path, recursive=True)
+                raise FileExistsError
+            raise FileNotFoundError
+
         try:
             await self._call_s3("delete_bucket", Bucket=path)
         except botocore.exceptions.ClientError as e:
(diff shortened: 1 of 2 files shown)


s3fs/core.py
```diff
@@ -919,6 +919,13 @@ async def _makedirs(self, path, exist_ok=False):
     makedirs = sync_wrapper(_makedirs)
 
     async def _rmdir(self, path):
+        bucket, key, _ = self.split_path(path)
+        if key:
+            if await self._exists(path):
+                # User may have meant rm(path, recursive=True)
+                raise FileExistsError
```
martindurant, line 926: By doing this, we are explicitly not dealing with directory placeholder files. Perhaps that should be stated in the docstring.

PR #797, opened 2023-09-21
Allow size= in open()

### LINKED ISSUE #794: Support `size=` in `S3File`
Currently the base `AbstractBufferedFile` class supports a `size=` keyword, but `s3fs` doesn't take advantage of it. It'd be nice if we did so we could accessing S3 unnecessarily 

cc @martindurant, who I was on a call earlier today with about this

martindurant:
Fixes #794 

@jrbourbeau , do you want to test? I'll think about how to test this in CI.

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -604,6 +604,7 @@ class S3FileSystem(AsyncFileSystem):
         fill_cache=None,
         cache_type=None,
         autocommit=True,
+        size=None,
         requester_pays=None,
         cache_options=None,
         **kwargs,
@@ -680,6 +681,7 @@ class S3FileSystem(AsyncFileSystem):
             autocommit=autocommit,
             requester_pays=requester_pays,
             cache_options=cache_options,
+            size=size,
         )
 
     async def _lsdir(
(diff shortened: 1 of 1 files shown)


jrbourbeau:
Thanks @martindurant! I'll test this out

Naively, looking at the PR diff, we may still be triggering an `s3.info` call here https://github.com/fsspec/s3fs/pull/797/files#diff-044e3dd45e47667275a210a6ac0033c31687eaea7d4e33e3b21b34baa1a050daR2083 (I could be wrong though)

martindurant:
That's only for buckets with versioning. I don't suppose I need to change too many code paths in one go here, but I has better yet add the docstring for size=

martindurant:
```
# after any initial call which does SSL and checks credentials. 
In [4]: %time f = fs.open("mymdtemp/icesat2-4.01.json")
CPU times: user 4.07 ms, sys: 1.76 ms, total: 5.83 ms
Wall time: 57.3 ms

In [5]: fs.invalidate_cache()

In [6]: %time f = fs.open("mymdtemp/icesat2-4.01.json", size=100)
CPU times: user 110 µs, sys: 0 ns, total: 110 µs
Wall time: 113 µs
```

jrbourbeau (approved):
Thanks for this update @martindurant, I'm also seeing performance improvements when opening with `size=` specified

PR #802, opened 2023-09-27
bump requirements.txt

jayzfbn:
referring https://github.com/fsspec/s3fs/issues/801


martindurant:
That's all it should take, and everything passes, so good to go.

PR #805, opened 2023-10-09
fix bugs in _get_file

troychiu:
I fix some bugs in _get_file.

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -1217,7 +1217,7 @@ class S3FileSystem(AsyncFileSystem):
                 Bucket=bucket,
                 Key=key,
                 **version_id_kw(version_id or vers),
-                **self.req_kw,
+                **kw,
             )
             return resp["Body"], resp.get("ContentLength", None)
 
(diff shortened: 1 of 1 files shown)


martindurant:
Did you have some specific problems that led to this PR? The req_kw change looks right, but I am uncertain about the Range. Do you think tests can be made?

troychiu:
I would love to write some test to validate this pr, but I am having trouble in how to simulate the fail of getting files. Do you have any idea?

martindurant:
Perhaps if you detail how you noticed the bug, we could go from there? Did you have a specific workflow that went wrong?

troychiu:
I found the bug when I was tracing the implementation, so there is not specific workflow that trigger the bug. I have done some research on simulating the error, but didn't find a effective way.

troychiu:
If we merge this section of code and meet any problem, I will spend time looking into them.
But I am also fine with deleting it.

martindurant:
I have convinced myself that the coe here is correct, and we can think about how to maybe test in a follow up

troychiu:
I agree with you that the tests should be done.

LukasBommes:
I can confirm the correctness of the fixes in thir PR. We spend the entire day debugging a problem related to this. Due to hitting [this bug in aiohttp](https://github.com/aio-libs/aiohttp/issues/4581) during `body.read` (in rare cases) the exception block was entered and here we noticed that upon retry the stream was read from the beginning instead of from the last byte read up to the exception. We fixed the issue exactly like in this PR. Obviously, we were on an older version of s3fs (2023.9.2) prior to this fix.

PR #818, opened 2023-11-03
return offset-aware modification time

### LINKED ISSUE #817: Return offset-aware datetimes
`s3fs` returns offset-naive datetimes, whereas https://github.com/fsspec/filesystem_spec/pull/1314 introduced datetimes at UTC in `fsspec`. 

I often get errors because operations mixing offset-naive and offset-aware datetimes are not allowed.
Here is an example:
```python
fsspec.filesystem("file").modified(...) - fsspec.filesystem("s3").modified(...)
# TypeError: can't subtract offset-naive and offset-aware datetimes
```

Would it be possible to introduce offset-aware datetimes in `s3fs` as well?

malmans2:
Closes #817

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -1981,7 +1981,7 @@ class S3FileSystem(AsyncFileSystem):
         if "LastModified" not in info:
             # This path is a bucket or folder, which do not currently have a modified date
             raise IsADirectoryError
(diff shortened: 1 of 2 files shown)


martindurant:
Merging this, having decided that keeping information should always be preferable (even though the timezone is always UTC).

PR #828, opened 2023-11-29
Use .get for LastModified to mitigate KeyError

pypeaday:
This PR addresses the issue where a KeyError is raised because the `LastModified` attribute is not on a retrieved object

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -1309,7 +1309,7 @@ class S3FileSystem(AsyncFileSystem):
                 )
                 return {
                     "ETag": out.get("ETag", ""),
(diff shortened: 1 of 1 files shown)


s3fs/core.py
```diff
@@ -1309,7 +1309,7 @@ async def _info(self, path, bucket=None, key=None, refresh=False, version_id=Non
                 )
                 return {
                     "ETag": out.get("ETag", ""),
-                    "LastModified": out["LastModified"],
+                    "LastModified": out("LastModified", ""),
```
martindurant, line 1312:
```suggestion
                    "LastModified": out.get("LastModified", ""),
```

PR #829, opened 2023-11-30
Relax aiobotocore versions to < 3.0.0

bolkedebruin:
This relaxes the accepted aiobotocore versions
to < 3.0.0 and provides a testing matrix against
all recent knowmn versions, that can be extended
at will.

I have been a bit daring and made it < 3.0.0 as the 
whole matrix passed. 

The CI will provide information on what exact minor version of 
aiobotocore is tested against including what boto3. As 
boto3 is synced there is some backtracking done by pip,
but I kept that as it is. We could provide some hints if required
but I thought it was clearer this way.

@martindurant @freddy-fostvedt @lou-k @manugarri

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,12 +4,13 @@ on: [push, pull_request]
 
 jobs:
   test:
-    name: Python ${{ matrix.python-version }}
+    name: Python ${{ matrix.python-version }} - AioBotocore ${{ matrix.aiobotocore-version }}
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
         python-version: ["3.8", "3.9", "3.10", "3.11"]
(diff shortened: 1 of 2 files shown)


lou-k (approved):
👏👏👏👏

Thank you!

manugarri:
awesome, thanks for the help @bolkedebruin :)

martindurant:
It is too many checks to add in py3.12 too? That one would only need the most recent aiobotocore, I think.

bolkedebruin:
Let' see what it does against older version - just to keep things simple

martindurant:
We may trim the number of runs in the future if we start having to wait too long for CI to finish. But this will do for now.

bolkedebruin:
Awesome. Thx!

potiuk:
Cool ! Thanks for that @bolkedebruin @martindurant :). That will save us quite a lot of headaches for Airflow

Issue #831, opened 2023-12-04
Unable to authenticate with web identity


kwohlfahrt:
I am trying to authenticate to S3 with a web identity token file, as created by the [AWS pod identity webhook](https://github.com/aws/amazon-eks-pod-identity-webhook).

The following snipped reproduces the issue:

```python3
import s3fs
fs = s3fs.S3FileSystem()
fs.ls("foo") # Fails with `AttributeError`
```

This is the traceback of the error:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.12/site-packages/fsspec/asyn.py", line 118, in wrapper
    return sync(self.loop, func, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/fsspec/asyn.py", line 103, in sync
    raise return_result
  File "/usr/local/lib/python3.12/site-packages/fsspec/asyn.py", line 56, in _runner
    result[0] = await coro
                ^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/s3fs/core.py", line 1023, in _ls
    files = await self._lsdir(path, refresh, versions=versions)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/s3fs/core.py", line 744, in _lsdir
    async for c in self._iterdir(
  File "/usr/local/lib/python3.12/site-packages/s3fs/core.py", line 775, in _iterdir
    await self.set_session()
  File "/usr/local/lib/python3.12/site-packages/s3fs/core.py", line 507, in set_session
    self.key = credentials.access_key
               ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/botocore/credentials.py", line 409, in access_key
    self._refresh()
  File "/usr/local/lib/python3.12/site-packages/botocore/credentials.py", line 502, in _refresh
    self._protected_refresh(is_mandatory=is_mandatory_refresh)
  File "/usr/local/lib/python3.12/site-packages/botocore/credentials.py", line 518, in _protected_refresh
    metadata = self._refresh_using()
               ^^^^^^^^^^^^^^^^^^^^^
[...]

PR #832, opened 2023-12-04
Don't assume attributes after connect

martindurant:
cc https://github.com/fsspec/s3fs/issues/831

@kwohlfahrt , can you try with this, please?


kwohlfahrt:
Yes, this fixes #831 for me.

PR #840, opened 2023-12-14
Fix the cache lookup in _info

YeahItsMeAgain:
The info function retrieved the `fullpath` from the cache in its initial lookup but used the given `path` (which is a user given variable) on the second instance.

Fixed it to use `fullpath` both times.

You can reproduce the bug with the following code:
```python
# Filling the cache.
fs.isdir("/bucket")

# The cache seems ok.
print(fs.dircache._cache)

# The cache "miss" happens now because the key doesn't start with "/".
print(fs.info("/bucket/file"))
```

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -1293,7 +1293,7 @@ class S3FileSystem(AsyncFileSystem):
                     if out:
                         return out[0]
                 else:
-                    out = [o for o in out if o["name"] == path]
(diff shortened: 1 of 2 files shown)


martindurant:
Since you already have a test in the PR description, can you please add this to the code to make sure we don't accidentally undo this sometime?

YeahItsMeAgain:
Sure, I changed the existing test :)

PR #849, opened 2024-02-13
Update core.py

obronco:
Refactor the code to improve mime type guessing by utilizing the destination file instead of relying on the source file. This change was prompted by the inconvenience of having uncompressed Minio objects when the source was a .tmp file that I would subsequently copy to .txt or .csv format.

Based on the observation, it seems there shouldn't be any drawbacks to this change, as the scenarios considered are:

1. Local files with unknown extensions being transferred to remote locations with known extensions like .txt, .csv, .json, or .zip.
2. Cases where both local and remote files have the same known extension.

The question arises: Does anyone currently utilize the scenario where local files have known extensions while remote files have unknown extensions?

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -1153,7 +1153,7 @@ class S3FileSystem(AsyncFileSystem):
         callback.set_size(size)
 
         if "ContentType" not in kwargs:
-            content_type, _ = mimetypes.guess_type(lpath)
+            content_type, _ = mimetypes.guess_type(rpath)
             if content_type is not None:
                 kwargs["ContentType"] = content_type
 


s3fs/core.py
```diff
@@ -1153,7 +1153,7 @@ async def _put_file(
         callback.set_size(size)
 
         if "ContentType" not in kwargs:
-            content_type, _ = mimetypes.guess_type(lpath)
+            content_type, _ = mimetypes.guess_type(rpath)
```
martindurant, line 1156: This would preserve current behaviour if indeed there are cases where the remote path is extensionless.
```suggestion
            content_type = mimetypes.guess_type(rpath)[0] or mimetypes.guess_type(lpath)[0]
```

martindurant:
Can we have a test for this please, now with an inferable path name for local, remote, and both.

PR #863, opened 2024-03-17
fix(core): accept kwargs in get file

shcheklein:
Fixes https://github.com/iterative/dvc-s3/issues/80

It is similar to `gcsfs` and `adlfs`.

On our end it seems `max_concurrency` is passed here https://github.com/iterative/dvc-objects/blob/main/src/dvc_objects/fs/generic.py#L210 and since the new version has this attr we pass it now, which most likely leads to this error.

I'm not sure why `_gef_file` part was not yet implemented. @pmrowla might have a better idea on was it complicated, or less priority. It seems it is the natural next step to do so.

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -1271,7 +1271,7 @@ class S3FileSystem(AsyncFileSystem):
         return out
 
     async def _get_file(
-        self, rpath, lpath, callback=_DEFAULT_CALLBACK, version_id=None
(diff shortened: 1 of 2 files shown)


pmrowla:
For reference, concurrency for `_get_file` was not implemented yet since re-assembling out-of-order chunks is more complex than the `_put_file` case (where it is handled by the server when using multipart uploads)

PR #865, opened 2024-03-21
fix(s3fs): Consolidate retry logic for rate limits

pingsutw:
Retry the request when we get the error (reduce your request rate).

We have 2000+ tasks, each running in the K8s pod and uploading multiple files to a certain bucket. We get `reduce your request rate` error because s3fs doesn't retry this request

https://stackoverflow.com/questions/46787072/s3-slowdown-please-reduce-your-request-rate-exception

image

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -118,10 +118,13 @@ async def _error_wrapper(func, *, args=(), kwargs=None, retries):
         except ClientError as e:
             logger.debug("Client error (maybe retryable): %s", e)
             err = e
+            wait_time = asyncio.sleep(min(1.7**i * 0.1, 15))
             if "SlowDown" in str(e):
-                await asyncio.sleep(min(1.7**i * 0.1, 15))
+                await wait_time
+            elif "reduce your request rate" in str(e):
+                await wait_time
             elif "XAmzContentSHA256Mismatch" in str(e):
-                await asyncio.sleep(min(1.7**i * 0.1, 15))
+                await wait_time
             else:
                 break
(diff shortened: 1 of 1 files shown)


pingsutw:
@martindurant mind taking a look? It will be great to cut a release once we merge it. We have some production workflow blocked by this.

martindurant:
The specific change here is to add "reduce your request rate" as a retriable condition? For reference, can you put details of the exception here, since I don't think this is one we've seen before.

s3fs/core.py
```diff
@@ -118,10 +118,13 @@ async def _error_wrapper(func, *, args=(), kwargs=None, retries):
         except ClientError as e:
             logger.debug("Client error (maybe retryable): %s", e)
             err = e
+            wait_time = asyncio.sleep(min(1.7**i * 0.1, 15))
```
martindurant, line 121: I'm pretty sure you will see a warning if you create this coroutine, but none of the following conditions apply, so you never wait on it

pingsutw:
image

s3fs/core.py
pingsutw: sorry, right, let me update it real quick

pingsutw:
@martindurant does it make sense

pingsutw:
@martindurant sorry to ping you again. qq: do we have an ETA for release?

martindurant:
Not certain yet, there are a few things in flight. There's always at least one release per month.

pingsutw:
Got it, thank you!

PR #866, opened 2024-03-25
_bulk_delete must always return a list, even when pathlist is empty

ghislainp:
Fixes https://github.com/fsspec/filesystem_spec/issues/1557

_bulk_delete returned None (no return) when pathlist is empty which is incompatible with the list returned in normal operation. This causes a bug in fsspec for instance, because a list was expected. This change returns an empty list when pathlist is empty

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -1954,7 +1954,7 @@ class S3FileSystem(AsyncFileSystem):
             Must have 0 < len <= 1000
         """
         if not pathlist:
-            return
+            return []
         buckets = {self.split_path(path)[0] for path in pathlist}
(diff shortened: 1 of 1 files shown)


martindurant:
Would calling bulk_delete with an empty list have triggered the error? If so, we can use that as a test.

ghislainp:
no unfortunately, my problem is coming from elsewhere... 

I have just tried calling fs.rm([], recursive=True) but it raise an exception file not found... an expected behavior I believe.

martindurant:
OK, I understand why this is correct - but I don't know why the situation occurs in the first place.

PR #884, opened 2024-06-17
Request Id in Errors

EngHabu:


martindurant:
This is a reasonable idea, please let me know when you are finished. Note that moto (the CI backend) probably has a different idea of request ID compared to AWS S3, and of course there are several other S3 implementations (e.g., minio) that might have something different again.

EngHabu:
Thank you for your quick comment, @martindurant.
Looks like moto uses the same s3 response structure for RequestId: https://github.com/getmoto/moto/blob/master/tests/test_s3/test_s3_metadata.py#L20-L25 ([added here](https://github.com/getmoto/moto/pull/3836/files))

I don't seem to find request ids in minio responses (sad). Not in the body and nor in the response headers.
image
image

EngHabu:
@martindurant mind taking a look?

martindurant:
This could use a test; any error on moto should have both the status and ID set (non-empty). I wonder whether instead of extending the message, we should add the information to the exception object, so it only shows up for those who want it. What do you think?

EngHabu:
I briefly looked into that as well because I would like to be able to extract that information separately. But it uses the native OSError type and not sure how you feel about subclassing that?

s3fs/tests/test_s3fs.py
```diff
@@ -98,7 +98,7 @@ def get_boto3_client():
 
     # NB: we use the sync botocore client for setup
     session = Session()
-    return session.create_client("s3", endpoint_url=endpoint_uri)
+    return session.create_client("s3", endpoint_url=endpoint_uri, region_name="us-east-1")
```
martindurant, line 101: Why is this necessary?

martindurant:
Sorry I let this drop.

> But it uses the native OSError type and not sure how you feel about subclassing that?

Yes, that's totally since, since `except` matches subclasses.

PR #888, opened 2024-08-01
Add support for cloudflare's R2 storage

### LINKED ISSUE #789: Incorrect MultiPartUpload Chunksize
**Description:**

I have encountered an issue with the MultiPartUpload functionality in the s3fs library where the chunk size used during upload appears incorrect. 

**Environment Information:**

- s3fs Version: 2023.9.1
- Python Version: 3.9
- Operating System: Ubuntu 22.04

**Issue Details:**

When using s3fs to upload large files to an S3 bucket on Cloudflare R2, I noticed that the chunk size used for MultiPartUpload needs to be consistent. 

**Expected Behavior:**

I expect s3fs to use the same chunk size for files with the body` (b"1" * 5 * 2**30) + b"kek" ` and `b"1" * 5 * 2**30`

**Steps to Reproduce:**

Working example:
```
import io

with fsspec.open('s3://path', mode='wb') as f:
    buffer = io.BytesIO(b"1" * 5 * 2**30)
    f.write(buffer.getvalue())
```

This code works perfectly.

But if we change the buffer size to:

```
with fsspec.open('s3://path', mode='wb') as f:
    buffer = io.BytesIO((b"1" * 5 * 2**30) + b"kek" )
    f.write(buffer.getvalue())
```
This code gives the error: 
**ClientError: An error occurred (InvalidPart) when calling the CompleteMultipartUpload operation: All non-trailing parts must have the same length.**

Many thanks for considering my request.

arogozhnikov:
fix #789 (also see for previous discussion)

Cloudflare's R2 storage has an additional restriction during multi-part uploads that all chunks are of the same size, with the exception of the last one. Last can be smaller than previous. I don't think this behavior is well documented, but I've checked it empirically.

This unfortunately means that clever merging of last chunk used by s3fs can't be used, and I suggest this simple implementation (I _guess_ by default boto follows the same chunking logic).

I have tested this to work by uploading/downloading to S3 and R2 random files of lengths: [0, 1, 10, 1024, 6_000_000, 12_000_000, 123_456_789]

BTW @martindurant thanks again for the lib.

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -2331,41 +2331,31 @@ class S3File(AbstractBufferedFile):
             and self.tell() < self.blocksize
         ):
             # only happens when closing small file, use on-shot PUT
-            data1 = False
+            pass
         else:
             self.buffer.seek(0)
-            (data0, data1) = (None, self.buffer.read(self.blocksize))
 
-        while data1:
-            (data0, data1) = (data1, self.buffer.read(self.blocksize))
-            data1_size = len(data1)
+            # previously last small chunk was merged into previous one 
+            # (and in some cases two chunks were equalized)
+            # However, R2 imposes additional restriction that all but last chunk
+            # are of same size, and last chunk can't be larger.
+            while data := self.buffer.read(self.blocksize):
+                part = len(self.parts) + 1
+                logger.debug("Upload chunk %s, %s" % (self, part))
 
-            if 0 < data1_size < self.blocksize:
-                remainder = data0 + data1
-                remainder_size = self.blocksize + data1_size
+                out = self._call_s3(
+                    "upload_part",
+                    Bucket=bucket,
+                    PartNumber=part,
+                    UploadId=self.mpu["UploadId"],
+                    Body=data,
+                    Key=key,
+                )
 
-                if remainder_size <= self.part_max:
-                    (data0, data1) = (remainder, None)
-                else:
-                    partition = remainder_size // 2
-                    (data0, data1) = (remainder[:partition], remainder[partition:])
-
-            part = len(self.parts) + 1
-            logger.debug("Upload chunk %s, %s" % (self, part))
-
-            out = self._call_s3(
-                "upload_part",
-                Bucket=bucket,
-                PartNumber=part,
-                UploadId=self.mpu["UploadId"],
-                Body=data0,
-                Key=key,
-            )
-
-            part_header = {"PartNumber": part, "ETag": out["ETag"]}
-            if "ChecksumSHA256" in out:
-                part_header["ChecksumSHA256"] = out["ChecksumSHA256"]
-            self.parts.append(part_header)
+                part_header = {"PartNumber": part, "ETag": out["ETag"]}
+                if "ChecksumSHA256" in out:
+                    part_header["ChecksumSHA256"] = out["ChecksumSHA256"]
+                self.parts.append(part_header)
 
         if self.autocommit and final:
             self.commit()


s3fs/core.py
```diff
@@ -2331,41 +2331,31 @@ def _upload_chunk(self, final=False):
...
-                    partition = remainder_size // 2
-                    (data0, data1) = (remainder[:partition], remainder[partition:])
 
-            part = len(self.parts) + 1
-            logger.debug("Upload chunk %s, %s" % (self, part))
+            # previously last small chunk was merged into previous one 
+            # (and in some cases two chunks were equalized)
+            # However, R2 imposes additional restriction that all but last chunk
+            # are of same size, and last chunk can't be larger.
+            while data := self.buffer.read(self.blocksize):
```
martindurant, line 2342: The first use of a walrus in this repo. Remind me, which version of python does this require?

arogozhnikov: 3.8 (see https://docs.python.org/3/whatsnew/3.8.html)

arogozhnikov:
I'll fix whitespace reported by pre-commit. 

@martindurant do you know what causes  "Your proposed upload is smaller than the minimum allowed object size." ?

martindurant:
There are test failures with "Your proposed upload is smaller than the minimum allowed object size".

martindurant:
In https://docs.aws.amazon.com/AmazonS3/latest/userguide/qfacts.html 

> 
> Part size | 5 MiB to 5 GiB. There is no minimum size limit on the last part of your multipart upload.

So apparently a non-last part is <5MB. We should *not* be reading the whole of the file internal buffer if there are less than blocksize bytes left and `final is False` - they should stay in the buffer for next time.


Actually, there is a change of behaviour here (I think). Previously, _uload_chunk would send as much data as it has whenever write()/flush() happens. Always splitting this into blocksize chunks (5MB!) may cause many more calls.

arogozhnikov:
@martindurant please take a look at updated version.

> Always splitting this into blocksize chunks (5MB!) may cause many more calls.

That's correct. If we want to support R2, we need to make a decision about chunk size beforehand. What's your recommendation?

> We should not be reading the whole of the file internal buffer if there are less than blocksize bytes left 

Is there a reliable way to check this? I guess this is tell/seek you talk about, but not all file-likes have them.

martindurant:
> That's correct. If we want to support R2, we need to make a decision about chunk size beforehand. What's your recommendation?

It sounds like it might need to be a filesystem-wide optional flag between the old behaviour and this new one.

martindurant:
> Is there a reliable way to check this? I guess this is tell/seek you talk about, but not all file-like objects have them.

The internal buffer is an io.BytesIO, so len(buffer.getbuffer()) is a zero-cost way to get the size.

arogozhnikov:
> It sounds like it might need to be a filesystem-wide optional flag between the old behaviour and this new one.

how about a flag `fixed_upload_part_size: int`?

arogozhnikov:
@martindurant updated suggested implementation.

Note: I fetch up to part_max bytes from input now.

s3fs/core.py
```diff
@@ -296,6 +296,7 @@ def __init__(
         asynchronous=False,
         loop=None,
         max_concurrency=1,
+        fixed_upload_size: int | None = None,
```
martindurant, line 299: Could i be a bool parameter instead, so that we have only one "size" to worry about?

arogozhnikov: done

s3fs/core.py
```diff
@@ -296,6 +296,7 @@ def __init__(
         asynchronous=False,
         loop=None,
         max_concurrency=1,
+        fixed_upload_size: bool = False,
```
martindurant, line 299: Could use a docstring description, probably explaining that this is required for writing files to R2.

s3fs/core.py
```diff
@@ -2331,41 +2333,41 @@ def _upload_chunk(self, final=False):
...
+                return len(self.buffer.getbuffer()) - self.buffer.tell()
 
-            part_header = {"PartNumber": part, "ETag": out["ETag"]}
-            if "ChecksumSHA256" in out:
-                part_header["ChecksumSHA256"] = out["ChecksumSHA256"]
-            self.parts.append(part_header)
+            min_chunk = 1 if final else self.blocksize
+            if self.fs.fixed_upload_size:
+                # all chunks have fixed size, exception: last one can be smaller
+                while n_bytes_left() >= min_chunk:
```
martindurant, line 2366:
```suggestion
                while n_bytes_left() >= self.blocksize:
```
?

arogozhnikov: won't cover last chunk if final=True

s3fs/core.py
arogozhnikov: sure

s3fs/core.py
martindurant: But it *will* include last part of the buffer, with size < block_size on each call to flush(), so if blocksize is 10000, 
```
f.write(11000) # uploads 10000 and 1000
f.write(11000) # fails, block size mismatch because of preceding 1000-byte block
```

arogozhnikov: unless I'm missing something, in you example:

```
f.write(11000)
# final = false -> min_chunk=10_000, uploads 10_000 and lefts 1_000 in buffer
f.write(11000)
# final = True (for example) -> min_chunk=1, uploads 10_000 and then final chunk 2_000

```

martindurant: Lets make a test that does exactly this (with fixed_upload_size=True) and make sure it passes, and 1000 is in the buffer between the two calls.

arogozhnikov: @martindurant  wrote, test fails. 

Now I'm lost about your previous messages - you recommended not reading last chunk, and I was under impression that it is carried over, but apparently it isn't.

Does it mean I should go back to implementation that adds a buffer to s3file to keep data that wasn't sent?

martindurant: The file object has a buffer; the return value of the upload method tells the caller whether to empty the buffer or not. In your case, you probably want to _replace_ the buffer you just read from with whatever was not yet sent because it's not big enough.

martindurant: Let me know if you need more information about this.

arogozhnikov:
@martindurant please have a look at updated version, it manually creates new buffer. I've also added a test that we discussed.

It isn't clear from fsspec what `self.offset` is supposed to mean, so I interpreted it as 'how many bytes were already uploaded'.

martindurant:
> It isn't clear from fsspec what self.offset is supposed to mean, so I interpreted it as 'how many bytes were already uploaded'.

Exactly.

s3fs/tests/test_s3fs.py
```diff
@@ -2749,3 +2752,32 @@ def test_bucket_versioning(s3):
...
+        client_kwargs={"endpoint_url": endpoint_uri},
+        fixed_upload_size=True,
+    )
+    s3_fixed.invalidate_cache()
+    yield s3_fixed
+
+
+def test_upload_parts(s3_fixed_upload_size):
+    with s3_fixed_upload_size.open(a, "wb", block_size=6_000_000) as f:
+        f.write(b" " * 6_001_000)
```
martindurant, line 2770: Can we please make sure that one or more of these writes are not exact multiples of the upload chunk size, so that something is left over when the file is closed.

arogozhnikov: I've added a new test below with two different prime pads, should cover various cases of size divisibility.

arogozhnikov:
ping, LMK if you want to add anything (also feel free to edit/rewrite PR the way you like)

arogozhnikov:
feel free to make these/any other changes yourself, Github will allow you to push to this branch

geekodour:
@martindurant is it possible to use this change without a release  happening? I want to use this change but some dependencies use a previous version, I am not even able to install directly from git.

Eg. `Because datasets (2.20.0) depends on fsspec[http] (>=2023.1.0,<=2024.5.0) `

martindurant:
Assuming `datasets` doesn't actually have conflicting calls with the current codebase, you can always install s3fs with `--no-deps` (I think all the installers accept this flag).
If `datasets` actually tests the package metadata, you could fool it with
```
pip install datasets # grabs old version of s3fs
git clone https://github.com/fsspec/s3fs/
cd s3fs
git checkout 2024.3.1
pip install -e .
git checkout main
```
pip will think it has 2024.3.1, but will use the current code. Of course, you might want to do the same with fsspec too.

Note that I intend to make a release perhaps yet this week.

arogozhnikov:
@geekodour 
we use this in requirements file:

```
s3fs @ git+https://github.com/fsspec/s3fs@def207eef16607643dcdaf02d2dc01b439ccfc8a
```

but formal release would be much nicer :)

martindurant:
> formal release would be much nicer

This is in 2024.10.0, no? Or do you mean an update to the `datasets` requirements?

PR #898, opened 2024-10-08
Test with Python 3.13 and add PyPI trove classifiers

edgarrmondragon:


.github/workflows/ci.yml
```diff
@@ -27,7 +33,7 @@ jobs:
         uses: conda-incubator/setup-miniconda@v3
         with:
           environment-file: ci/env.yaml
-          python-version: ${{ matrix.PY }}
+          python-version: ${{ matrix.python-version }}
```
edgarrmondragon, line 36: This was previously causing all tests to run on `conda-incubator/setup-miniconda`'s default Python (i.e. 3.12)

martindurant: oops

edgarrmondragon:
FWIW the 3.13 tests are 🟢 on my fork: https://github.com/edgarrmondragon/s3fs/actions/runs/11228720331/job/31213080263

martindurant:
py3.13 has apparently come out, but only for `defaults` so far

PR #912, opened 2024-11-07
Parse the query parameter of url without using infer_storage_options

wang2bo2:
Parsing S3 access point URI, like `s3://arn:aws:s3:us-west-2:1234:accesspoint/abc/123.jpg`, causes problems in `fsspec.utils.infer_storage_options` because the `:accesspoint` was parsed as port number.

This patch removes the need to call `fsspec.utils.infer_storage_options` in s3fs, as discussed in https://github.com/fsspec/filesystem_spec/pull/1743#issuecomment-2460232401

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -14,7 +14,7 @@ from urllib3.exceptions import IncompleteRead
 
 import fsspec  # noqa: F401
 from fsspec.spec import AbstractBufferedFile
-from fsspec.utils import infer_storage_options, tokenize, setup_logging as setup_logger
+from fsspec.utils import tokenize, setup_logging as setup_logger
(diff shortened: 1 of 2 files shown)


martindurant:
For an example URL like this, does the bucket end up being "arn:aws:s3:us-west-2:1234:accesspoint"?

We should have a test ensuring that this does the right thing.

martindurant:
> There’s already a test

but the codepath you fixed is for url_to_fs, not split_path, right? That test was already passing without this change.

wang2bo2:
The added test would check `S3FileSystem`'s compatibility with `fsspec.url_to_fs` and this test would fail without the fix in this PR.

PR #939, opened 2025-02-11
create new session if refresh is True

hutch3232:
One observation is that `kwargs` is an unused argument. Should I remove it? Or do something else to handle it?

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -480,9 +480,26 @@ class S3FileSystem(AsyncFileSystem):
 
     async def set_session(self, refresh=False, kwargs={}):
         """Establish S3 connection object.
+
+        ``connect`` is an alias of ``set_session``. This method called with
+        ``refresh=True`` is useful if new credentials have been created and
+        the instance needs to be reestablished.
+
+        Parameters
+        ----------
+        refresh : bool (False)
+            If True, create a new session even if one already exists.
+        kwargs : dict
+
         Returns
         -------
         Session to be closed later with await .close()
+
+        Examples
+        --------
+        >>> s3 = S3FileSystem(profile="<profile name>")  # doctest: +SKIP
+        >>> s3.set_session()  # doctest: +SKIP
+        >>> s3.connect(refresh=True)  # doctest: +SKIP
         """
         if self._s3 is not None and not refresh:
             return self._s3
@@ -522,7 +539,7 @@ class S3FileSystem(AsyncFileSystem):
             config_kwargs["signature_version"] = UNSIGNED
 
         conf = AioConfig(**config_kwargs)
-        if self.session is None:
+        if self.session is None or refresh:
             self.session = aiobotocore.session.AioSession(**self.kwargs)
 
         for parameters in (config_kwargs, self.kwargs, init_kwargs, client_kwargs):


s3fs/core.py
```diff
@@ -480,9 +480,26 @@ def _prepare_config_kwargs(self):
...
+        kwargs : dict
+
         Returns
         -------
         Session to be closed later with await .close()
+
+        Examples
+        --------
+        >>> s3 = S3FileSystem(profile="<profile name>")  # doctest: +SKIP
+        >>> s3.set_session()  # doctest: +SKIP
```
martindurant, line 501: This method is actually async, so you would only call it with `await` within a coroutine, as happens elsewhere in the class. The exact alias is `_connect` (with the underscore), whereas `connect` (without the underscore) is the blocking version.

s3fs/core.py
```diff
@@ -480,9 +480,26 @@ def _prepare_config_kwargs(self):
 
     async def set_session(self, refresh=False, kwargs={}):
         """Establish S3 connection object.
+
+        ``connect`` is an alias of ``set_session``. This method called with
```
martindurant, line 484: This method is called by any operation on an s3fs instance. 

You might call this directly from async code, so that you can assign the client object to a local variable
(^ this is implied in the documentation on async, but should no longer be important for most people).

martindurant (approved):
I have a couple of small comments on wording.

On the kwargs: this is probably a leftover from previous code. It could be removed, but let's not do that in this PR.

s3fs/core.py
hutch3232: Thank you for the clarifications. I haven't directly worked with `async` before and overlooked that. Let me know if you have additional suggestions for my edits.

martindurant:
I suppose a test would be nice for completeness: call connect(), and assert that the ._s3 and .session attributes have both changed. I wonder if this will cause warnings that those resources weren't closed, or if garbage collection will take care of it.

hutch3232:
> I wonder if this will cause warnings that those resources weren't closed, or if garbage collection will take care of it.

I added a test that s3 and session has changed but I'm too sure how to test that the resources are closed.

This line seems to be working:
https://github.com/fsspec/s3fs/blob/744d9c65eb461eeb018c6b6a541f166a7bdd4ffc/s3fs/core.py#L557

I (locally) added a print message in here:
```python
print(f"closing session for {s3}!!!\n")
```
https://github.com/fsspec/s3fs/blob/744d9c65eb461eeb018c6b6a541f166a7bdd4ffc/s3fs/core.py#L566-L567

And when I run `pytest` it printed:
```
closing session for <aiobotocore.session.ClientCreatorContext object at 0x7f2f21b4dd00>!!!

closing session for <aiobotocore.session.ClientCreatorContext object at 0x7f2f22c0d1c0>!!!
```

martindurant:
> I (locally) added a print message

That will do, thank you.

If they were *not* getting closed, it would result in warnings at the end of the test run. I did a merge to main, to make sure all tests would run and will check for warnings when done.

martindurant:
There is one warning of the correct type, but because this is involving zarr, I don't think this PR is implicated. Will merge.
```
s3fs/tests/test_s3fs.py::test_with_xzarr
  /usr/share/miniconda/envs/test/lib/python3.10/asyncio/base_events.py:674: RuntimeWarning: coroutine 'ClientCreatorContext.__aexit__' was never awaited
    self._ready.clear()
```

PR #940, opened 2025-02-12
re-enable all tests in CI

hutch3232:
Perhaps this is intentional (if so feel free to close this PR) but it looks like only one test is being selected to run with the rest being skipped. Suspect that is not wanted.


martindurant:
Seems to do exactly the right thing.

PR #944, opened 2025-03-05
Add **kwargs to cat_file, mirroring pipe_file.

asford:
Extend cat_file interface with kwargs to match pipe_file, allowing conditional `IfMatch` passed into s3 requests.


martindurant:
(please let me know when you are done)

PR #954, opened 2025-03-28
Replace get_event_loop by get_running_loop

kafai-lam:
Since `asyncio.get_event_loop` was deprecated since version 3.12, using `asyncio.get_running_loop()` was preferred

Ref: https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_event_loop


martindurant:
OK, I agree. The previous version left open the possibility of making a new loop during shutdown, which would not be a good idea anyway.

PR #958, opened 2025-04-01
Update requirements.txt

### LINKED ISSUE #957: fsspec==2025.3.2.* does not exist
Dependency `fsspec==2025.3.2.*` in `requirements.txt` does not exist.

See https://pypi.org/project/fsspec/2025.3.2/

piskunow:
Fix #957

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
 aiobotocore>=2.5.4,<3.0.0
-fsspec==2025.3.2.*
+fsspec==2025.3.2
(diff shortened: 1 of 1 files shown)


martindurant:
Are you having problems with the current syntax? 2025.3.2.* does indeed match 2025.3.2

martindurant:
Which parsers do you have in mind aside from pip?
Note that the PEP link specifically does have four component versions, and adds that extra "*" is also permitted.

Having said that, I don't recall why the version bound became like this. Perhaps others will see this thread and comment.

Issue #960, opened 2025-04-29
Using S3 list_objects_v2 call to check the existence of a bucket is too costly


yima77:
Here, I refer to the following code lines in s3fs/core.py:

`
1107             try:
1108                 await self._call_s3(
1109                     "list_objects_v2", MaxKeys=1, Bucket=bucket, **self.req_kw
1110                 )
1111                 return True
1112             except Exception:
1113                 pass
`

The purpose of these code lines is to check if the bucket exists. This can be accomplished by calling "head_bucket", instead. Listing objects, even with max-keys=1, can be costly on a bucket whose index is managed in multiple shards, like ceph does.

jacobtomlinson:
This has been resolved in #961 and can be closed

PR #961, opened 2025-04-29
Replace call to list_objects_v2 with head_bucket

yima77:
For better performance and lower stress on the S3.

Address #960

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -1104,9 +1104,7 @@ class S3FileSystem(AsyncFileSystem):
                 # might still be a bucket we can access but don't own
                 pass
             try:
-                await self._call_s3(
-                    "head_bucket", Bucket=bucket, **self.req_kw
-                )
+                await self._call_s3("head_bucket", Bucket=bucket, **self.req_kw)
                 return True
             except Exception:
                 pass


martindurant:
Thanks for your contribution.

I'm afraid this is not a suitable solution. head_bucket is not as likely to be permitted for a user as list_bucket, so you will get erroneous False in such cases. Similarly, list_buckets may or may not be allowed, and may or may not include the bucket in question (depending on whether it is "owned" by the user or not).

So, the situation is complex, and list_objects seems like the most likely to work for most people, even though it indeed has a cost. 
I could imagine there being an option to use head_bucket for cases where the user knows it will work for them - it would join a long list of other options that the user needs to know about.

yima77:
> Thanks for your contribution.
> 
> I'm afraid this is not a suitable solution. head_bucket is not as likely to be permitted for a user as list_bucket, so you will get erroneous False in such cases. Similarly, list_buckets may or may not be allowed, and may or may not include the bucket in question (depending on whether it is "owned" by the user or not).
> 
> So, the situation is complex, and list_objects seems like the most likely to work for most people, even though it indeed has a cost. I could imagine there being an option to use head_bucket for cases where the user knows it will work for them - it would join a long list of other options that the user needs to know about.

Thanks for the comment. It prompted me to double check on the permission related to those ops. It seems that ListObjects, ListObjectV2 and HeadBucket all depend on the same permission -- s3:ListBucket, based on this page (https://docs.aws.amazon.com/AmazonS3/latest/userguide/using-with-s3-policy-actions.html#using-with-s3-policy-actions-related-to-buckets). So, it seems that when list_objects_v2 works, so should head_bucket, unless I am missing something. I also verified this with ceph source code.

martindurant:
I have been trying to trawl through the AWS docs to refute your claim, but I have not been able to. Yet, this was not the case before! There definitely used to be list-specific permissions.

I wonder if it is possible to have list/read permissions for just some part of a given bucket? I also notice, that there is now a new directory-bucket type, and I don't know how this might affect things.

yima77:
> I have been trying to trawl through the AWS docs to refute your claim, but I have not been able to. Yet, this was not the case before! There definitely used to be list-specific permissions.
> 
> I wonder if it is possible to have list/read permissions for just some part of a given bucket? I also notice, that there is now a new directory-bucket type, and I don't know how this might affect things.

It is possible that both were treated differently by AWS S3 before. At this moment, I think AWS should stick to what it publishes, i.e. the same permission is applied to both. I also double checked the history of ceph code. It has been consistently applying the same permission for both since at least 14 years ago when it just used the general read permission. 8 years ago, it started to apply s3ListBucket permission for both and it remains so ever since.

PR #963, opened 2025-05-20
changelog

martindurant:


martindurant:
@d-v-b , see #966 ; only zarr cares about this?

For reference, distributed depends on an exact pin of dask ( https://github.com/dask/distributed/blob/main/pyproject.toml#L31 ).

d-v-b:
> @d-v-b , see #966 ; only zarr cares about this?
> 
> For reference, distributed depends on an exact pin of dask ( https://github.com/dask/distributed/blob/main/pyproject.toml#L31 ).

Anyone who attempts to install 

```
    'fsspec @ git+https://github.com/fsspec/filesystem_spec',
    's3fs @ git+https://github.com/fsspec/s3fs',
```

in the same environment with a solver that cares about a consistent set of dependencies will care about this.

PR #966, opened 2025-05-23
Allow installing from main branches of fsspec and s3fs

maxrjones:
Pinning fsspec at 2025.5.0 means that it is not possible to install from dev branches of both fsspec and s3fs, which is useful for downstream libraries testing upstream compatibility. Would you consider instead pinning to >=2025.5.0 or 2025.5.* instead?

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
 aiobotocore>=2.5.4,<3.0.0
-fsspec==2025.5.0
+fsspec>=2025.5.0
 aiohttp!=4.0.0a0, !=4.0.0a1


martindurant:
We do not want to accidentally pick up a potentially incompatible future release of fsspec, though. A previous suggestion, which does work for this use case (and we used in the past) was "2025.5.0.*", but this confuses some environment managers like poetry. 
If you use git+http in the requirements for both fsspec and s3fs, the dependency here should not matter.

maxrjones:
> We do not want to accidentally pick up a potentially incompatible future release of fsspec, though. A previous suggestion, which does work for this use case (and we used in the past) was "2025.5.0.*", but this confuses some environment managers like poetry.
> If you use git+http in the requirements for both fsspec and s3fs, the dependency here should not matter.

Makes sense. It's too bad that the solution to https://github.com/fsspec/s3fs/issues/957 seems to fix some environment managers like poetry but breaks others like uv and pixi. E.g., I get the following error when using `git+http`:

```
    × failed to solve the pypi requirements of 'upstream' 'osx-64'
    ├─▶ failed to resolve pypi dependencies
    ╰─▶ Because only s3fs==2025.5.0 is available and s3fs==2025.5.0 depends
        on fsspec==2025.5.0, we can conclude that all versions of s3fs depend
        on fsspec==2025.5.0.
        And because there is no version of fsspec==2025.5.0 and you require
        s3fs, we can conclude that your requirements are unsatisfiable.
```
with the specification in https://github.com/zarr-developers/VirtualiZarr/blob/43209b32b8d84f7f57f91942c5b8d3b6cf2db0f4/pyproject.toml#L92-L104.

We can just only test against upstream s3fs moving forward - https://github.com/zarr-developers/VirtualiZarr/pull/602

martindurant:
What tool is reporting that failure?

Another option, of course, it to install in two steps or with ``--no-deps`` (having previously installed all deps or the release versions).

maxrjones:
> Anecdotally, I have heard the sentiment "great for projects, difficult for library dev" of pixi and uv.

fwiw I've been quite happy generally with pixi and uv for library development. I think that this error is accurate to the relationship between fsspec and s3fs right now. The requirements.txt in s3fs states that I should only be able to use the main branch with an exact tag of fsspec. I am trying to do something different in my pyproject.toml. uv is raising an error, which seems appropriate. Either I should request that s3fs changes its expectations (this closed PR) or change my expectations (e.g., https://github.com/zarr-developers/VirtualiZarr/pull/602).

d-v-b:
> We do not want to accidentally pick up a potentially incompatible future release of fsspec, though. A previous suggestion, which does work for this use case (and we used in the past) was "2025.5.0.*", but this confuses some environment managers like poetry. If you use git+http in the requirements for both fsspec and s3fs, the dependency here should not matter.

What about picking up a _compatible_ future release of fsspec? Presumably this is much more likely than an incompatible release, at least in the short time horizon, but we cannot do that as long as fsspec is pinned to an exact version. 

And dealing with an incompatible future release of fsspec is relatively simple, from a library POV: you exclude exactly the incompatible version (up to a range, if you know that all future versions will be incompatible).

In the short term I will fork s3fs and change the deps myself

maxrjones:
Does 2025.5.x (note difference from the previously used 2025.5.0.x) also cause issues with package managers? I thought the final x in CalVer usually indicates bug fixes such that there wouldn't be breaking changes sneaking through with that syntax.

martindurant:
> I thought the final x in CalVer usually indicates bug fixes 

Usually, but it's not a guarantee. I don't tend to release more than once a month, but ...

d-v-b:
i found [this post](https://iscinumpy.dev/post/bound-version-constraints/) helpful for explaining why an upper bound for a version can be problematic.

martindurant:
The post is very long it will take some time for me to have a look. I am not doubting that ">=" has some advantages, but I am also pointing out that it definitely also has disadvantages, and this repo in particular has previously been hit by them.

d-v-b:
in any case, our tests are running normally again (my fork was unnecessary in the end) :)

martindurant:
> Relaxing the upper bound on the fsspec dependency would solve this, at the risk that some future version of fsspec might be incompatible with s3fs, but in this case, people affected by that problem can do something: exclude the incompatible version of fsspec.

Unfortunately, the size of the community using released s3fs is far far bigger than those who need dev s3fs and dev fsspec, so this is not a risk that we can take.

d-v-b:
isn't it possible to test for incompatibility between s3fs and fsspec, and then update s3fs' dependencies accordingly?

martindurant:
No, you cannot change the upper bound of the released version once it is on pypi. Did I misunderstand you? Perhaps you have a concrete suggestion of what to do.

martindurant:
Yes, it would be totally fine to use .dev1 versions in s3fs; and indeed to ditch versioneer in favour of something else. But I think it's the version of fsspec that's at issue, not s3fs. Actually, you want the source install **not** to add a suffix to the version string.

d-v-b:
i think it's important and correct to track the difference between version `X` and `X+commit hash`, and to treat the latter as a newer version than the former. We can solve this on the zarr side by respecting the fact that the dev versions of s3fs and fsspec are not generally compatible, and testing them separately as needed.

Issue #871, opened 2024-04-26
Working example of using Async/Await


Sarafudinov:
Hello, I’m working with s3fs and haven’t found any examples of how to work with asynchrony; after looking through your checklog I saw that you’ve already added functionality. Could you please tell me how to use asynchronous methods for reading/writing and listing files
the same question was asked in 2020
then this functionality was not ready and only a link to a small piece was provided https://s3fs.readthedocs.io/en/latest/#async

Thanks in advance for your attention!)

martindurant:
What kind of workflow would you like? Perhaps I can sketch out what you need to do, and you help update the docs?

Sarafudinov:
Thank you, I would like to learn more, see examples of how to use the reading (s3.open), writing and listing (s3.ls) methods in asynchronous code
1) I need the ability to asynchronously download files from a bucket (as far as I understand, this is the open_async method)
2) I would also like to see an example of asynchronously writing files to a bucket
3) and regarding the listing of files, I don’t quite understand whether there is an analogue when initializing s3 = S3FileSystem(asynchronous=True) since to work with the listing (s3.ls(path)) you have to use synchronous initialization s3 = S3FileSystem()

Sarafudinov:
all I need is an example of how to read files from the bucket, write files to the bucket and look at the listing of files on the bucket

at the same time, so that it is all asynchronous
As far as I understand from your changelog, you have already implemented this, but I haven’t found any examples

2023.4.0
Add streaming async read file (#722)
2023.1.0
async iterator for listings (#670)

or for now link only to the botocore library https://aiobotocore.readthedocs.io/en/latest/examples/s3/basic_usage.html

martindurant:
Sorry to be slow, thanks for the ping.
Here is an example:
```
import asyncio

async def f():
    import fsspec
    mybucket = "mybucket"   # <- change this
    fs = fsspec.filesystem("s3", asynchronous=True, anon=False)
    print(await fs._ls(mybucket))
    await fs._pipe_file(f"{mybucket}/afile", b"hello world")
    print(await fs._cat_file(f"{mybucket}/afile"))

asyncio.run(f())
```

You should note that the file-like interface (via `fs.open(...)`) is *not* async, because the upstream python file-like object (io.IOBase) is not async. This is why open_async exists, but this "streaming" API is incomplete.

Sarafudinov:
do not quite understand😅
As far as I understand, asynchronous reading is designed through s3fs.open_async()
[...]

PR #967, opened 2025-05-24
Add paritial read feature for open_async

Bing1996:
## Feature
This feature supports to the open_async function in S3FileSystem Class, enabling efficient async reading of S3 objects without blocking the event loop. This feature is particularly valuable for applications that need to process large S3 files asynchronously or implement streaming data processing pipelines. And add a relative test for async partial read.

## Benefits
- Flexibility: Support for both full and partial file access

## Example
A paritial read example for open_async based on #871 
```python
fn = test_bucket_name + "/target"
data = b"hello world" * 1000
out = []

async def read_stream():
    fs = S3FileSystem(
        anon=False,
        skip_instance_cache=True,
    )
    await fs._mkdir(test_bucket_name)
    await fs._pipe(fn, data)
    f = await fs.open_async(fn, mode="rb", loc=0, size=len(data) // 2)

    while True:
        got = await f.read(1000)
        assert f.size == len(data) // 2
        assert f.tell()
        if not got:
            break
        out.append(got)

asyncio.run(read_stream())
assert len(b"".join(out)) == len(data) // 2
assert b"".join(out) == data[: len(data) // 2]
```

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -48,7 +48,6 @@ logger = logging.getLogger("s3fs")
 
 
 def setup_logging(level=None):
-
     setup_logger(logger=logger, level=(level or os.environ["S3FS_LOGGING_LEVEL"]))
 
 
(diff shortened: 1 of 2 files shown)


martindurant:
Sorry for being slow to respond.
I am trying to wrap my head around this - it feels like half way between the random-access normal file and the streaming file. So you do get a streaming response (with whatever internal chunking or none), but over only part of the data.

Bing1996:
Yes, it does indeed achieve this kind of purpose. For example, with a large object that has a bytes location index file, we only need to asynchronously fetch the parts we need from this large file based on this index file, and then collect them together at the end. This will greatly accelerate the reading step, and is especially suitable for targeted extraction processes (like partial variables in some earth science datasets) from big data files that don't require full data retrieval.

PR #974, opened 2025-07-01
core.py: fix exclusive write for small files

JayThomason:
Prior to this change the exclusive write mode "x" did not behave properly for small files, instead acting the same as "w".

This change solves that by passing the IfNotMatch header to put_object even when writing small files that don't require a multipart upload.

Fixes https://github.com/fsspec/s3fs/issues/973

Tested by running the test script linked in the issue as well as updating the unit tests.

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -2458,6 +2458,7 @@ class S3File(AbstractBufferedFile):
 
     def commit(self):
         logger.debug("Commit %s" % self)
+        match = {"IfNoneMatch": "*"} if "x" in self.mode else {}
         if self.tell() == 0:
             if self.buffer is not None:
                 logger.debug("Empty file committed %s" % self)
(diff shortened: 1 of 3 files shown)


martindurant:
Can we please have a test function added? Perhaps also a zero-length file would be nice.

JayThomason:
> Can we please have a test function added? Perhaps also a zero-length file would be nice.

Sure let me work on that.

JayThomason:
@martindurant I updated the unit tests and also discovered a related bug in moto: https://github.com/getmoto/moto/issues/9028

Once this bug is fixed in moto the xfail decorator can be removed from the remaining tests.

martindurant:
(check the version of moto and botocore you have versus the one in CI)

JayThomason:
It looks like the tests will fail if the botocore version is < 1.33.0 because that is when IfNoneMatch support for conditional writes was added to the s3 client. I'll see if we can conditionally mark the tests as xfail depending on the botocore version to ensure they continue passing on newer botocore versions moving forward.

JayThomason:
@martindurant is there anything else you want to see in this PR?

martindurant:
Still catching aiobotocore versions >=2.8.0,<2.9.0

Also some formatting

JayThomason:
Thanks @martindurant  I just took another stab at it. If you can trigger Ci that would be great.

After manually searching through the botocore docs of different versions I found IfNoneMatch for put_object and complete_multipart_upload wasn't added to botocore until 1.36.0, not 1.33.0.

PR #983, opened 2025-08-23
Skip the current directory entity when running 'ls' or 'listdir'

rilshok:
When using `S3FileSystem.ls` or `S3FileSystem.listdir`, zero-sized objects that serve as directory markers in S3 were incorrectly listed as files.

Amazon S3 employs a flat object storage model, but its console interface allows users to create "folders" by adding zero-size objects with keys ending in a delimiter (e.g., `myfolder/`). These objects act as placeholders to visually represent directories but are not actual files.

This PR implements a check to skip zero-size objects that match the current directory prefix during listing operations.

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -813,7 +813,7 @@ class S3FileSystem(AsyncFileSystem):
                 self._fill_info(c, bucket, versions=False)
                 yield c
             for c in i.get(contents_key, []):
-                if c["Size"] == 0 and c["Key"] == i["Prefix"]:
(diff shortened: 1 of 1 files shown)


martindurant:
I am surprised to see only one failure! Of course, the tests do NOT tend to make placeholder files, since we know they are problematic and unnecessary.

I would expect, that bulk rm may fail in the case of placeholders, since the files deleted will be those returned by find(), and not include them - so only normal files will be deleted and not the "directories", leaving the tree structure erroniously in place. This should be explicitly tested.

rilshok:
Yes, you are right that this scenario needs to be considered.

PR #987, opened 2025-10-01
If no key get bucket info

jacobtomlinson:
In #961 @martindurant mentions that it would be nice to handle the case where no key is specified and use `head_bucket` to gather more information about the bucket. (https://github.com/fsspec/s3fs/pull/961#issuecomment-2847836001)

This PR implements that and returns the `VersionId` for the bucket. Is there anything else that would be useful to include in here?


martindurant:
With the caveat (from the linked thread), that I remember there being a permission issue for head_bucket that I am not assured is no longer a a problem.

I would expect it to work on moto, though, so we have a simple test case to look for that extra information.

PR #989, opened 2025-10-07
Prevent duplicated entries in find() in presence of directory placeholders

### LINKED ISSUE #988: glob with `**` leads to incorrect listings_cache state (missing and duplicated directories listings)
It seems that usage of `**` in `glob` can lead to incorrect listings cache state so that subsequent commands like `glob` or `ls` produce incorrect results.

I have two examples.

### Missing directories

```
import s3fs

print(s3fs.__version__)
s3 = s3fs.S3FileSystem()
s3.glob("s3://<redacted bucket>/<redacted directory>/*/<redacted>/**/*.txt")
print(s3.ls("s3://<redacted bucket>/<redacted directory>"))
```

Here `s3.ls` misses several directories that are contained in `<redacted directory>`. If I comment `s3.glob` out, or replace `**` with `*` (the output of `glob` does not change, as there is only one subdirectory level in `<redacted>`), no missed folders in `ls` output.

### Duplicated directories

I tried to reproduce the previous example with a simpler directory structure, and encountered different problem.

I have a very simple structure inside my bucket:

```
a/
b/somefile.pdf
c/
```

Then I do

```
import s3fs

print(s3fs.__version__)
s3 = s3fs.S3FileSystem()

s3.glob("s3://<test bucket>/**/*.pdf")
print(s3.ls("s3://<test bucket>/"))
```
The output:

```
2025.9.0
['<test bucket>/a/', '<test bucket>/b', '<test bucket>/b', '<test bucket>/b/', '<test bucket>/c/']
```

So directory `b` is present three times (one with a leading `/`, the other two without).

If I replace `**` with `*`, I get different output:

```
2025.9.0
['<test bucket>/a', '<test bucket>/b', '<test bucket>/c']
```

In all cases, if I initialize the filesystem with `s3 = s3fs.S3FileSystem(use_listings_cache=False)`, the problems disappear. So I believe the problem is the interaction between `**` in glob and the cache.

I am using 2025.9.0 version of s3fs, Python 3.12.9 on Mac OS.

martindurant:
Fixes #988 

@ischurov , can you try this?

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -891,32 +891,37 @@ class S3FileSystem(AsyncFileSystem):
             # not self._parent, because that strips "/" from placeholders
             par = o["name"].rsplit("/", maxsplit=1)[0]
             o["Key"] = o["name"]
-            if par not in sdirs:
-                sdirs.add(par)
-                d = False
-                if len(path) <= len(par):
-                    d = {
-                        "Key": par,
-                        "Size": 0,
-                        "name": par,
-                        "StorageClass": "DIRECTORY",
-                        "type": "directory",
-                        "size": 0,
-                    }
-                    dirs.append(d)
-                thisdircache[par] = []
-                ppar = self._parent(par)
-                if ppar in thisdircache:
-                    if d and d not in thisdircache[ppar]:
-                        thisdircache[ppar].append(d)
-            if par in sdirs and not o["name"].endswith("/"):
-                # exclude placeholdees, they do not belong in the directory listing
-                thisdircache[par].append(o)
+            name = o["name"]
+            while "/" in par:
+                if par not in sdirs:
+                    sdirs.add(par)
+                    d = False
+                    if len(path) <= len(par):
+                        d = {
+                            "Key": par,
+                            "Size": 0,
+                            "name": par,
+                            "StorageClass": "DIRECTORY",
+                            "type": "directory",
+                            "size": 0,
+                        }
+                        dirs.append(d)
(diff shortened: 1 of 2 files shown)


ischurov:
Now I have this output (for the same example as [here](https://github.com/fsspec/s3fs/issues/988#issuecomment-3372221779)):

```
s3fs.__version__='2024.10.0+31.ge8f64a8'
['<test bucket>/a/', '<test bucket>/b', '<test bucket>/b/', '<test bucket>/c/']
```

Note that the output still differs from what I see if I do not invoke `s3.find` before `s3.ls`:

```
s3fs.__version__='2024.10.0+31.ge8f64a8'
['<test bucket>/a', '<test bucket>/b', '<test bucket>/c']
```

I don't know whether it is expected behavior or not.

Also, my initial problem with missing directories is still present.

martindurant:
Hmm, I'll try again when I have the time. I agree the two outputs should be the same... Whether ls() should show the placeholders, I'm not sure, since there're not "within" the listed directory (i.e., "<listed_path>/..."). I should definitely have tests around this!

martindurant:
@ischurov : I think I have it. Can you please test for your usecase?

martindurant:
OK, I've updated the test, so that if fails in the way you specify - I'll have to look again.

ischurov:
Added [the testcase](https://github.com/fsspec/s3fs/issues/988#issuecomment-3436727753), reproduces at 9a911bb.

martindurant:
@ischurov test again?

PR #1007, opened 2026-02-09
Add concurrent downloads to get_file and cat_file

### LINKED ISSUE #841: Add transfer configuration to support concurrent downloading
I am trying to speed up downloading of large files from S3.

I can enable parallel downloading of file parts in boto3 using https://boto3.amazonaws.com/v1/documentation/api/latest/guide/s3.html:
```
    config = TransferConfig(
        multipart_threshold=1000000000,
        multipart_chunksize=50000000,
        max_concurrency=128,
    )
    s3_client.download_file(bucket, key, dst, Config=config)
```

Can I do the same in s3fs? if seems that passing `s3_additional_kwargs` doesn't help:
```
s3 = s3fs.S3FileSystem(anon=False, s3_additional_kwargs={'Config': config}) 
```

It would be great to have this feature in s3fs.

dwo:
Closes #841

Adds parallel byte-range GET support to `_get_file` and `_cat_file`, mirroring the existing `_upload_file_part_concurrent` pattern used by `put_file`.

When a file is larger than `chunksize` and `max_concurrency > 1`, downloads are split into byte-range requests executed in batches via `asyncio.gather`. Smaller files and explicit range requests (`start`/`end`  on `cat_file`) fall through to the existing sequential path unchanged.

As for my motivation: boto3's `download_file` was a lot faster than s3fs `get_file` for a 20GB file in Cloudflare R2 from our cloud instances.

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -329,7 +329,8 @@ class S3FileSystem(AsyncFileSystem):
         simultaneous connections is ``max_concurrency * batch_size``.
         Increasing this value will result in higher
         memory usage during multipart transfer operations (by
-        ``max_concurrency * chunksize`` bytes per file).
+       ``max_concurrency * chunksize`` bytes per file), and may result in timeouts on slow
+       networks. Set to 1 if you are having connection issues.
     fixed_upload_size : bool (False)
         Use same chunk size for all parts in multipart upload (last part can be smaller).
         Cloudflare R2 storage requires fixed_upload_size=True for multipart uploads.


martindurant (commented):
I am generally positive about the changes here.

The problem with concurrency on slow networks, is that connections will time out if they are launches at the same time, but then one swamps the bandwidth. Therefor, it seems best to default to concurrency 1 unless we can reliably determine the bandwidth to the data or whether the call location is coresident.

s3fs/core.py
```diff
@@ -323,11 +323,12 @@ class S3FileSystem(AsyncFileSystem):
...
-        conjunction with ``S3FileSystem.put(batch_size=...)`` the maximum number of
-        simultaneous connections is ``max_concurrency * batch_size``. We may extend
-        this parameter to affect ``pipe()``, ``cat()`` and ``get()``. Increasing this
-        value will result in higher memory usage during multipart upload operations (by
+        upload (``put()``, ``pipe()``), download (``get()``) and read
+        (``cat()``) operations. Defaults to 10. When used in conjunction
+        with ``S3FileSystem.put(batch_size=...)`` the maximum number of
+        simultaneous connections is ``max_concurrency * batch_size``.
+        Increasing this value will result in higher
+        memory usage during multipart transfer operations (by
```
martindurant: Please add something that says: if you are seeing timeout errors, typical of slower networks, you should reduce the concurrency, perhaps all the way down to 1.

PR #1008, opened 2026-02-10
Add option to check file expiry locally

freundTech:
Some S3 compatible services don't support the `If-Match` header, which s3fs uses to check if files have been changed since they have been opened.  Namely VAST S3 refuses all request with the `If-Match` header set, to not accidentally return unexpected data.

This change adds a `local_expiry_check` option to `S3FileSystem`, which disables sending the `If-Match` header and instead compares `ETag`s locally to make sure the file didn't change.

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -332,6 +332,10 @@ class S3FileSystem(AsyncFileSystem):
     fixed_upload_size : bool (False)
         Use same chunk size for all parts in multipart upload (last part can be smaller).
         Cloudflare R2 storage requires fixed_upload_size=True for multipart uploads.
+    local_expiry_check : bool (False)
+        Perform expiry checks when using range reads locally instead of having the
+        server perform them. VAST S3 requires local_expiry_check=True because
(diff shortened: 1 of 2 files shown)


s3fs/core.py
```diff
@@ -2656,10 +2663,10 @@ def _fetch_range(fs, bucket, key, version_id, start, end, req_kw=None):
         )
         return b""
     logger.debug("Fetch: %s/%s, %s-%s", bucket, key, start, end)
-    return sync(fs.loop, _inner_fetch, fs, bucket, key, version_id, start, end, req_kw)
+    return sync(fs.loop, _inner_fetch, fs, bucket, key, version_id, details, start, end, req_kw)
 
 
-async def _inner_fetch(fs, bucket, key, version_id, start, end, req_kw=None):
+async def _inner_fetch(fs, bucket, key, version_id, details, start, end, req_kw=None):
```
martindurant: I don't think this is used from anywhere except your calls, but it would be good to add extra args only as kwargs at the end to make sure the signature doesn't drift.

PR #1016, opened 2026-03-23
Allow anonymous access to be configured by an environment variable

### LINKED ISSUE #675: ENH: Eliminate the need for anon=True when you might be opening a public bucket & object.
My routine is given a s3 url to a file. I don't know apriori if the bucket is public read-only bucket or has been secured via IAM role. 

The current s3fs interface, requires my code needs to pass `anon=True` if the bucket is public if I want to avoid the python exception handling that generates the "Unable to locate credentials" error.

This enhancement request is to have s3fs automatically handle opening public s3 buckets & paths whether they're public (permitting anonymous access) or private and requiring a supplied credential. Eliminate the need for `anon=True`

braingram:
Closes https://github.com/fsspec/s3fs/issues/675

Add support for reading an environment variable `S3FS_ANONYMOUS` when a value for `anon` is not provided to `S3FileSystem.__init__`.

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -243,7 +243,8 @@ also has the appropriate environment variables, config files or IAM roles
 available.
 
 If none of the credential methods are available, only anonymous access will
-work, and ``anon=True`` must be passed to the constructor.
+work, and ``anon=True`` must be passed to the constructor (or the environment
+variable ``S3FS_ANONYMOUS`` set to ``True`` or ``1``).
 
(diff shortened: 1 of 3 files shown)


martindurant:
As noted in the sibling issue, you can already do this. You can provide JSON for the whole of the s3 config, or set just the one value using FSSPEC_S3_ANON= (actually this is bugged, and any value will evaluate as True, if set).

braingram:
Thanks. I wasn't aware of that option.

PR #1026, opened 2026-05-13
Allow installation vs. fsspec git tip

crusaderky:
Make it possible to install both fsspec and s3fs from git tip.

This PR fixes this failure:
```
$ pip install git+https://github.com/fsspec/filesystem_spec git+https://github.com/fsspec/s3fs
[...]
ERROR: Cannot install fsspec 2026.4.0.post5+g85facaaf2 (from git+https://github.com/fsspec/filesystem_spec) and s3fs==2026.4.0+1.g2db8d26 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested fsspec 2026.4.0.post5+g85facaaf2 (from git+https://github.com/fsspec/filesystem_spec)
    s3fs 2026.4.0+1.g2db8d26 depends on fsspec==2026.4.0

Additionally, some packages in these conflicts have no matching distributions available for your environment:
    fsspec

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip to attempt to solve the dependency conflict

ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts
```

## Context
The upstream CI workflow in dask installs both s3fs and fsspec from git tip.
It currently works by forcing its way through with manual hacks. This will no longer be possible after migrating to pixi.

CC @martindurant


martindurant:
Supersedes #1023 - I think this is the right way to do it.

PR #1028, opened 2026-05-24
fix: close stale S3BucketRegionCache on session refresh; guard _sessions=None; serialize set_session() under asyncio.Lock

josh-ag2:
## Problem

Three related issues in `S3FileSystem.set_session()` that together cause `aiohttp` `Unclosed client session` / `Unclosed connector` warnings when the server closes connections mid-session:

1. **Stale `_s3creator` is silently abandoned on refresh.** When `set_session(refresh=True)` replaces `self._s3creator` with a new `S3BucketRegionCache`, the old cache is never closed. It holds aiobotocore-created `aiohttp.ClientSession` objects (1 general + up to N regional). GC'd without an explicit `close()` → aiohttp emits warnings.

2. **Vacuous-truth refresh on empty `_sessions`.** `all(_.closed for _ in [])` is always `True`, so a freshly-entered `AIOHTTPSession` with `_sessions = {}` (no requests sent yet) incorrectly triggers `refresh=True`. This causes the next issue to fire spuriously.

3. **Concurrent callers (e.g. `asyncio.gather`) each create their own creator.** When `self._s3 is None`, multiple concurrent coroutines all fall through to the setup path. Without serialization, Task B's Fix 1 (close stale creator) tears down Task A's creator while Task A is mid-request — setting `_sessions = None` on an active `AIOHTTPSession`, producing:
   ```
   AttributeError: 'NoneType' object has no attribute 'get'
   ```
   at `aiobotocore/httpsession.py`.

The stale-check also needs to guard against newer aiobotocore setting `_sessions = None` (rather than `{}`) after `__aexit__`.

## Changes

**`s3fs/core.py`**

- **Fix 1 — Close stale `_s3creator` on refresh**: Before `self._s3creator = s3creator`, call `await old_creator.__aexit__(None, None, None)` so all regional aiobotocore clients and their aiohttp sessions are closed cleanly.

- **Fix 2 — Guard stale-session check**: Change the stale-session refresh condition to:
  ```python
  if sessions is not None and sessions and all(_.closed for _ in sessions.values()):
      refresh = True
  ```
  - `sessions is not None` — newer aiobotocore sets `_sessions = None` after `__aexit__`; treat `None` as "not stale, just closed"
  - `sessions` (non-empty) — an empty dict means no requests have been sent yet; not stale

- **Fix 3 — Serialize creator setup under `asyncio.Lock`**: Wrap the entire creator-setup path in `async with self._setup_lock`. Concurrent tasks wait at the lock, then re-check `self._s3 is not None` and reuse the already-created session rather than creating their own. This makes Fix 1 safe: the old creator is only closed once, by the task that wins the lock, before any new sessions are open.

**`s3fs/tests/test_s3fs.py`**

- `test_stale_creator_closed_on_refresh` — patches `__aexit__` on the old creator and asserts it was called during `set_session(refresh=True)`

**`docs/source/changelog.rst`** — Unreleased section added.

## Motivation

Discovered while integrating zarr v3's `FsspecStore.close()` (zarr-developers/zarr-python#4003). zarr's `asyncio.gather` fires 4 concurrent `FsspecStore.get()` calls on the same `S3FileSystem` instance, triggering issue 3. Issues 1 and 2 produce residual unclosed-session warnings once the render completes.

Related: aiobotocore/aiobotocore#866

## Test

All existing tests pass. New test confirms `__aexit__` is called on the old creator during refresh.

### DIFF
--- a/s3fs/core.py
+++ b/s3fs/core.py
@@ -600,96 +600,115 @@ class S3FileSystem(AsyncFileSystem):
             hsess = getattr(getattr(self._s3, "_endpoint", None), "http_session", None)
             if hsess is not None:
                 sessions = hsess._sessions  # None after __aexit__ in newer aiobotocore
-                if sessions is not None and all(_.closed for _ in sessions.values()):
+                # Only refresh when sessions exist AND all are server-closed.
+                # An empty dict means no requests have been sent yet — not stale.
+                if sessions is not None and sessions and all(_.closed for _ in sessions.values()):
                     refresh = True
             if not refresh:
                 return self._s3
-        logger.debug("Setting up s3fs instance")
 
-        client_kwargs = self.client_kwargs.copy()
-        init_kwargs = dict(
-            aws_access_key_id=self.key,
-            aws_secret_access_key=self.secret,
-            aws_session_token=self.token,
-            endpoint_url=self.endpoint_url,
-        )
-        init_kwargs = {
-            key: value
-            for key, value in init_kwargs.items()
-            if value is not None and value != client_kwargs.get(key)
-        }
-        if "use_ssl" not in client_kwargs.keys():
-            init_kwargs["use_ssl"] = self.use_ssl
-        config_kwargs = self._prepare_config_kwargs()
-        if self.anon:
-            from botocore import UNSIGNED
+        # Serialize creator setup.  Callers such as zarr may launch concurrent
+        # coroutines via asyncio.gather that all see self._s3 is None and each
+        # try to build their own creator.  Without this lock every concurrent
+        # task would create a separate creator and then Fix 1 (close stale
+        # creator) would tear down a creator that a sibling task is still using,
+        # causing ``_sessions = None`` mid-request.  asyncio.Lock is
+        # event-loop–aware; creating it between two non-await lines is atomic
+        # so there is no race on the hasattr check.
+        if not hasattr(self, "_setup_lock"):
+            self._setup_lock = asyncio.Lock()
+        async with self._setup_lock:
+            # Re-check under the lock: a concurrent task may have set up the
+            # session while we were waiting.
+            if self._s3 is not None and not refresh:
+                return self._s3
 
-            drop_keys = {
-                "aws_access_key_id",
-                "aws_secret_access_key",
-                "aws_session_token",
-            }
+            logger.debug("Setting up s3fs instance")
+
+            client_kwargs = self.client_kwargs.copy()
+            init_kwargs = dict(
+                aws_access_key_id=self.key,
+                aws_secret_access_key=self.secret,
+                aws_session_token=self.token,
+                endpoint_url=self.endpoint_url,
+            )
             init_kwargs = {
-                key: value for key, value in init_kwargs.items() if key not in drop_keys
-            }
-            client_kwargs = {
                 key: value
-                for key, value in client_kwargs.items()
-                if key not in drop_keys
+                for key, value in init_kwargs.items()
+                if value is not None and value != client_kwargs.get(key)
             }
-            config_kwargs["signature_version"] = UNSIGNED
+            if "use_ssl" not in client_kwargs.keys():
+                init_kwargs["use_ssl"] = self.use_ssl
+            config_kwargs = self._prepare_config_kwargs()
+            if self.anon:
+                from botocore import UNSIGNED
 
-        conf = AioConfig(**config_kwargs)
-        if self.session is None or refresh:
-            self.session = aiobotocore.session.AioSession(**self.kwargs)
+                drop_keys = {
+                    "aws_access_key_id",
+                    "aws_secret_access_key",
+                    "aws_session_token",
+                }
+                init_kwargs = {
+                    key: value for key, value in init_kwargs.items() if key not in drop_keys
+                }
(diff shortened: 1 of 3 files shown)


s3fs/tests/test_s3fs.py
```diff
@@ -3189,6 +3189,38 @@ async def run_program(run):
...
 
+def test_stale_creator_closed_on_refresh(s3):
+    """Old _s3creator must have __aexit__ called when set_session refreshes.
+
+    Without the fix, the stale aiobotocore ClientSession held by the old
+    _s3creator is garbage-collected without explicit close(), producing
+    aiohttp 'Unclosed client session' warnings at process exit.
+    """
+
+    async def run():
```
martindurant, line 3200: Could use @pytest.mark.asyncio instead of embedding the coroutine.

s3fs/core.py
```diff
@@ -599,81 +599,116 @@ async def set_session(self, refresh=False, kwargs={}):
...
+        # try to build their own creator.  Without this lock every concurrent
+        # task would create a separate creator and then Fix 1 (close stale
+        # creator) would tear down a creator that a sibling task is still using,
+        # causing ``_sessions = None`` mid-request.  asyncio.Lock is
+        # event-loop–aware; creating it between two non-await lines is atomic
+        # so there is no race on the hasattr check.
+        if not hasattr(self, "_setup_lock"):
+            self._setup_lock = asyncio.Lock()
+        async with self._setup_lock:
+            # Re-check under the lock: a concurrent task may have set up the
```
martindurant, line 657: ~Since there was no `await` between creation of the lock and acquiring it, no other coroutine could have run. The only way the lock might have been acquired is in another thread - but asyncio threads are not thread-safe anyway (and all async code should be running in the one event loop on the one thread)~

OK, so the model is, that one coroutine might have run to the `awaits` below (`get_client`, `__aenter__`) while this coroutine is waiting for the lock?

s3fs/core.py
```diff
@@ -599,81 +599,116 @@ async def set_session(self, refresh=False, kwargs={}):
...
+        # Serialize creator setup.  Callers such as zarr may launch concurrent
+        # coroutines via asyncio.gather that all see self._s3 is None and each
+        # try to build their own creator.  Without this lock every concurrent
+        # task would create a separate creator and then Fix 1 (close stale
+        # creator) would tear down a creator that a sibling task is still using,
+        # causing ``_sessions = None`` mid-request.  asyncio.Lock is
+        # event-loop–aware; creating it between two non-await lines is atomic
+        # so there is no race on the hasattr check.
+        if not hasattr(self, "_setup_lock"):
+            self._setup_lock = asyncio.Lock()
```
martindurant: Prefer to use `self._setup_lock = None` in `__init__`.
Note that we don't remove the lock - should it be made in `__init__` instead?

martindurant (commented):
OK, I think I have wrapped my head around it now. It's a pity that indenting such a big block messes us the look of the diff so badly; it's only the bit with `await`s that needs to go inside the lock block correct? Maybe the kwargs can be worked out beforehand with the same indentation as before.

s3fs/tests/test_s3fs.py
josh-ag2: Happy to switch to `@pytest.mark.asyncio` — worth noting that `pytest-asyncio` isn't currently a test dependency and the existing async tests in this file use `asyncio.run()` for consistency. I'll align with whichever pattern you prefer: add `pytest-asyncio` and use the decorator, or keep `asyncio.run()` to match the existing style.

s3fs/core.py
josh-ag2: Yes, that's exactly right. A coroutine can yield control at the `await get_client(...)` and `await __aenter__()` calls below the lock, allowing a sibling coroutine (e.g. launched via `asyncio.gather`) to also enter `set_session()` while the first is suspended. Without the lock they'd each create a separate `_s3creator`, and when the first resumes and replaces `self._s3creator`, it calls `__aexit__` on the stale one — tearing down a session the sibling is still actively using.

s3fs/core.py
josh-ag2: Good point on declaring it in `__init__` — happy to add `self._setup_lock = None` there for clarity. The reason the actual `asyncio.Lock()` creation stays lazy rather than moving entirely to `__init__` is that `AbstractFileSystem` uses instance caching via `__new__`, so an `S3FileSystem` instance may be constructed outside any async context and potentially reused across different event loops. Creating the `Lock` lazily inside `set_session()` (which is always called from within an async context) ensures it's bound to the right loop. On Python 3.10+ this is less of a concern since locks no longer bind at construction time, but keeping it lazy is the safer cross-version choice.

s3fs/tests/test_s3fs.py
martindurant: > pytest-asyncio isn't currently a test dependency

Oh right - never mind then.

s3fs/core.py
martindurant: > On Python 3.10+ this is less of a concern

We don't support 3.9 any more

martindurant:
Can you check with https://github.com/fsspec/s3fs/pull/1020 - does that also offer the right guard now?

PR #1038, opened 2026-07-31
Include ServerSideEncryption in_ info() output.

### LINKED ISSUE #1037: Include ServerSideEncryption in _info() output
Hello,

currently `_info()` function returns a subset of S3 object properties:
https://github.com/fsspec/s3fs/blob/609950a67e1d2f26bd98b5053b016200272a35dd/s3fs/core.py#L1749-L1756
In particular, `ServerSideEncryption` is missing, which makes it hard to check whether ETag is supposed to contain MD5 checksums. Would it be possible to include it? Thanks!

belegdol:
This makes it possible to check whether ETag contains MD5 hash or not. Fixes #1037.


belegdol:
I still need to test it properly hence the draft status.

belegdol:
I was able to test this now. Unfortunately, just as with `ContentType`, there is some inconsistency depending on whether `refresh=True` is passed:

- with `refresh=True`, `ContentType` and `ServerSideEncryption` are reported
- with the default `refresh=False`, they are not. 

This is due to what `ListObjectsV2` produces.

PR #1042, opened 2026-09-16
Retry transient errors on each listing page

### LINKED ISSUE #982: Inconsistent results from find()/glob() under concurrency due to paginator bypassing retry logic
When many workers concurrently call find()/glob() on a prefix with a large number of objects, some workers intermittently get incomplete listings (e.g., 89k+ instead of 90k). Results vary run-to-run. This is extremely hard to reproduce and not deterministic but happened every so often on a workload with a few thousand simultaneous workers that were all globbing `s3://commoncrawl/crawl-data/CC-MAIN-2024-30/segments/` with `"*/warc/*"` (should yield 90k files, see https://data.commoncrawl.org/crawl-data/CC-MAIN-2024-30/index.html).

I suspect that this is due to transient issues in the pagination, where some requests might be failing silently.

`_iterdir` uses the aiobotocore paginator directly, which bypasses s3fs’ `_call_s3/_error_wrapper` retry handling used elsewhere. Timeouts, IncompleteRead, throttling/invalid-XML parse errors, etc can terminate pagination early or drop a page without consistent exceptions being surfaced or retried.

Ideally the pagination requests should be wrapped in the error handling/retrying (or at the very least an exception should be raised when some of the paginated requests fail imo)

cocofabio:
Closes #982.

`_iterdir` drives the aiobotocore paginator directly, so a page request that hits a timeout / throttling / bad-XML error never goes through the `_error_wrapper` retries every other call gets via `_call_s3`. `find()` then raises, and `glob()` — whose directory walk treats a failed `ls` as empty — silently drops the affected subtree, which matches the intermittent short listings in the issue.

Each page is an independent list request, so this routes the paginator's per-page call through `_error_wrapper` (small helper, also applied to the `list_object_versions` paginator in `_rm_versioned_bucket_contents`). The iterator only advances its token after a successful page, so a retry re-issues the same request.

Reproduced with moto by injecting one `socket.timeout` into `list_objects_v2`: on main `glob("bkt/segments/*/warc/*")` returned 4 of 6 files with no error; with this change it returns all 6. The added test does the same through `find()` (fails on main with `FSTimeoutError`).

Tests: `pytest s3fs` — 377 passed, 1 skipped, 4 xfailed; `black --check` / `flake8` clean.


s3fs/core.py
```diff
@@ -942,6 +943,17 @@ async def _iterdir(
...
 
+    def _paginate_with_retries(self, paginator, **kwargs):
+        # The paginator calls the client method directly, bypassing the retry
+        # logic every other request gets via _call_s3, so a transient error on
+        # any page fails (or silently truncates) the whole listing (#982).
+        it = paginator.paginate(**kwargs)
+        make_request = it._make_request
+        it._make_request = lambda kw: _error_wrapper(
+            make_request, args=(kw,), retries=self.retries
+        )
```
martindurant: We should ping whoever owns this code to say we are relying on it - they are probably not expecting anyone to ever call this internal method. Is these no other way?

cocofabio:
Fair enough. Switched to a plain `IsTruncated` loop through `_call_s3` (same shape as `_object_version_info`), so nothing touches botocore internals any more. `max_items` keeps its MaxKeys/truncation behaviour and the versioned-bucket removal uses the same loop.

s3fs/core.py
```diff
@@ -935,12 +926,33 @@ async def _iterdir(
...
+            if remaining is not None:
+                contents = contents[:remaining]
+                remaining -= len(contents)
+            for c in contents:
                 if not self.version_aware or c.get("IsLatest") or versions:
                     c["type"] = "file"
                     c["size"] = c["Size"]
                     self._fill_info(c, bucket, versions=versions)
                     yield c
+            if remaining == 0:
```
martindurant:
```suggestion
            if remaining <= 0:
```
(just in case!)

PR #1043, opened 2026-09-16
Skip the empty-named entries a "/" key produces in listings

### LINKED ISSUE #953: Infinite recursion in walk() when S3 bucket contains object with key /
We've run into a reproducible issue where calling walk() on an S3 bucket causes infinite recursion if the bucket contains an object with the key /

This happens because:

- S3 allows an object to be created with the key /
- ls() returns it as a "directory" (i.e. common prefix /)
- walk() sees that as a subdirectory and recurses into it
- But that "subdirectory" is just the bucket root again → leading to infinite recursion
- This eventually leads to a RecursionError: maximum recursion depth exceeded.

Let us know if this is something you'd consider filtering out at the s3fs level. Happy to provide a test case or additional details!

Thanks for the great library 🙏

itzzdev09:
Fixes #953.

An object whose key is just `/` shows up as an entry with an empty name. In `ls("bucket")` it is the directory `bucket/`, which names the bucket itself, and in `find` it is `bucket//`. Walking into that directory is what caused the recursion in the issue. A key like `dir//x` does the same thing one level down: `ls("bucket/dir")` lists `bucket/dir/` both as the placeholder and again as an empty-named directory.

`_iterdir` now skips a common prefix that is exactly the listed prefix plus `/`, and a content key of exactly `/`. The prefix check only applies when the listing prefix is empty or ends with `/`, so a stem filter such as `prefix="mxpfx"` still returns the real `mxpfx/` directory. Objects under `dir//` are still returned by `find`.

Added `test_slash_only_key`, which fails on `main`. On Windows, `test_s3fs.py` gives the same two failures before and after the change (`test_content_type_is_not_overrided`, `test_get_directory_recursive`); both look like local path-separator issues.


martindurant:
This is going to have a time versus placeholder directories (i.e., "/" ending key names with zero size, not including those ending "//"). Of course, they are also valid keys and can contain data. I see you included some likely tests, but there could be more variants - the batch operations (like cp()) care about whether a given path is directory-like or not.

itzzdev09:
Good point. I compared ls/info/isdir/find/walk/glob/du/expand_path and recursive cp/get/rm between main and this branch, with placeholders both empty and holding data: the only differences are the "/" and "dir//" entries. I added `test_slash_only_key_keeps_placeholders` for that. One thing that stays broken either way: `rm(bucket, recursive=True)` can't remove a "/" key (main fails on the bucket name "bucket/", this branch leaves the key and the bucket delete fails). Should I handle that here too, or keep this PR to listings?

martindurant:
If you can fix it, sure!

itzzdev09:
Sure, will do it tomorrow, it's already late!

itzzdev09:
Done in eac97db: a recursive rm of a whole bucket now checks for a "/" key and deletes it before removing the bucket, with test_rm_bucket_with_slash_only_key. Versioned buckets already go through _rm_versioned_bucket_contents, which lists versions directly, so they aren't affected.

s3fs/core.py
```diff
@@ -2395,13 +2395,25 @@ async def _rm(self, path, recursive=False, **kwargs):
...
         ]
         return sum(out, [])
 
+    async def _rm_slash_key(self, bucket):
+        try:
+            await self._call_s3("head_object", Bucket=bucket, Key="/")
+        except (FileNotFoundError, PermissionError):
+            return
+        await self._call_s3("delete_object", Bucket=bucket, Key="/")
+
```
martindurant: Making a very special case for "I am deleting a bucket, but it contains a `'/'` key that doesn't show up otherwise" seems like the wrong thing to do.  

(not also, that we are not necessarily deleting whole buckets here anyway)

itzzdev09: Yeah, fair enough. Looking at it again it was worse than that too: it ran for every directory being removed, so rm("bucket/subdir", recursive=True) would have deleted the bucket's / key as well.

I've reverted it (6ee2eb2), so the PR is back to just the listing fix. A bucket with a / key still can't be removed with a recursive rm, but that's no different from main. Happy to leave that as a separate issue if you think it's worth tracking.

s3fs/core.py
```diff
@@ -927,6 +927,10 @@ async def _iterdir(
         )
         async for i in it:
             for l in i.get("CommonPrefixes", []):
+                if prefix[-1:] in ("", "/") and l["Prefix"] == prefix + "/":
+                    # keys like "/" or "dir//" give an empty-named prefix that
+                    # would list as the directory itself, so walk loops on it
+                    continue
```
martindurant: The real problem is the recursion, right? We do still want to see these keys, in case we want to read, copy or delete them (else they are just left behind). 
So can we put something in walk() to stop the recursion instead?

PR #1047, opened 2026-09-22
Raise errors for failed objects in bulk deletes

### LINKED ISSUE #838: s3fs rm fails silently
I'm observing that `rm` is swallowing underlying errors and not deleting the files it was requested to delete:
```python
>>> fs.rm(path)
[]
>>> fs.rm_file(path)
<snip>
File C:\python\envs\dev-py310\lib\site-packages\s3fs\core.py:1910, in S3FileSystem._rm_file(self, path, **kwargs)
   1907 self.invalidate_cache(path)
   1909 try:
-> 1910     await self._call_s3("delete_object", Bucket=bucket, Key=key)
   1911 except ClientError as e:
   1912     raise translate_boto_error(e)
  
PermissionError: Access Denied
```

i.e. `rm` is failing silently.

My expectation when I call `rm` is that only one of 2 things will happen - either:
1. *All* files were deleted successfully
2. Or an error is raised

I'm fine with some of the files having been deleted and some not so long as an exception is raised telling me that not all of the files/folders I requested to be deleted were able to be deleted.

pentaoa:
`rm()` currently succeeds when S3 returns HTTP 200 with per-object `Errors` from `DeleteObjects`, even if some or all requested objects remain. Raise the existing Python exception mapping for the first failed object, include its path in the message, and retain the complete S3 response in the exception cause so callers can inspect all failures. Successful responses and cache invalidation keep their existing behavior.

Fixes #838. Regression tests exercise synchronous and asynchronous removal against a local Moto HTTP server with real bucket-policy denials, including partial success. Additional cases cover quiet responses, error-code translation, and an error after a successful 1,000-object batch.

Validation on Python 3.12: the eight new error cases fail on the original code; all ten new cases pass with this change, and the final focused run reports 34 passed. The non-derived suite reports 248 passed, two skipped, three expected failures and two Windows path failures. Both failures also reproduce with the unchanged implementation; the inherited derived tests likewise use Windows separators in S3 keys and fail during setup. Black, flake8 and diff checks pass. [Linux CI](https://github.com/fsspec/s3fs/actions/runs/35696261163) now passes on Python 3.10–3.14, as does pre-commit. The Python 3.12 run reports 387 passed, one skipped and four expected failures, with all ten new regressions executed and passing.

AI assistance: OpenAI Codex implemented the change, wrote the regression tests, and ran the reported local checks.


martindurant:
OK, makes sense. We may tweak this in the future, but this is definitely an improvement.

Question: where does parameter "Quiet" come from?

pentaoa:
`Quiet` is an option in S3's `DeleteObjects` request (`Delete.Quiet`), and `_bulk_delete()` already set it to `True` before this change. [AWS documents](https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObjects.html) that quiet mode omits successful deletions from the response but still reports failures. The regression cases cover this response shape so missing `Deleted` entries do not hide `Errors`.

martindurant:
Thanks for the reminder :)

